PR #21, opened 2019-09-04
Run refactored benchmarks on EC2

danieljanes:


xain/benchmark/benchmark.py
```diff
@@ -69,30 +71,55 @@ def __init__(self, tasks: List[Task], aggregation_name: str):
...
+    logging.info(
+        f"Attempting to run task on EC2: {model}, {dataset}, {R}, {E}, {C}, {B}"
+    )
+    run.ec2(
+        image=docker_image_name,
+        timeout=300,  # TODO dynamic from benchmark config
+        instance_cores=2,  # TODO dynamic from benchmark config
+        # The following arguments will be passed as absl flags:
+        group_name=docker_image_name,
+        task_name=f"task_{dataset}_{model}_{R}_{E}_{C}_{B}",
```
tanertopal: Would be interesting to store instance_cores also in task name

xain/benchmark/exec/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    "group_name", None, "Group name used to gather the tasks of one benchmark"
+)
+
+flags.DEFINE_string("task_name", None, "")
+
+flags.DEFINE_string(
+    "model",
+    None,
+    # pylint: disable-msg=consider-iterating-dictionary
+    f"Model name, one of {[fn for fn in model_fns.keys()]}",
```
tanertopal: Works also without `.keys()` just `[fn for fn in model_fns]`

PR #38, opened 2019-09-05
Get applicable aggregation from config and run

tanertopal:


xain/benchmark/aggregation/task_accuracies.py
```diff
@@ -59,9 +59,13 @@ def prepare_aggregation_data(group_name: str) -> List[PlotValues]:
     data: List[PlotValues] = []
 
     for value in values:
-        print(value)
-        task_class, task_accuracies = value
-        indices = list(range(1, len(task_accuracies) + 1))
+        task_class, task_accuracies, E = value
```
danieljanes: This could be moved into up, i.e. `for task_class, task_accuracies, E in values:`

An even nicer alternative would be to replace the entire block using a map-operation.

PR #44, opened 2019-09-09
Add file based global config setup

tanertopal:
This change will automatically init a global config file in project root where the user can centrally configure the project
Some config flags will serve as defaults for various abseil flags

### DIFF
--- a/xain/__init__.py
+++ b/xain/__init__.py
@@ -23,7 +23,7 @@ flags.DEFINE_string(
 )
 
 flags.DEFINE_string(
-    "S3_bucket",
+    "S3_results_bucket",
     c.get("S3", "results_bucket", fallback=None),
     "Bucket name for the results to be uploaded to",
 )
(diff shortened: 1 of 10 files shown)


xain/__init__.py
```diff
@@ -1,25 +1,29 @@
...
 )
 
 flags.DEFINE_string(
-    "results_dir", results_dir_default, "Results directory as absolute path"
+    "results_dir", c.get("Path", "results_dir"), "Results directory as absolute path"
 )
 
 flags.DEFINE_string(
-    "S3_bucket", "xain-results", "Bucket name for the results to be uploaded to"
+    "S3_results_bucket",
```
danieljanes: 👍

xain/benchmark/__init__.py
```diff
@@ -1,5 +1,7 @@
 from absl import flags
 
+import xain.config
```
danieljanes: Is this import really required?

PR #59, opened 2019-09-11
Make code structure easier to reason about

tanertopal:
This PR changes the generator code base so that its easier to understand what it actually doing without changing the actual output. 
I verified that the resulting hash files are identical.


xain/generator/partition_volume_distributions.py
```diff
@@ -58,18 +57,6 @@ def _generate_volume_distribution(xs: np.ndarray, a: float, b: float) -> List[in
     return [int(exponential_decay(x, a=a, b=b)) for x in xs]
 
 
-def plot_fashion_mnist_dist():
```
danieljanes: why was this deleted?

tanertopal: It creates another matplotlib tkinter import error and I thought it's only purpose was for debugging while development. As it did not have any use in the codebase I thought better reduce surface than let it be.

Alternatively I can not remove it but will have to import matplotlib as in:

https://github.com/xainag/xain/blob/master/xain/benchmark/aggregation/plot.py

PR #61, opened 2019-09-12
Add numpy <-> ndarray example conversion

matklad:
Hi!

So this adds a minimal bridge between Python's numpy.array and Rust's ndarray. 

It mostly just works. https://crates.io/crates/ndarray-npy is used for reading `.npy / .npz`. 

It probably doesn't make sense to merge this PR as is, as it just shows the basic stuff.


matklad:
@tanertopal apparently, I can't put an PR in draft mode. I think we should wait until #66 is merged (which sets up Rust CI), and then merge this PR in form of a smoke test which loads a fixed numpy array.

matklad:
@tanertopal re-made this as a test

controller/Cargo.lock
```diff
@@ -204,16 +296,82 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "unicode-xid 0.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "unicode-xid"
+version = "0.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
```
matklad: I am not too happy about all these dependencies that `ndarray-npy` pulls in, especially about *old* pest & syn. I think at some point in the future we should send a couple of PRs to that crate to modernize its deps, or maybe even roll our own npy reading. Note that, because npy format includes a Python's dict literal as ascii string, reading it is unfortunately not trivial, so *some* parsing dependency, like nom or pest, is understandable.

controller/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+    use ndarray::prelude::*;
+    use ndarray_npy::{ReadNpyExt, WriteNpyExt};
+
+    #[test]
+    fn read_numpy_array() {
+        let fixture = test_data_dir().join("arange(5.0).npy");
+        let file = fs::File::open(fixture.as_path()).unwrap();
+
+        let data = Array1::<f64>::read_npy(file).unwrap();
+        assert_eq!(data, Array1::range(0.0, 5.0, 1.0));
```
matklad: testing the writing part turned out to be cumbersome: ndarray_npy adds an extra comma to the header, which is valid from the spec's point of view, but makes `write . read != id`

matklad:
I think we can and should merge this now

PR #63, opened 2019-09-17
Rendezvous participant coordinator

r-marques:
This merge request adds and initial implementation of a grpc communication between a coordinator and participant by implementing the a _rendezvous_ method.

A coordinator can be started with the number or required participants and will ACCEPT connections until the required number of participants is reached. After that it will reply with LATER and participants try to periodically reconnect.

Currently the tests are integration tests that spawn a coordinator for each test. Another approach proposed by @tanertopal could be to just mock the coordinator to test the participant and vice-versa.


xain/grpc/coordinator.py
```diff
@@ -0,0 +1,48 @@
...
+
+import grpc
+
+from xain.grpc import coordinator_pb2, coordinator_pb2_grpc
+
+_ONE_DAY_IN_SECONDS = 60 * 60 * 24
+
+
+class Coordinator(coordinator_pb2_grpc.CoordinatorServicer):
+    def __init__(self, required_participants=10):
```
tanertopal: I am not sure but it might be useful/easier to grasp to use similar concepts as in the FL package?

For example using something like: `num_participants=100` and `C=0.1` and derive `required_participants` from those numbers (`num_participants * C`).

What do you think?

xain/grpc/coordinator.py
```diff
@@ -0,0 +1,48 @@
...
+        self.required_participants = required_participants
+        self.n_participants = 0
```
tanertopal: Maybe name it `num_selected_participants` or `num_accepted_participants`?

xain/grpc/coordinator.py
danieljanes: I agree

protobuf/xain/grpc/coordinator.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+
+package xain.protobuf.coordinator;
+
+service Coordinator {
+  rpc Rendezvous(google.protobuf.Empty) returns (RendezvousReply) {}
```
danieljanes: I'd propose to introduce a `RendezvousRequest` as a container for e.g. information the client would like to share.

xain/grpc/coordinator.py
r-marques: I agree but what I'm trying at this point is to just get all the necessary grpc communication done before starting to add any business logic.

This was added more for testing purposes. In a next iteration when we start adding businees logic we could make the APIs the same as in the fl module

protobuf/xain/grpc/coordinator.proto
r-marques: :+1: but I will leave the changes to `setup.py` so that we have access to google defined types if needed. I think it's helpful

xain/grpc/coordinator.py
tanertopal: Ok, got it. So lets resolve this for now.

danieljanes: Ok for me

PR #66, opened 2019-09-17
Add controller stub and circle CI configuation

skade:
New version of #65, checking if CircleCI runs.


.circleci/config.yml
```diff
@@ -32,6 +32,24 @@ jobs:
...
+  tests_rust:
+    docker:
+      - image: circleci/rust:1.3.7
+    steps:
+      - checkout
+
+      - run:
+          name: install dependencies
+          command: |
+              rustup install nightly
```
tanertopal: Is it ok if we use a fixed version of rust instead of the nightly or do you see any advantages in using the nightly?

skade: Yes, specifically `async/await`. It's 8 weeks out and I wouldn't start a networked codebase without it nowadays. We will be able to switch to beta by September 26.

tanertopal: Ok, cool. Lets fix the version as soon as there is a stable rust version with `async/await`

tanertopal:
Feel free to merge

.circleci/config.yml
matklad: It might be a good idea to pin to a specific nightly though

skade: Right. Added a rust-toolchain file and skipped the install step.

PR #67, opened 2019-09-18
Tighten up Rust CI

matklad:


controller/rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+version = "Two"
+use_small_heuristics = "Max"
```
matklad: That's not the default, but this is what [compiler uses](https://github.com/rust-lang/rust/blob/64c09694a6ecc434cd3a61ade89beb1de17770c5/rustfmt.toml) and what I personally prefer as well. 

By default, `rustfmt` puts every thing on a separate line, which makes for a low-dencity code (so that formatting code with rustfmt for the first time is something like 1.5x increase in sloc).

This might be better to review diffs, but, to my mind, it makes reading code outside of diff (ie, most of the time) harder. 

I don't have a strong opinion here thogh, we can stick to vanilla config as well

r-marques: I have no preference from my side so I'm ok to go with this.

matklad: @stjepang, @skade any preferences here?

matklad:
Merging optimistically, changing style is trivial while we have a dozen lines of code.

PR #69, opened 2019-09-18
Port federated average to Rust

matklad:


controller/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+mod aggregation;
+
+type ModelParams = ndarray::ArrayD<f64>;
```
matklad: I feel like this would be one of the core types for the coordinator, so it we should pick a good name for it. In the original .py it is called `KerasWeights`, which I don't like, because `Keras` is specific library iiuc, and weights can be easily confused with federated weights. 

Note that this is an array of dynamic dimensionality. While in Rust we can statically-type dimensionality, to have compile time checking and more efficient code, I think that, with external model, we'll fundamentally will have to deal with runtime-determined dimensionality.

Hard-coding `f64` as a data type seems more arbitrary, but I don't think there's a good way to be runtime-polymorphic over element type in Rust.

r-marques: @danieljanes any preference on the naming?

tanertopal: I think its named KerasWeights because the weights are specific to how they are used in a Keras model. They might not make much sense outside of the context of Keras model. I might be wrong though. @danieljanes is the expert on that.

r-marques: This seems good to me. I will approve it and we can rename later if needed.

@tanertopal please approve also if you are fine with this.

danieljanes: That's right @tanertopal (as far as the types go). For naming the variables we also use `theta` (i.e. all weights/parameters of a model), and `weights` for one individual weight tensor.

For clarification: A model usually consists of multiple weight tensors of different dimensionality, for prototyping we used the type alias `KerasModel` for `List[np.ndarray]`. The notion of a model consisting of multiple weight tensors is universal though, we should only get rid of the Keras naming.

PR #80, opened 2019-09-19
gRPC participant <-> coordinator hearbeat

r-marques:
This merge request adds the ability for participants to periodically send heartbeats to the coordinator.
The coordinator keeps a list of connected participants and periodically checks if any participant heartbeat has expired. If so the coordinator drops that participant.

Also added an initial implementation of a thread safe ParticipantContext on the coordinator side. This will allow the coordinator to keep any state if wants about the participants. Currently it only stores the expiration time for the next hearbeat but in the future we can extend it as we wish, for instance to store the current state of participants (IDLE, RUNNING, ....).

This also changes of grpc related tests are marked, so that they are only marked as `xfail` on macos. This way the CI will still fail if any grpc test fails.

To start the coordinator:
```bash
$ python -m xain.grpc.coordinator
```

To start participants:
```bash
$ python -m xain.grpc.participant
```


r-marques:
Still missing tests for the heartbeat code.

r-marques:
I'm still looking into the issue with running the grpc tests in macos.

r-marques:
> how about factoring out the common setup (i.e. the 5 lines or so) into a separate function?

Personally when it comes to tests I don't mind duplication as long as a test is self contained and easy to understand.

In this particular case is only two tests. If it were more we could potentially create a fixture for it.

PR #96, opened 2019-09-22
Xp 101 plot participant selection history

PanicButtonPressed:
This MR is related to a ticket that is about adding a graph for the participant selection history.

Within the MR functions were created for:

- looping through task results of a group result and creating a plot for each hist metrics of a task result
- getting the history metrics for the task results from a group result
- creating a numpy matrix (participant indices x train rounds) as heatmap data input for the 2D regular raster plot
- reading task metrics from a task result
- adjusting a default colormap in order to have color grey for value 0
- plotting and saving the heatmap

Additionally:

- added endpoint `participant_history` to get participant selection history for each task in a group
- implemented functionality to get `num_participants` and `hist_metrics` from a task result

Bildschirmfoto 2019-09-23 um 15 19 38


xain/benchmark/aggregation/aggregation.py
```diff
@@ -8,6 +8,7 @@
     learning_rate,
     task_accuracies,
 )
+from xain.benchmark.aggregation.participant_hist import participant_history
```
danieljanes: Please import the necessary module using the the same structure as previous lines do

PanicButtonPressed: Updated.

PR #98, opened 2019-09-23
Add rust bindings for the .proto files

stjepang:
We're using the [`grpcio`](https://github.com/pingcap/grpc-rs) library developed by PingCAP to generate Rust bindings for the `.proto` files. This code is not runnable - it's merely a library with raw API calls.

As a follow-up to this PR, I'll add an example that establishes a connection to a service and makes RPC calls.


rust/xain-proto/generate.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -e
+cd "$(dirname "$0")"
+
+if ! [[ -x "$(command -v grpc_rust_plugin)" ]]; then
+    echo "Error: grpc_rust_plugin was not found"
+    echo
+    echo "To install, run: cargo install grpcio-compiler"
```
matklad: Maybe just run `cargo install grpcio-compiler --version x.y.z` here? That would be a nice way to pin the version of `grpcio-compiler` we are using.

rust/xain-proto/generate.sh
```diff
@@ -0,0 +1,28 @@
...
+for proto in $proto_files; do
+    echo "Processing: $proto"
+    protoc \
+        --rust_out=$PWD/src \
+        --grpc_out=$PWD/src \
+        --plugin=protoc-gen-grpc=`which grpc_rust_plugin` \
+        --proto_path=../../numproto \
+        --proto_path=../../protobuf/xain/grpc \
+        $proto
+done
```
matklad: Can we run this script on CI and check that the generated file is fresh?

rust/xain-proto/generate.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if ! [[ -x "$(command -v grpc_rust_plugin)" ]]; then
+    echo "Error: grpc_rust_plugin was not found"
+    echo
+    echo "To install, run: cargo install grpcio-compiler"
+    exit 1
+fi
+
+proto_files="
+../../numproto/numproto/protobuf/ndarray.proto
```
r-marques: This file does not exist in this repo. Is this something that we need to manually add before being able to compile the proto files?

If so it would be helpful to have a README under the `rust/` folder detailing the steps to build the rust side.

rust/xain-proto/generate.sh
stjepang: Done.

rust/xain-proto/generate.sh
stjepang: Ooops, I just cloned the `numproto` repository manually. Now I've added it as a submodule to this repository and the `generate.sh` script will update that submodule before trying to generate Rust bindings for `ndarray.proto`.

rust/xain-proto/generate.sh
stjepang: Not sure, maybe. A roadblock with that approach might be the fact that generated Rust files are not always perfect and need some small manual tweaks before they compile. 

We didn't run into any of those issues here, but I did last time I played with grpcio. One of the issues was that methods named `mod` were not properly escaped with `r#mod`.

stjepang:
I have addressed comments in this PR. If this looks good, could we merge it so that #110 can be rebased on top of it?

r-marques:
It's fine by me

PR #99, opened 2019-09-24
implement coordinator's state machine scaffold

matklad:
This implements the scaffold for the Coordinator protocol. It is implemented as an explicit state machine (in contrast to Python implementation, which constructs a state machine implicitly on the stack, by blocking). It is not bound to any particular transport, and so should be testable in isolation. 

I this is a WIP, as I want to tweak the representation a bit, to make happy path more clear. 

I am also not sure what the actual protocol do we want here: currently it's just:

* clients send join to coordinator
* for n rounds
  * coordinator broadcasts model to clients
  * clients send updated models to the coordinator

### DIFF
--- a/rust/xain-coordinator/src/training.rs
+++ b/rust/xain-coordinator/src/training.rs
@@ -1,0 +1,337 @@
+use std::{
+    any::Any,
+    collections::{BTreeMap, BTreeSet},
+    fmt,
+    time::Duration,
+};
+
+use crate::{Model, ModelDim};
+
+macro_rules! err {
+    ($($tt:tt)*) => {
+        ProtocolError { msg: format!($($tt)*) }
+    }
+}
+
+macro_rules! bail {
+    ($($tt:tt)*) => {
+        return Err(err!($($tt)*))
+    }
+}
+
+#[derive(Debug)]
+pub struct ProtocolError {
+    msg: String,
+}
+
+impl fmt::Display for ProtocolError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(&self.msg, f)
+    }
+}
+
+impl std::error::Error for ProtocolError {}
+
+type ParticipantId = u32;
+
+#[derive(Debug)]
+pub struct FromParticipant<T> {
+    pub from: ParticipantId,
+    pub payload: T,
+}
+
+#[derive(Debug)]
+pub enum InMessage {
+    Joined(FromParticipant<()>),
+    Update(FromParticipant<Model>),
+    RoundTimeout,
+}
+
+impl InMessage {
+    pub fn new_joined(from: ParticipantId) -> InMessage {
+        InMessage::Joined(FromParticipant { from, payload: () })
+    }
+
+    pub fn new_update(from: ParticipantId, model: Model) -> InMessage {
+        InMessage::Update(FromParticipant { from, payload: model })
+    }
+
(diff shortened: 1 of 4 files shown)


matklad:
Should be ready for review now. The main question here is to we actually want this reified protocol state machine to work with? I am not sure, but I lean towards "yes".

rust/xain-coordinator/src/training.rs
```diff
@@ -0,0 +1,337 @@
...
+type ParticipantId = u32;
+
+#[derive(Debug)]
+pub struct FromParticipant<T> {
+    pub from: ParticipantId,
+    pub payload: T,
+}
+
+#[derive(Debug)]
+pub enum InMessage {
```
skade: Just out of interest: Do we also want to handle client timeouts through InMessage?

rust/xain-coordinator/src/training.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+#[derive(Debug)]
+pub struct OutMessage {
+    pub to: ParticipantId,
+    pub kind: OutMessageKind,
+}
+
+#[derive(Debug)]
+pub enum OutMessageKind {
```
skade: I'm unsure about the semantics of OutMessage. Are those all broadcasted to participants?

rust/xain-coordinator/src/training.rs
```diff
@@ -0,0 +1,337 @@
...
+impl Training {
+    pub fn new(params: TrainingParams) -> Training {
+        Training { params, state: TrainingState::Joining { participants: BTreeSet::new() } }
+    }
+
+    pub fn on_message(&mut self, message: InMessage, io: &mut dyn IO) -> Result<(), ProtocolError> {
+        log::info!("on_message({:?})", message);
+        match (&mut self.state, &message) {
+            (TrainingState::Joining { participants }, InMessage::Joined(joined)) => {
+                let is_new = participants.insert(joined.from);
```
skade: I'm unsure of this case: joining twice might happen when the first join only half succeeds (connection breaking suddenly). I think joining twice should be valid.

Maybe, joining should carry a _participant state_.

rust/xain-coordinator/src/training.rs
matklad: Yeah, that's my plan. The coordinator will receive `InMessageKind::Leave` message if the client drops. It's the job of the transport layer to synthesize `Leave` if the client seems dead. That is, I think it makes sence to treat hearbeats as a transport layer concern.

rust/xain-coordinator/src/training.rs
matklad: So, `OutMessage` has a `to` field, so they are not broadcasted. However, in the actual protocol, at the moment, we indeed just send every message to every client (by just looping). Seems like it might make sense to support explicit broadcast message.

rust/xain-coordinator/src/training.rs
matklad: Good question! I think behavior in case of duplicate messages should be explicitly designed. I think we can assume that transport layer guarantees that messages are delivered at most once, so we only need to protect from cases where the client stops and restarts. Can we push this responsibility to the transport layer? That is, make the transport layer tag each message beetwen join/leave with a session cookie, and enforce the invariant that sessions are disjoint?

matklad:
Let's merge this for know. I think this indeed doens't handle all of the deails about joining and broadcasting at the moment, but it seems like that would be easier to figure out once we add a grpc layer

PR #104, opened 2019-09-24
Small benchmark adjustments

danieljanes:


xain/benchmark/aggregation/final_task_accuracies.py
```diff
@@ -112,9 +112,9 @@ def aggregate() -> str:
 
     fpath = plot(
         data,
-        title="Max achieved accuracy for unitary and federated learning",
-        xlabel="partitioning grade",
-        ylabel="accuracy",
+        title="",  # TODO
```
r-marques: Is this something we want to do in this pr?

xain/benchmark/aggregation/task_accuracies.py
```diff
@@ -96,8 +96,8 @@ def aggregate() -> str:
 
     fpath = plot(
         data,
-        title="Validation set accuracy for unitary and federated learning",
-        ylabel="accuracy",
+        title="",  # TODO
```
r-marques: Same as above

xain/benchmark/aggregation/final_task_accuracies.py
danieljanes: No, it's not for this PR. I left the comment in there to follow up on this - ideally we want to include a way in our benchmark "DSL" to configure this per group or per task

xain/benchmark/aggregation/task_accuracies.py
danieljanes: Same as above

PR #110, opened 2019-10-01
Rust coordinator grpc

skade:
Adds a GRPC server using the generated GRPC services. Incorporates #98, the issues there should be resolved

The server does:

* Client/Server authentication using TLS
    * signed certificates for localhost are included for testing, _do not use in production_
    * password for the CA certificate is 123456
* Accepts both coordinator requests and ndarray requests.
* A test client is also included


stjepang:
Rebased on master after #98 got merged.

rust/xain-grpc-api/src/bin/client.rs
```diff
@@ -0,0 +1,62 @@
...
+use grpcio::{ChannelBuilder, ChannelCredentials, ChannelCredentialsBuilder, EnvBuilder};
+use xain_proto::coordinator::RendezvousRequest;
+use xain_proto::coordinator_grpc::CoordinatorClient;
+
+use clap::{App, Arg};
+
+type AppError = Box<dyn std::error::Error + Send + Sync>;
+
+fn load_certificates(args: &clap::ArgMatches) -> Result<ChannelCredentials, AppError> {
+    // TODO: load certificates dynamically
```
stjepang: What does this TODO suggest? How would we load certificates dynamically?

skade: Heh, missed removing it. The lines below used hard-coded filenames.

stjepang:
I did several changes:

* Merged crates `xain-proto` and `xain-grpc-api` into `xain-grpc` to mimic the directory structure of the Python implementation better.
* Split the client/server into 4 binaries `{coordinator,numproto}_{client,server}`.
* Simplified logging setup so that there is less code and we print info level on the screen by default.
* Added more comments to the code and a new readme file to `xain-grpc`.

Hopefully the code is a bit easier to follow now and I didn't make any unintended changes :)

As far as I'm concerned, this PR is now fully ready for review.

stjepang:
I can also help take over #111 if this looks good!

stjepang:
pinging @r-marques for review

rust/xain-grpc/certs/ca-key.pem
```diff
@@ -0,0 +1,30 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
tanertopal: Can't we generate these locally instead of commiting them? E.g. generate if not present. I think this would avoid potential mishaps in the long run.

tanertopal:
Left few comments. Can't review the actual Rust code for lack of expertise in Rust.
@r-marques did you already have a look?

rust/xain-grpc/certs/ca-key.pem
skade: We could do so, though I don't see a huge potential for this, as Rust code is usually shipped out of the target directory.

We could move them to the test directory first.

I'll write a script to generate them, also to make sure we can regenerate them with the same settings.

skade:
Added a command to generate the certificates instead of putting them into the repos.

rust/xain-grpc/generate.sh
```diff
@@ -11,21 +11,21 @@ if ! [[ -x "$(command -v protobuf-bin-gen-rust-do-not-use)" ]]; then
...
 
-git submodule update --init --recursive
+NUMPROTO_DIR=`pip3 show numproto | grep Location | sed -e 's/^Location: //'`
 
-proto_files="
+PROTO_FILES="
 ../../protobuf/xain/grpc/coordinator.proto
 ../../protobuf/xain/grpc/hellonumproto.proto
-./numproto/numproto/protobuf/ndarray.proto
+$numproto/numproto/protobuf/ndarray.proto
```
r-marques: I think this also needs to be renamed to `NUMPROTO_DIR`

r-marques:
I also enabled CircleCI to to run on forked pull requests. Hopefully when a new commit comes in the CI should be triggered.

stjepang:
We have CI failures, but they seem unrelated to this PR.

r-marques:
> We have CI failures, but they seem unrelated to this PR.

There were some missing system dependencies for the rust job on the CI (cmake for grpcio-sys and go for boringssl).

Syncing we the current master should fix that.

r-marques:
And the failing python tests are because they require access to aws and forked pull requests don't have access to those environment variables

stjepang:
Merged with the `master` branch and now the only remaining problem is that this PR can't access AWS.

PR #113, opened 2019-10-04
gRPC coordinator <-> participant training round

finiteprods:
_Work in progress_. This adds the ability for the coordinator and participant to exchange messages of the appropriate structure for a training round over gRPC.

A participant's training round begins by sending a `startTraining` request, to indicate availability for training. The coordinator responds with a reply, containing the model weights (with properly serialised `numpy` arrays via [`numproto`](https://github.com/xainag/numproto)) and other data.

The training round ends when the participant sends an `endTraining` request containing a model and history data. Note that unlike the current implementation of a _local_ training round, there is no separate call for metrics - to minimise communication, metrics are sent along in the same request.

For now, some hard-coded data is used to demonstrate that messages are (de)constructed, sent and received correctly by both parties.


xain/grpc/participant.py
```diff
@@ -33,6 +35,53 @@ def rendezvous(channel):
         response = reply.response
 
 
+def startTraining(channel):
```
r-marques: python methods should use snake case

xain/grpc/participant.py
```diff
@@ -33,6 +35,53 @@ def rendezvous(channel):
...
+    nda1, nda2 = np.arange(20, 30), np.arange(30, 40)
+    num = 2
+    print(f"Participant requests: theta' {nda1}, {nda2}; {num} examples")
+    theta_prime = [ndarray_to_proto(nda1), ndarray_to_proto(nda2)]
+    theta_update = coordinator_pb2.EndTrainingRequest.ThetaUpdate(
+        theta_prime=theta_prime, num_examples=num
+    )
+    # history data
+    k1, k2 = "aaa", "bbb"
+    v1, v2 = [1.1, 2.1], [3.1, 4.1]
```
r-marques: Instead of having some hardcoded values here could we instead turn this into a test.

This would also help us make the `start` and `end` training methods more generic

xain/grpc/coordinator.py
```diff
@@ -158,6 +177,28 @@ def Heartbeat(self, request, context):
         self.participants.update_expires(context.peer())
         return coordinator_pb2.HeartbeatReply()
 
+    def StartTraining(self, request, context):
+        print(f"Received: {type(request)} from {context.peer()}")
+        # NOTE self.theta is assumed to be initialised
```
r-marques: In this case `theta` shouldn't be an optional argument

xain/grpc/conftest.py
```diff
@@ -39,3 +39,15 @@ def participant_stub():
     stub = coordinator_pb2_grpc.CoordinatorStub(channel)
 
     return stub
+
+
+# similar to coordinator_service but yields the coordinator servicer
```
r-marques: Instead of a new fixture can't we just change `coordinator_service`? It shouldn't break existing tests.

xain/grpc/coordinator.py
finiteprods: Following the suggestion in the [Hitchhiker's guide](https://docs.python-guide.org/writing/gotchas/), `theta` now has a sensible default of `[]` (without running into the "dangerous default" warning) so that `StartTraining` can be called safely.

PR #112, opened 2019-10-02
Port existing documentation to sphinx

r-marques:
- Added versioning. Specially important later on for readthedocs
- Added support for writing documentation in markdown
- Integrated with github (show fork banner)
- Added "logo" and images more as an example

To build the documentation:
```bash
# this will also install new doc related dependencies
$ pip install -e .[dev]

$ cd docs/
$ make html
```
Open `docs/_build/html/index.html` in the browser

Screenshot from 2019-10-02 14-46-59

PR #114, opened 2019-10-07
Add code documentation to sphinx

r-marques:
Depends on #112 

- Configured sphinx to be able to generate documentation from docstrings.
- Fixed some bad formatting in some docstrings

The instructions to build the docs also changed since we also need to generate the documentation from the docstrings. This is done with a new make target:
```bash
$ cd docs/
$ make docs
```

### DIFF
--- a/xain/benchmark/aggregation/final_task_accuracies.py
+++ b/xain/benchmark/aggregation/final_task_accuracies.py
@@ -67,7 +67,7 @@ def prepare_aggregation_data(
 
     Returns:
         A tuple containing a list of `PlotValues` and a list of tuples
-        containing (`XticksLocations`, `XticksLabes`)
+        containing (`XticksLocations`, `XticksLabels`)
 
     """
     group_dir = os.path.join(FLAGS.results_dir, group_name)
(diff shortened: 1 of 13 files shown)


docs/quick.md
```diff
@@ -75,7 +78,7 @@ $ pull_results
 $ plot_final_task_accuracies --group_name=GROUP_NAME
```
tanertopal: This script is actually removed and its now only:

```shell
aggregate --group_name GROUP_NAME
```

xain/benchmark/aggregation/final_task_accuracies.py
```diff
@@ -60,17 +60,15 @@ def group_values_by_class(
...
         group_name (str): group name for which to construct the curves
 
     Returns:
-        (
-            [("unitary", unitary_accuracies, indices),
-            ("federated", federated_accuracies, indices)],
-            (xticks_locations, xticks_labels))
-        )
+        A tuple containing a list of `PlotValues` and a list of tuples
+        containing (`XticksLocations`, `XticksLabes`)
```
tanertopal: Typo in `XticksLabes`

docs/quick.md
r-marques: Will be fixed once I rebase #112

PR #116, opened 2019-10-08
Fixes broken docker build

tanertopal:
Fixes broken docker build. A followup will be to add an e2e test for docker builds as we are currently in CI not testing if the image can still be build.


tanertopal:
Thank you! Will do in a separate PR

PR #117, opened 2019-10-08
Pass AWS credentials into docker container

tanertopal:
Pass environment variables so on local docker run AWS credentials can be present. Currently results can't be pushed as the AWS_PROFILE even when set is local to the machine.
I assume as the local docker run is more for debugging this is an acceptable implementation for now.


xain/ops/run.py
```diff
@@ -44,11 +44,14 @@ def docker(image: str, timeout: int = 300, instance_cores=2, **kwargs):
     command = [
         "docker",
         "run",
-        "--rm",
         "-d",
         f"--stop-timeout={timeout}",
         f"--cpus={instance_cores}",
         "-e",
+        f"AWS_ACCESS_KEY_ID",
```
r-marques: Do we need the `f`?

xain/ops/run.py
```diff
@@ -44,11 +44,14 @@ def docker(image: str, timeout: int = 300, instance_cores=2, **kwargs):
...
+        "-e",
+        f"AWS_SECRET_ACCESS_KEY",
```
r-marques: same as above

tanertopal: True we don't! Will adjust.

PR #129, opened 2019-10-31
Standalone gRPC-enabled Participant

finiteprods:
_Work in Progress_. This takes the essentials of the existing Participant logic and makes it a standalone process, which is important for the proper distributed setting of federated learning. This involves implementing a state machine that communicates with the Coordinator. The communication will use the gRPC messages (and message handling) already implemented in the [rendezvous](https://github.com/xainag/xain/pull/63), [heartbeat](https://github.com/xainag/xain/pull/80) (with an extension for [detecting state changes](https://github.com/xainag/xain/pull/127)) and [training rounds](https://github.com/xainag/xain/pull/113).

Some simplifying assumptions are made for now. For example, a `partition_id` is assumed to be known when the Participant is started.


xain/grpc/participant.py
```diff
@@ -65,14 +80,46 @@ def end_training(
...
+    # ready:
+    theta, epochs, base = start_training(channel)
+    # training:
+    theta_n, his, _dict = participant.train_round(theta, epochs, base)
+    # NOTE _dict is the opt_config - ignore for now
+    met = participant.metrics()
+    end_training(channel, theta_n, his, met)
+    # back to standby unless terminate event says otherwise
+    if not terminate.isSet():
+        standby(channel, participant, terminate, selected)
```
r-marques: Instead of a recursive call couldn't we just wrap the body of the this method with a `while not terminate.is_set()`?

xain/grpc/participant.py
```diff
@@ -65,14 +80,46 @@ def end_training(
...
+        print(f"Not yet selected for round. Retrying in {RETRY_TIMEOUT}s")
+    # ready:
+    theta, epochs, base = start_training(channel)
+    # training:
+    theta_n, his, _dict = participant.train_round(theta, epochs, base)
+    # NOTE _dict is the opt_config - ignore for now
+    met = participant.metrics()
+    end_training(channel, theta_n, his, met)
+    # back to standby unless terminate event says otherwise
+    if not terminate.isSet():
```
r-marques: We should prefer using `is_set` instead of `isSet`. That must be there for backwards compatibility https://github.com/python/cpython/blob/3.7/Lib/threading.py#L511

xain/grpc/participant.py
```diff
@@ -81,18 +128,27 @@ def run():
...
-    try:
-        # never returns unless there is an exception
-        heartbeat_thread.join()
-    except KeyboardInterrupt:
-        terminate_event.set()
-        channel.close()
+    # standby:
+    standby(channel, p, terminate_event, selected_event)
+
+    # try:
```
r-marques: I think it's helpful to keep this logic if we want to shutdown the participant without a `FINISHED` event

xain/grpc/participant.py
```diff
@@ -3,20 +3,35 @@
...
         reply = stub.Heartbeat(coordinator_pb2.HeartbeatRequest())
         print(f"Participant received: {type(reply)}")
+        if reply.state == coordinator_pb2.State.FINISHED:
+            terminate_event.set()
+            return
+        if reply.state == coordinator_pb2.State.ROUND:
+            # signal "round open" to main thread
+            selected_event.set()
+        # not much to do for State.STANDBY
+        selected_event.clear()
```
r-marques: Here you are possibly setting and clearing the event very fast. I don't know if this is enough for the `train_when_selected` to pick it up. I think it would be safer for the event to be cleared inside `train_when_selected`.

Also this still needs logic to check the number of the round or else you may be doing multiple trainings per round.

finiteprods: I thought about this too at the time - a quick search suggested it was fine, but I take your suggestion as an appropriate place to `clear()`.

Yes OK, because the Coordinator moves from ROUND i -> ROUND i+1. I had a slightly different mental model of the Coordinator so that the Participant could maybe get away with not tracking the round number. Maybe we can discuss later.

finiteprods:
Updated to properly cope with [heartbeat messages from Coordinator](https://github.com/xainag/xain/pull/127). A thread-safe `StateRecord` object tracks the state of the Participant (consisting of a `ParState` enum value and round number). The state transition function is applied by a "message loop" thread that handles the incoming heartbeats. This state machine should be able to deal with per-round Participant _selection_ immediately. 

Further **TODO**s:
- [ ] logging
- [ ] more handling around edge cases
- [ ] empty heartbeats are sent for now (easy to add later if necessary)
- [ ] unit tests
- [ ] update state diagram

xain/grpc/participant.py
```diff
@@ -65,34 +86,163 @@ def end_training(
...
+            elif msg == coordinator_pb2.State.FINISHED:
+                st.state = ParState.DONE
+                st.cv.notify()
+        elif st.state == ParState.POST_TRAINING:
+            if msg == coordinator_pb2.State.STANDBY:
+                # not selected
+                st.state = ParState.WAITING_FOR_SELECTION
+                # prob ok to keep st.round as it is
+                st.cv.notify()
+            elif msg == coordinator_pb2.State.ROUND and r == st.round + 1:
```
r-marques: I think that here we should check if `r > st.round`. That way the participant is not required to participate in every round. Makes it a bit more flexible

finiteprods: Sensible suggestion but it's more flexible than it might look. There's no selection yet in the Coordinator, but let's imagine it's communicated in perhaps the simplest way: it advertises `ROUND i+1` to the selected, and `STANDBY` to those not. Then a Participant in `POST_TRAINING i` expects just one of those 2 messages - if it's not required in round i+1 it should go back to `WAITING_FOR_SELECTION` (the case above). So a message with `r > st.round + 1` would be unexpected.

r-marques: The case that I was thinking of is if let's the participant was selected for ROUND 1 and skipped for ROUND 2 and then selected again for ROUND 3. Would this work?

finiteprods: It should. The participant is selected via the heartbeat `ROUND 1` (thereby transitioning `WAITING_FOR_SELECTION` -> `TRAINING 1`). Once training is finished, it waits (`TRAINING 1` -> `POST_TRAINING 1`). At some point it realises from the heartbeat `STANDBY` that it's not required for round 2 and so it's back to more waiting (`POST_TRAINING 1` -> `WAITING_FOR_SELECTION`). Then round 2 is over, and `ROUND 3` signals it back into action (`WAITING_FOR_SELECTION` -> `TRAINING 3`). So on.

xain/grpc/participant.py
```diff
@@ -1,17 +1,44 @@
...
+    """
+
+    WAITING_FOR_SELECTION = auto()
+    TRAINING = auto()
+    POST_TRAINING = auto()
+    DONE = auto()
+
+
+# deprecated: see message_loop
 def heartbeat(channel, terminate_event):
```
r-marques: If we are not using this we can just remove it.

finiteprods: Some of the gRPC tests use this. I can see whether it makes sense to port them, or remove them if need be.

xain/grpc/participant.py
```diff
@@ -1,17 +1,45 @@
+"""Module implementing the networked Participant using gRPC.
+"""
 import threading
 import time
+from enum import Enum, auto
 from typing import Tuple
 
 import grpc
+
+# from absl import app, flags
```
r-marques: All of this commented code should be removed if not needed

xain/grpc/participant.py
```diff
@@ -65,34 +118,240 @@ def end_training(
...
     )
     # send request to end training
     reply = stub.EndTraining(req)
     print(f"Participant received: {type(reply)}")
 
 
-def run():
-    # create a channel to the coordinator
-    channel = grpc.insecure_channel("localhost:50051")
+# def init_participant() -> Participant:
```
r-marques: Same here. Can we remove this commented code?

xain/grpc/participant.py
```diff
@@ -65,34 +118,240 @@ def end_training(
...
+                st.state = ParState.TRAINING
+                st.round = r
+                st.cv.notify()
+            elif msg == coordinator_pb2.State.FINISHED:
+                st.state = ParState.DONE
+                st.cv.notify()
+        elif st.state == ParState.POST_TRAINING:
+            if msg == coordinator_pb2.State.STANDBY:
+                # not selected
+                st.state = ParState.WAITING_FOR_SELECTION
```
r-marques: There is a problem with these state transtions.

Imagine the case where a  participant goes down at the middle of a round and eventually a new one connects. During this time the coordinator with have the following state transitions: `ROUND -> STANDBY -> ROUND`. They all happen in the same round so the round number remains the same.

If this happens while a participant is in the `POST_TRAINING` state the participant with transition to `WAITING_FOR_SELECTION` and then `TRAINING` possibly more than once in the same round.

I think that every transition into `TRAINING` should check whether the participant has already trained in the round or not.

xain/grpc/participant.py
```diff
@@ -65,34 +118,240 @@ def end_training(
...
+        with self.cv:
+            self.cv.wait_for(
+                lambda: self.state
+                in {ParState.TRAINING, ParState.WAITING_FOR_SELECTION, ParState.DONE}
+            )
+            # which one was it?
+            return self.state
+
+
+def transit(st, beat_reply):
```
r-marques: For future reference I think that we could use Jan's suggestion of removing the POST_TRAINING state and maybe reuse only the proto defined states `READY, TRAINING and FINISHED`

xain/grpc/participant.py
```diff
@@ -65,34 +118,240 @@ def end_training(
...
+        pass
 
-    try:
-        # never returns unless there is an exception
-        heartbeat_thread.join()
-    except KeyboardInterrupt:
-        terminate_event.set()
-        channel.close()
 
+def begin_training(st, chan, part):
```
r-marques: Also for future reference I would like to have a better way to organize this logic so that we don't have constant recursive calls between `begin_selection_wait` and `begin_training`

xain/grpc/test_participant.py
```diff
@@ -0,0 +1,88 @@
...
+    st = StateRecord(state=ParState.POST_TRAINING, round=5)
+    start_state = st.lookup()
+    hb = coordinator_pb2.HeartbeatReply(state=coordinator_pb2.ROUND, round=5)
+    transit(st, hb)
+    assert st.lookup() == start_state
+    # old round? shouldn't affect me...
+    hb.round = 0
+    transit(st, hb)
+    assert st.lookup() == start_state
+    # future round? not expecting this...
```
r-marques: Why not? What it the participant was selected for round 5, skipped for round 6 and selected again for round 7?

xain/grpc/test_participant.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_posttraining_to_done():
+    st = StateRecord(state=ParState.POST_TRAINING, round=6)
+    hb = coordinator_pb2.HeartbeatReply(state=coordinator_pb2.FINISHED)
+    transit(st, hb)
+    assert st.lookup() == (ParState.DONE, 6)
+
+
+def test_posttraining_to_waiting():
```
r-marques: Could we try a test similar to this where the coordinator goes to STANDY and then ROUND without changing the round number?

xain/grpc/participant.py
finiteprods: Yes. To reiterate what's said, in this situation the coordinator decides to "pause" the round, by telling all participants `STANDBY`. Later it resumes it by sending `ROUND` again (with the same round number as before). So when a participant receives `STANDBY` it shouldn't necessarily assume that the round is already over (as it currently does). Yes, as suggested it could check the round number at this point to make sure it doesn't repeat work already done. But equally, the coordinator (which has already received the update from it) could just continue to tell this participant `STANDBY` - this would be part of the participant-selection logic in the coordinator. I propose postponing an immediate "quick-fix" because
1. I consider round pausing / resuming not critical at this point in time especially as we haven't considered all the fault scenarios very much anyway.
2. This problem should be resolved by a proposal to reduce the participant state machine (which _will_ check round numbers as suggested) and / or participant-selection logic in the coordinator.

xain/grpc/participant.py
finiteprods: Yes I have a proposal for reducing the participant states down to two making it a bit simpler conceptually

xain/grpc/participant.py
finiteprods: The main reason for favouring function calls here was mainly for readability / close correspondence with the state transition system - e.g. you can think of a `begin_training` call to be a state transition to the `TRAINING` state. But that is maybe a matter of taste. I would be in favour of a different way if it is more readable or there is a performance impact.

xain/grpc/test_participant.py
finiteprods: Then it should first receive a `STANDBY` before getting `ROUND 7`. But I agree this could be more intuitive, and the simpler 2-state machine should help with this,

xain/grpc/participant.py
finiteprods: updated `test_participant_heartbeat` to use `message_loop`

PR #132, opened 2019-11-05
Call `pip` using `python -m` to ensure the right Python interpreter g…

danieljanes:
…ets used

Details/motivation: https://snarky.ca/why-you-should-use-python-m-pip/

Even though it might not be strictly necessary for Docker, it's still good to use a uniform approach for calls to `pip`, hence the changes to `Dockerfile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /opt/app
 RUN mkdir output
 
 # Upgrade pip and setuptools
-RUN pip install -U pip==19.3.1 setuptools==41.6.0
+RUN python -m pip install -U pip==19.3.1 setuptools==41.6.0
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR /opt/app
 RUN mkdir output
 
 # Upgrade pip and setuptools
-RUN pip install -U pip==19.3.1 setuptools==41.6.0
+RUN python -m pip install -U pip==19.3.1 setuptools==41.6.0
```
matklad: it’s might be better to use python3 rather than python. The latter might refer to python2 in some environments. Doesn’t matter at all for docker, but might make a difference in readme

PR #133, opened 2019-11-05
hello world example to confirm i'm set up

wilk10:
Just a simple helloworld script to confirm i am set up correctly and experience the workflow, and also to create the folder where to add examples


danieljanes:
@wilk10 please adjust scripts (such as `format.sh`) so that they capture code in `examples` as well

setup.py
```diff
@@ -90,6 +90,8 @@ def run(self):
     "twine==2.0.0",  # Apache License 2.0
 ]
 
+examples_require = ["tensorflow==1.14.0"]  # Apache 2.0
```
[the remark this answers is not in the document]
danieljanes: We're in the process of refactoring the `xain` package in a way that will remove the need for `tensorflow` in `install_requires`, so I'd recommend to keep the redundant require for a while - other examples will require other ML frameworks and it would be helpful to have those dependencies already separated from the `xain` ones

scripts/test.sh
```diff
@@ -24,19 +24,19 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 # check format of proto files
 clang_format && echo "===> clang-format says: well done <===" &&
 
 # lint
-pylint --rcfile=pylint.ini xain && echo "===> pylint says: well done <===" &&
+pylint --rcfile=pylint.ini xain examples && echo "===> pylint says: well done <===" &&
 
 # type checks
-mypy --ignore-missing-imports xain && echo "===> mypy says: well done <===" &&
+mypy --ignore-missing-imports xain examples && echo "===> mypy says: well done <===" &&
```
r-marques: Instead of `examples` you should have `examples/*`.

The reason for this is that examples is not a python module and probably we don't want it to be so we need to instruct mypy to recursively check for python files inside that folder.

wilk10: thanks, changed as suggested

wilk10:
Closing because i messed up the history with the rebase from master.. opening a cleaner new one here: https://github.com/xainag/xain/pull/142

PR #134, opened 2019-11-05
Move benchmark sub-packge to new top-level benchmarks directory

danieljanes:
Work towards XP-175


xain/fl/participant/participant_test.py
```diff
@@ -31,33 +31,34 @@ def test_Participant_x_y_shape_invalid():
...
-        use_lr_fn=False,
-    )
-    weights = model_provider.init_model().get_weights()
-
-    # Execute
-    (_, num_examples_actual), _, _ = participant.train_round(weights, 2, 0)
-
-    # Assert
-    assert num_examples_actual == num_examples_expected
+# FIXME refactor to remove dependency on benchmark code
```
r-marques: Should this be fixed now or is there any task created to solve this?

xain/fl/participant/participant_test.py
```diff
@@ -1,10 +1,10 @@
 import numpy as np
 import pytest
 
-from xain.benchmark.net import model_fns
+# from benchmarks.benchmark.net import model_fns  # FIXME refactor
```
r-marques: Should this be fixed now or is there any task created to solve this?

xain/fl/participant/participant_test.py
```diff
@@ -1,10 +1,10 @@
...
 from xain.datasets import load_splits
 
-from .model_provider import ModelProvider
+# from .model_provider import ModelProvider  # FIXME refactor
```
r-marques: Same as above

setup.py
```diff
@@ -132,7 +132,7 @@ def run(self):
         "Operating System :: MacOS :: MacOS X",
         "Operating System :: POSIX :: Linux",
     ],
-    packages=find_packages(exclude=["*_test.py"]),
+    packages=find_packages(where=".", exclude=["*_test.py"]),
```
r-marques: Shouldn't we exclude the `benchmarks` folder?

xain/fl/participant/participant_test.py
danieljanes: We're in the process of refactoring how model loading works substantially, so I'd say we don't need to invest the effort to fix this now only to re-work it next week. I've created a ticket to not loose track of it: XP-184

xain/fl/participant/participant_test.py
danieljanes: Same as above

PR #142, opened 2019-11-07
Example tf v2

wilk10:
Creating a new PR because example-tf commit history is messy. This PR involves creating a new examples module where scripts with examples on how to use the platform will be placed. Adding a helloworld example and fixing scripts


danieljanes:
I'd recommend to give the subdirectory under `examples` a more descriptive name since other examples will follow. Also TensorFlow is written without a space, so maybe `examples/tensorflow_hello_world` would make sense.

PR #148, opened 2019-11-12
change to Google docstrings style. remove TODO comment

wilk10:


examples/tensorflow_hello_world/helloworld.py
```diff
@@ -9,91 +9,112 @@
...
+    This function fetches the data split into training partitions, validation and test
+    dataset. A Keras model is then declared and compiled, potential Participants are
+    initiated, as well as the Coordinator. We then call fit and evaluate on the
+    Coordinator and print some basic stats as well as final performance metrics.
+
+    Args:
+        None
+
+    Returns:
+        None
```
r-marques: You don't have the `Args` or `Returns` section it there is nothing to return.

examples/tensorflow_hello_world/helloworld.py
```diff
@@ -230,12 +265,15 @@ def print_dataset_stats(xy_train_partitions, xy_validation, xy_test):
...
-    Printing the validation metrics to the terminal.
+    """Prints the validation metrics to the terminal.
+
+    Args:
+        validation_metrics: Dictionary containing loss and accuracy per each round.
 
-    :param validation_metrics: dictionary containing loss and accuracy per each round
-    :return: None
+    Returns:
+        None
```
r-marques: Same as above

r-marques:
Just some minor comments but the it looks good :+1:

wilk10:
Thanks. Corrected and pushed again

examples/tensorflow_hello_world/helloworld.py
```diff
@@ -9,91 +9,106 @@
...
 
-# TODO: this feels a bit hacky. it's the only flag that needs to be added here.
-#  any better solution?
+# Defining the 'task_name' flag here, to be used by the absl-py app.
 FLAGS = flags.FLAGS
 flags.DEFINE_string("task_name", None, "")
 
 
-# We first define some constants for this specific example
+DATASET_NAME = "fashion-mnist-100p-iid-balanced"
```
danieljanes: Reading it now it seems like we can inline all single-use constants, i.e. remove the definition here and add the value (+ documentation) where it is actually being used.

examples/tensorflow_hello_world/helloworld.py
```diff
@@ -9,91 +9,106 @@
...
-    # Evaluate the model with the test set
+    # Evaluating the model with the test set.
     loss, accuracy = coordinator.evaluate(xy_test)
 
-    # Print final test loss and accuracy
+    # Printing final test loss and accuracy.
     print(f"---\nTest completed!\nLoss: {loss} | Test accuracy: {accuracy}")
 
 
 def learning_rate_fn(epoch_base, lr_initial=0.002, k=0.01):
```
danieljanes: Add mypy type annotation where it cannot be inferred, i.e. `epoch_base`.

danieljanes: (this also hold for other parts of the example module)

examples/tensorflow_hello_world/helloworld.py
wilk10: uhm i think it's important the user can familiarise herself with the federated hyper-parameters and has a centralised place where to set those parameters, instead of inline.

wilk10:
thanks!

PR #162, opened 2019-11-26
Hackaton code comments

r-marques:
This pull request is just to leave some remarks about some changes that were done during the hackaton and that could be used for future improvements.

Most of the code changes are related to the python package rename.

The relevant files changed during the hackaton are:
- [`xain_fl/grpc/coordinator.py`](https://github.com/xainag/xain/pull/162/files#diff-882035fa169b165ae3b7ebc59c338be1)
- `xain_fl/grpc/participant.py`


xain_fl/grpc/coordinator.py
```diff
@@ -656,26 +743,36 @@ def monitor_heartbeats(
...
     events and threads and configures and starts the gRPC service.
     """
-    coordinator = Coordinator()
     terminate_event = threading.Event()
     monitor_thread = threading.Thread(
         target=monitor_heartbeats, args=(coordinator, terminate_event)
     )
 
-    server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
+    options = [
```
r-marques: These options remove the size limit of sent and received messages. Right now the reason is for the messages that send numpy arrays. In the future the numpy arrays may not be shared through gRPC or if they are we need to choose and bound the maximum message size.

xain_fl/grpc/participant.py
```diff
@@ -0,0 +1,377 @@
...
+    """
+    coord = coordinator_pb2_grpc.CoordinatorStub(chan)
+    while not terminate.is_set():
+        start = datetime.today()
+        print(f"sending heartbeat {start}")
+        if st.state in [ParState.WAITING_FOR_SELECTION, ParState.POST_TRAINING, ParState.DONE]:
+            state = coordinator_pb2.State.Value("READY")
+        else:
+            state = coordinator_pb2.State.Value("TRAINING")
+        req = coordinator_pb2.HeartbeatRequest(state=state, round=st.round)
```
r-marques: Send state and current round in heartbeat request

xain_fl/grpc/participant.py
```diff
@@ -0,0 +1,377 @@
...
+    Args:
+        chan: gRPC channel to Coordinator.
+        st (obj:`StateRecord`): Participant state record.
+        terminate (obj:`threading.Event`): Event to terminate message loop.
+    """
+    coord = coordinator_pb2_grpc.CoordinatorStub(chan)
+    while not terminate.is_set():
+        start = datetime.today()
+        print(f"sending heartbeat {start}")
+        if st.state in [ParState.WAITING_FOR_SELECTION, ParState.POST_TRAINING, ParState.DONE]:
```
r-marques: Ideally these states would be the same as defined in the proto file

xain_fl/grpc/participant.py
```diff
@@ -0,0 +1,377 @@
...
+            state = coordinator_pb2.State.Value("TRAINING")
+        req = coordinator_pb2.HeartbeatRequest(state=state, round=st.round)
+        reply = coord.Heartbeat(req)
+        transit(st, reply)
+        end = datetime.today() - start
+        print(f"ended sending heartbeat {end}")
+        time.sleep(HEARTBEAT_TIME)
+
+
+def go(part: Participant, coordinator_address: str):
```
r-marques: Add the ability to pass in a coordinator address instead of using an harcoded one.

xain_fl/grpc/participant.py
```diff
@@ -0,0 +1,377 @@
...
+    WAITING_FOR_SELECTION. When selected, it moves to TRAINING followed by
+    POST_TRAINING. If selected again for the next round, it moves back to
+    TRAINING, otherwise it is back to WAITING_FOR_SELECTION.
+
+    Args:
+        part (obj:`Participant`): Participant object for training computation.
+    """
+
+    # channel options
+    options = [
```
r-marques: These options remove the size limit of sent and received messages. Right now the reason is for the messages that send numpy arrays. In the future the numpy arrays may not be shared through gRPC or if they are we need to choose and bound the maximum message size.

xain_fl/grpc/coordinator.py
finiteprods: Will add a task to keep this on the radar.

xain_fl/grpc/participant.py
finiteprods: Should be addressed in the update to Participant state machine.

xain_fl/grpc/participant.py
finiteprods: Yes. But it's not currently used by the coordinator, right?

r-marques: No. In the hackaton this was only used to print the metrics.

xain_fl/grpc/participant.py
r-marques: We should also create a task for that

PR #163, opened 2019-11-28
Replace all print and abseil logs with xain.logger

tanertopal:


tanertopal:
Awaiting merge of #129 so I can make final changes before this can be merged. Will be reopened afterwards.

xain/config.py
```diff
@@ -17,10 +16,12 @@
 output_dir_default = root_dir.joinpath("output")
 results_dir_default = root_dir.joinpath("results")
 
+logger = get_logger(__name__, level=os.environ.get("XAIN_LOGLEVEL", "INFO"))
+
 
 def init_config():
     """Creates initial config file if non exists"""
-    logging.info(f"Initializing config in {config_file}")
+    logger.info("Initializing config in %s", config_file)
```
finiteprods: what's wrong with a f-string?

tanertopal: The linter does not like it (=> `W1202 - logging-fstring-interpolation` warning). I personally would prefer f-string.

PR #172, opened 2019-12-04
XP-248 - Use structlog for logging

tanertopal:
### References
Resolves XP-248

### Summary
Replaces current logging mechanism with structlog

### DIFF
--- a/xain_fl/config.py
+++ b/xain_fl/config.py
@@ -21,7 +21,7 @@ logger = get_logger(__name__, level=os.environ.get("XAIN_LOGLEVEL", "INFO"))
 
 def init_config():
     """Creates initial config file if non exists"""
-    logger.info(f"Initializing config in {config_file}")
+    logger.info("Initializing config", config_file=config_file)
 
     config = configparser.ConfigParser()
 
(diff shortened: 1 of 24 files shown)


atymoshchuk:
@tanertopal Please use full github template for every pull request.

xain_fl/config.py
```diff
@@ -21,7 +21,7 @@
 
 def init_config():
     """Creates initial config file if non exists"""
-    logger.info("Initializing config in %s", config_file)
+    logger.info(f"Initializing config in {config_file}")
```
atymoshchuk: Please do not use f-strings in logs, it's a rule in python, according to https://www.python.org/dev/peps/pep-0498/

xain_fl/datasets/storage.py
```diff
@@ -55,7 +55,7 @@ def fetch_ndarray(url, fpath):
     """
     r = requests.get(url, stream=True)
 
-    logger.info("Fetching file %s", url)
+    logger.info(f"Fetching file {url}")
```
atymoshchuk: benefit of structlog is that it allows to use variables key=value in logs:
`logger.info("Fetching file", url=str(url))`
Please adjust log messages accordingly.

tanertopal:
Closing as the branch name ist not accepted. Will open another PR.

PR #208, opened 2020-01-07
Do 43 docker compose minio

dkravetz:
### References

[DO-43](https://xainag.atlassian.net/browse/DO-43) [DO-26](https://xainag.atlassian.net/browse/DO-26) 

### Summary

This PR adds a docker-compose for development and release that spins up minio and xain-fl together.

No.

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [ ] Passed scope checks.


README.md
```diff
@@ -79,13 +79,15 @@ $ python xain_fl/cli.py --help
...
+There are two docker-compose files, one for development and one for release.
+
 Development image
 ---
 
 To run the coordinator's development image, first build the Docker image:
 
 ```shell
-$ docker build -t xain-fl-dev -f dev.dockerfile .
+$ docker build -t xain-fl-dev -f Dockerfile.dev .
```
dkravetz: Renamed the Dockerfile to follow the same convention as the MinIO project.

README.md
```diff
@@ -110,6 +112,22 @@ And then run it (this example assumes you'll want to use the default port):
...
+
+```shell
+$ docker-compose -f docker-compose-dev.yml up
+```
+
+Release
+---
+
+```shell
+$ docker-compose up
```
Robert-Steiner: The release command does not work on my machine.
I get:
```
xain-fl_1  | xain_fl-0.2.0
xain-fl_xain-fl_1 exited with code 0
```

docker-compose-dev.yml
```diff
@@ -0,0 +1,40 @@
...
+    ports:
+      - "9000:9000"
+
+  xain-fl:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
+    command: coordinator -f test_array.npy
+    volumes:
+      - '.:/app'
```
little-dude: Do we need the whole repo to be mounted? The problem with that is that the container's `root` user might end up owning some files if it writes to them.

README.md
dkravetz: looks like you have another image with the same name (probably from building directly from the Dockerfile?). Could you try again after removing the containers called `xain-fl`, `xain-fl_xain-fl_1` ?

docker-compose-dev.yml
dkravetz: We need two files that are at the root of the repo (README.md and test_array.npy) in order for pip to be able to install the packages and the coordinator to start, respectively. We can only volume-mount directories, thus, the root of the repo. 

We could leave this as-is, I could update the Dockerfile to use a custom username, where it makes sense, or we could investigate into tools like https://podman.io/

little-dude: > We can only volume-mount directories, thus, the root of the repo.

I think that we can actually mount individual files as well. This [SO answer](https://stackoverflow.com/a/42260979/1836144) suggests the only restriction is that you need to use an absolute path and that the file has to exist in the image (which can be done by `touch`ing it in the Dockerfile).

Anyway that's not a blocker.

README.md
Robert-Steiner: Ok it works after I have removed the images.

Robert-Steiner: But now I get an error when run `docker-compose -f docker-compose-dev.yml up`

can we rename the `xain-fl` service in docker-compose-dev.yml to `xain-fl-dev`?

dkravetz: Done! 
Renamed minio and the networks with the same prefix, to have a clearer separation between the two

docker-compose-dev.yml
dkravetz: You are correct. 
Updated the volume mounts to only include necessary files.

docker-compose-dev.yml
```diff
@@ -19,22 +19,25 @@ services:
...
+  xain-fl-dev:
     build:
       context: .
       dockerfile: Dockerfile.dev
     command: coordinator -f test_array.npy
     volumes:
-      - '.:/app'
       #  don't use the local egg-info, if one exists
-      - '/app/xain_fl.egg-info'
+      - /app/xain_fl.egg-info
```
little-dude: What does it do when there's no `:` ? Docker mounts the local `xain_fl.egg-info` if it exists and does nothing otherwise?

dkravetz: The engine handles the mount, this is only to explicitly don't use the local `egg-info`, even if you mount the entire directory

PR #214, opened 2020-01-10
Xp 373 add sdk as dependency in fl

wilk10:
### References

[XP-373](https://xainag.atlassian.net/browse/XP-373)

### Summary

Add link to SDK repo inside `setup.py` and reactivates the tests that were disabled both by the comments and by the `@pytest.mark.skip` decorator.

 I had to refactor a bit `test_end_training` test. The refactor addresses two things:
1. renaming variables to make them more readable
2. i changed `np.testing.assert_allclose` to `np.testing.assert_equal` because it was failing with a numpy error. 
**My guess** is that since this test has been skipped for a while, the fact that in the meantime we changed numpy version affected this test. So had this test not been skipped when the numpy version was changed, it would have been picked up back then (just a guess on the reason now it was failing, not a blame). The slight change i made to the test applies a condition that is more restrictive, so i think it's conceptually better than before. Hope it's OK with the reviewers!

This should unblock a bunch of tickets, but no blockers for this one.

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [x] Passed scope checks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,13 +41,13 @@ dev_require = [
     "twine==2.0.0",  # Apache License 2.0
     "wheel==0.33.6",  # MIT
     "typing-extensions==3.7.4.1",  # PSF
-    "xain-sdk @ git+https://github.com/xainag/xain-sdk.git@development#egg=xain_sdk-0.1.0",
 ]
 
 tests_require = [
     "pytest==5.3.2",  # MIT license
     "pytest-cov==2.8.1",  # MIT
     "pytest-watch==4.2.0",  # MIT
+    "xain-sdk @ git+https://github.com/xainag/xain-sdk.git@development#egg=xain_sdk-0.1.0",
 ]
 
 docs_require = ["Sphinx==2.2.1", "m2r==0.2.1", "sphinxcontrib-mermaid==0.3.1"]
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ dev_require = [
     "twine==2.0.0",  # Apache License 2.0
     "wheel==0.33.6",  # MIT
     "typing-extensions==3.7.4.1",  # PSF
+    "xain-sdk @ git+https://github.com/xainag/xain-sdk.git@development#egg=xain_sdk-0.1.0",
 ]
 
 tests_require = [
--- a/tests/test_grpc.py
+++ b/tests/test_grpc.py
@@ -29,6 +29,8 @@ from xain_fl.coordinator.coordinator_grpc import CoordinatorGrpc
 from xain_fl.coordinator.heartbeat import monitor_heartbeats
 from xain_fl.coordinator.participants import Participants
 
+
+# TODO: it should be fixed with: https://xainag.atlassian.net/browse/XP-119
 # Some grpc tests fail on macos.
 # `pytestmark` when defined on a module will mark all tests in that module.
 # For more information check
@@ -315,14 +317,14 @@ def test_end_training(coordinator_service):
 
     # simulate trained local model data
     test_weights, number_samples = [np.arange(20, 30), np.arange(30, 40)], 2
-    metrics = {"metric": [np.arange(10, 20), np.arange(5, 10)]}
+    test_metrics = {"metric": [np.arange(10, 20), np.arange(5, 10)]}
 
     with grpc.insecure_channel("localhost:50051") as channel:
         # we first need to rendezvous before we can send any other request
         rendezvous(channel)
         # call endTraining service method on coordinator
         end_training(  # pylint: disable-msg=no-value-for-parameter
-            channel, test_weights, number_samples, metrics
+            channel, test_weights, number_samples, test_metrics
         )
     # check local model received...
 
@@ -331,15 +333,15 @@ def test_end_training(coordinator_service):
     round_ = coordinator_service.coordinator.round
 
     # first the weights update
-    _, update = round_.updates.popitem()
-    tu1, tu2 = update["weight_update"]
-    assert tu2 == number_samples
-    np.testing.assert_equal(tu1, test_weights)
+    _, round_update = round_.updates.popitem()
+    update_item1, update_item2 = round_update["weight_update"]
+    assert update_item2 == number_samples
+    np.testing.assert_equal(update_item1, test_weights)
 
-    met = update["metrics"]
-    assert met.keys() == metrics.keys()
-    for k, vals in metrics.items():
-        np.testing.assert_allclose(met[k], vals)
+    round_update_metrics = round_update["metrics"]
+    assert round_update_metrics.keys() == test_metrics.keys()
+    for key, values in test_metrics.items():
+        np.testing.assert_equal(round_update_metrics[key], values)
 
 
 @pytest.mark.integration
--- a/tests/test_grpc.py
+++ b/tests/test_grpc.py
@@ -1,9 +1,5 @@
 """XAIN FL tests for gRPC coordinator"""
 
-# TODO: https://xainag.atlassian.net/browse/XP-241 will break this test
-# TODO: https://xainag.atlassian.net/browse/XP-373 will fix it again
-# (please bear with us in the meantime)
-
 from concurrent import futures
 import sys
 import threading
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -41,6 +41,7 @@
     "twine==2.0.0",  # Apache License 2.0
     "wheel==0.33.6",  # MIT
     "typing-extensions==3.7.4.1",  # PSF
+    "xain-sdk @ git+https://github.com/xainag/xain-sdk.git@development#egg=xain_sdk-0.1.0",
```
Robert-Steiner: If we need the xain-sdk dependency for tests only, we should move it to `tests_require`.

tests/test_grpc.py
```diff
@@ -308,47 +303,45 @@ def test_start_training_failed_precondition(  # pylint: disable=unused-argument
...
+    # first the weights update
+    _, round_update = round_.updates.popitem()
+    update_item1, update_item2 = round_update["weight_update"]
+    assert update_item2 == number_samples
+    np.testing.assert_equal(update_item1, test_weights)
+
+    round_update_metrics = round_update["metrics"]
+    assert round_update_metrics.keys() == test_metrics.keys()
+    for key, values in test_metrics.items():
+        np.testing.assert_equal(round_update_metrics[key], values)
```
finiteprods: back when this test was first written, `assert_allclose` was used here (rather than `assert_equal`) since we're comparing floats. I understand that we have since used a more up to date `numpy` which may be the reason it fails but nevertheless it seems odd that it does!

setup.py
wilk10: sounds good, i'll update the JIRA ticket too

tests/test_grpc.py
wilk10: this is the error `TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`. actually i think we went back to numpy=1.15 from 1.16 (but the fact that it's numpy related it's guess, i haven't investigated it with 1.16 . i can investigate if you want, try to update the 1.16 and see what happens).

`________________________________________________ test_end_training _________________________________________________

coordinator_service = <xain_fl.coordinator.coordinator_grpc.CoordinatorGrpc object at 0x1126b7518>

    @pytest.mark.integration
    def test_end_training(coordinator_service):
        """[summary]

        .. todo:: Advance docstrings (https://xainag.atlassian.net/browse/XP-425)

        Args:
            coordinator_service ([type]): [description]
        """

        assert coordinator_service.coordinator.round.updates == {}

        # simulate trained local model data
        test_weights, number_samples = [np.arange(20, 30), np.arange(30, 40)], 2
        test_metrics = {"metric": [np.arange(10, 20), np.arange(5, 10)]}

        with grpc.insecure_channel("localhost:50051") as channel:
            # we first need to rendezvous before we can send any other request
            rendezvous(channel)
            # call endTraining service method on coordinator
            end_training(  # pylint: disable-msg=no-value-for-parameter
                channel, test_weights, number_samples, test_metrics
            )
        # check local model received...

        assert len(coordinator_service.coordinator.round.updates) == 1

        round_ = coordinator_service.coordinator.round

        # first the weights update
        _, round_update = round_.updates.popitem()
        update_item1, update_item2 = round_update["weight_update"]
        assert update_item2 == number_samples
        np.testing.assert_equal(update_item1, test_weights)

        round_update_metrics = round_update["metrics"]
        assert round_update_metrics.keys() == test_metrics.keys()
        for key, values in test_metrics.items():
>           np.testing.assert_allclose(round_update_metrics[key], values)

tests/test_grpc.py:344:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../environments/fl369/lib/python3.7/site-packages/numpy/testing/_private/utils.py:1438: in compare
    equal_nan=equal_nan)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = array([array([10, 11, 12, 13, 14, 15, 16, 17, 18, 19]),
       array([5, 6, 7, 8, 9])], dtype=object)
b = array([array([10, 11, 12, 13, 14, 15, 16, 17, 18, 19]),
       array([5, 6, 7, 8, 9])], dtype=object)
rtol = 1e-07, atol = 0, equal_nan = True

    def isclose(a, b, rtol=1.e-5, atol=1.e-8, equal_nan=False):
        """
        Returns a boolean array where two arrays are element-wise equal within a
        tolerance.

        The tolerance values are positive, typically very small numbers.  The
        relative difference (`rtol` * abs(`b`)) and the absolute difference
        `atol` are added together to compare against the absolute difference
        between `a` and `b`.

        .. warning:: The default `atol` is not appropriate for comparing numbers
                     that are much smaller than one (see Notes).

        Parameters
        ----------
        a, b : array_like
            Input arrays to compare.
        rtol : float
            The relative tolerance parameter (see Notes).
        atol : float
            The absolute tolerance parameter (see Notes).
        equal_nan : bool
            Whether to compare NaN's as equal.  If True, NaN's in `a` will be
            considered equal to NaN's in `b` in the output array.

        Returns
        -------
        y : array_like
            Returns a boolean array of where `a` and `b` are equal within the
            given tolerance. If both `a` and `b` are scalars, returns a single
            boolean value.

        See Also
        --------
        allclose

        Notes
        -----
        .. versionadded:: 1.7.0

        For finite values, isclose uses the following equation to test whether
        two floating point values are equivalent.

         absolute(`a` - `b`) <= (`atol` + `rtol` * absolute(`b`))

        Unlike the built-in `math.isclose`, the above equation is not symmetric
        in `a` and `b` -- it assumes `b` is the reference value -- so that
        `isclose(a, b)` might be different from `isclose(b, a)`. Furthermore,
        the default value of atol is not zero, and is used to determine what
        small values should be considered close to zero. The default value is
        appropriate for expected values of order unity: if the expected values
        are significantly smaller than one, it can result in false positives.
        `atol` should be carefully selected for the use case at hand. A zero value
        for `atol` will result in `False` if either `a` or `b` is zero.

        Examples
        --------
        >>> np.isclose([1e10,1e-7], [1.00001e10,1e-8])
        array([True, False])
        >>> np.isclose([1e10,1e-8], [1.00001e10,1e-9])
        array([True, True])
        >>> np.isclose([1e10,1e-8], [1.0001e10,1e-9])
        array([False, True])
        >>> np.isclose([1.0, np.nan], [1.0, np.nan])
        array([True, False])
        >>> np.isclose([1.0, np.nan], [1.0, np.nan], equal_nan=True)
        array([True, True])
        >>> np.isclose([1e-8, 1e-7], [0.0, 0.0])
        array([ True, False], dtype=bool)
        >>> np.isclose([1e-100, 1e-7], [0.0, 0.0], atol=0.0)
        array([False, False], dtype=bool)
        >>> np.isclose([1e-10, 1e-10], [1e-20, 0.0])
        array([ True,  True], dtype=bool)
        >>> np.isclose([1e-10, 1e-10], [1e-20, 0.999999e-10], atol=0.0)
        array([False,  True], dtype=bool)
        """
        def within_tol(x, y, atol, rtol):
            with errstate(invalid='ignore'):
                return less_equal(abs(x-y), atol + rtol * abs(y))

        x = asanyarray(a)
        y = asanyarray(b)

        # Make sure y is an inexact type to avoid bad behavior on abs(MIN_INT).
        # This will cause casting of x later. Also, make sure to allow subclasses
        # (e.g., for numpy.ma).
        dt = multiarray.result_type(y, 1.)
        y = array(y, dtype=dt, copy=False, subok=True)

>       xfin = isfinite(x)
E       TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

../environments/fl369/lib/python3.7/site-packages/numpy/core/numeric.py:2363: TypeError`

finiteprods: for posterity: we did a quick check and found that we still get the above error with `assert_allclose` even when reverted back to `numpy==1.16`. So the underlying reason for this change of behaviour seems to be due to something else. In any case, the proposed change to use a more standard assert is fine.

PR #215, opened 2020-01-10
XP-308: store aggregated weights in S3 buckets

little-dude:
### References

https://xainag.atlassian.net/browse/XP-40
https://xainag.atlassian.net/browse/XP-308
https://xainag.atlassian.net/browse/XP-309
https://xainag.atlassian.net/browse/XP-311

## `xain-fl` changes

This commits adds a new module `xain-fl.coordinator.store`, which
contains a `Store()` class that can read from and write to S3
buckets. At the end of each round, the aggregated weights are uploaded
to an S3 bucket. Note that as a first step, the aggregated weights are
still sent to the participants via gRPC.

##### Implementation notes:

We're trying to not have IO related code in the Coordinator class, so
the Store instance is held by the CoordinatorGrpc instead of
Coordinator. However, since the Coordinator is currently generating
the grpc messages, which are inherently coupled to the IO logic. This
means that if at some point the messages need to contain information
related to storage, we'll end up with storage logic in the
Coordinator, and things might become pretty messy. Thus, we should
already start thinking about refactoring xain-fl.coordinator such that
the state machine is completely decoupled for the IO logic.

## Tests changes

In order to test the store, we needed to go through a valid round, since weights are only stored at the end of a round after the weights have been aggregated. This was initially not possible because in the unit tests, all the participant stubs used the same TCP connection, so they were all identified by the coordinator as the same participant. To work around this, we implemented a simple TCP port forwarding mechanism so that we can create a new TCP connection for each stub.

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [x] Passed scope checks.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -10,7 +10,8 @@ services:
     environment:
       MINIO_ACCESS_KEY: minio
       MINIO_SECRET_KEY: minio123
-    command: server /data
+    command:
+        server /data
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
       interval: 30s
@@ -47,10 +48,13 @@ services:
       "
 
   xain-fl-dev:
+    environment:
+      MINIO_ACCESS_KEY: minio
+      MINIO_SECRET_KEY: minio123
     build:
       context: .
       dockerfile: Dockerfile.dev
-    command: coordinator -f test_array.npy
+    command: sh -c "coordinator -f test_array.npy --storage-endpoint http://minio-dev:9000 --storage-key-id $${MINIO_ACCESS_KEY} --storage-secret-access-key $${MINIO_SECRET_KEY} --storage-bucket xain-fl-aggregated-weights"
     volumes:
       #  don't use the local egg-info, if one exists
       - /app/xain_fl.egg-info
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -6,5 +6,5 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd $DIR/../
 
 set -x
-isort --indent=4 -rc setup.py xain_fl
-black --line-length 100 setup.py xain_fl
+isort --indent=4 -rc setup.py xain_fl tests
+black --line-length 100 setup.py xain_fl tests
--- a/tests/port_forwarding.py
+++ b/tests/port_forwarding.py
@@ -165,6 +165,7 @@ class ConnectionManager:
         terminate_event = Event()
         thread = Thread(
             target=forward,
+            name=f"{host}:{port}<->{target_host}:{target_port}",
             args=(host, port, target_host, target_port, terminate_event),
             daemon=True,
         )
@@ -195,5 +196,10 @@ class ConnectionManager:
             threads.append(thread)
         for thread in threads:
             thread.join(timeout=1)
-            assert not thread.isAlive()
+            # FIXME: sometimes some threads don't terminate in CI, so
+            # we're dumping the stacktraces for all the current
+            # threads for debugging purpose.
+            if thread.isAlive():
+                faulthandler.dump_traceback()
+                raise Exception(f"Thread {thread.name} is still alive")
         self.forwarders = {}
--- a/tests/port_forwarding.py
+++ b/tests/port_forwarding.py
@@ -1,0 +1,199 @@
+"""
+This module provides logic to perform port forwarding. We use this to
+artificially open new TCP connections when creating gRPC clients.
+"""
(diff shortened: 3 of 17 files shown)


little-dude:
Just realized `pylint` was not running for the `tests` package, and the new code doesn't pass...

scripts/format.sh
```diff
@@ -6,5 +6,5 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd $DIR/../
 
 set -x
-isort --indent=4 -rc setup.py xain_fl
-black --line-length 100 setup.py xain_fl
+isort --indent=4 -rc setup.py xain_fl tests
```
Robert-Steiner: There is one small thing missing. We also need to enable it in the CI:
* `.github/workflows/development.yml`
* `.github/workflows/master.yml`

docker-compose-dev.yml
```diff
@@ -50,7 +51,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile.dev
-    command: coordinator -f test_array.npy
+    command: coordinator -f test_array.npy --storage-endpoint http://minio-dev:9000 --storage-key-id $${MINIO_ACCESS_KEY} --storage-secret-access-key $${MINIO_SECRET_KEY} --storage-bucket xain-fl-aggregated-weights
```
Robert-Steiner: During the build of the docker images I get the following error:
```
Running command git clone -q https://github.com/xainag/xain-proto.git /tmp/pip-install-jvk79hx_/xain-proto
ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?
```

xain_fl/cli.py
```diff
@@ -197,7 +222,15 @@ def main():
...
     )
 
-    serve(coordinator=coordinator, host=parameters.host, port=parameters.port)
+    store_config = StoreConfig(
+        parameters.storage_endpoint,
+        parameters.storage_key_id,
+        parameters.storage_secret_access_key,
+        parameters.storage_bucket,
+    )
+    store = Store(store_config)
```
Robert-Steiner: Maybe it makes sense to create a in-memory Store in a later task and use it as the default store. I just want to make sure that it is still possible to run the coordinator without the requirement of minio. What do you think @atymoshchuk, @little-dude, @wilk10?

little-dude: Agree. I was thinking about the same thing while writing the tests.

scripts/format.sh
little-dude: thanks, done!

docker-compose-dev.yml
little-dude: It should be fixed now, could you give it another try?

little-dude: I also made `pip install` verbose, otherwise it may seem like the command is hanging.

little-dude: cc @dkravetz  for the docker files changes.

Robert-Steiner: There is something not right with `$${MINIO_SECRET_KEY}` and `$${MINIO_ACCESS_KEY}`. 
I get: 

```
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.UNKNOWN
	details = "Exception calling application: An error occurred (InvalidAccessKeyId) when calling the PutObject operation: The access key ID you provided does not exist in our records."
	debug_error_string = "{"created":"@1579007385.100986000","description":"Error received from peer ipv4:127.0.0.1:50051","file":"src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"Exception calling application: An error occurred (InvalidAccessKeyId) when calling the PutObject operation: The access key ID you provided does not exist in our records.","grpc_status":2}"
```

But it works when I use the hardcoded values `minio` and `minio123`

little-dude: oh I actually simply forgot to add the `environment` section that set those variables...

xain_fl/cli.py
Robert-Steiner: I created a ticket [XP-435](https://xainag.atlassian.net/browse/XP-435).

tests/port_forwarding.py
```diff
@@ -0,0 +1,191 @@
...
+            - returns
+    """
+    # Set up a server socket that will wait for incoming connections
+    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    server.bind((host, port))
+    server.listen(1)
+
+    # Block until the first client connects
+    client_conn, _ = server.accept()
```
little-dude: This is blocking so if not client connects, this thread never terminates...

little-dude:
I think I've fixed the spurious CI failures, so do you mind re-approving @Robert-Steiner ?

PR #219, opened 2020-01-15
XP-436 Reinstate FINISHED heartbeat from Coordinator

finiteprods:
### References

- [XP-436](https://xainag.atlassian.net/browse/XP-436)

### Summary

This bug appears to have been introduced as part of [Participant selection](https://github.com/xainag/xain-fl/pull/179) whereby the Coordinator no longer advertises the `FINISHED` state when the whole training session is over. This is in turn a problem for Participants: they end up waiting indefinitely (in the `POST_TRAINING` state), unable to terminate. The bug was discovered as part of [XP-208](https://xainag.atlassian.net/browse/XP-208) which was previously blocked by this.

In a little more detail, this PR:

1. rectifies the above bug so that `FINISHED` states are indeed advertised.
2. adds an integration test exercising the entire lifecycle of a Participant (including termination - which was previously lacking), communicating with a Coordinator.
3. as part of 2., adds simple Aggregator and Controller implementations for purposes of mocking.

- None, but this complements [extended test coverage of Participant](https://github.com/xainag/xain-sdk/pull/30).

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [x] Passed scope checks.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,9 @@ from xain_fl.coordinator.coordinator import Coordinator
 from xain_fl.coordinator.coordinator_grpc import CoordinatorGrpc
 from xain_fl.helloproto.numproto_server import NumProtoServer
 
+from .port_forwarding import ConnectionManager
+from .store import TestStore
+
 
 @pytest.fixture
 def greeter_server():
@@ -35,7 +38,8 @@ def coordinator_service():
 
     server = grpc.server(futures.ThreadPoolExecutor(max_workers=1))
     coordinator = Coordinator(minimum_participants_in_round=10, fraction_of_participants=1.0)
-    coordinator_grpc = CoordinatorGrpc(coordinator)
+    store = TestStore()
+    coordinator_grpc = CoordinatorGrpc(coordinator, store)
     coordinator_pb2_grpc.add_CoordinatorServicer_to_server(coordinator_grpc, server)
     server.add_insecure_port("localhost:50051")
     server.start()
@@ -59,3 +63,41 @@ def participant_stub():
     yield stub
 
     channel.close()
+
+
+def port_generator():
+    """A generator that yields incrementing port numbers.
+
+    """
+    port = 50052
+    while True:
+        yield port
+        port += 1
+
+
+@pytest.fixture
+def participant_stubs():
+    """Generator that yields functions instantiate participant stubs.
+    Each participant creates a new TCP connection, so that they get a
+    different participant ID.
+
+    """
+
+    ports = port_generator()
+    connections = ConnectionManager()
+    channels = []
+
+    def generate_participant_stubs():
+        for port in ports:
+            connections.start("localhost", port, "localhost", 50051)
+            channel = grpc.insecure_channel(f"localhost:{port}")
+            channels.append(channel)
+            stub = coordinator_pb2_grpc.CoordinatorStub(channel)
+            yield stub
+
+    yield generate_participant_stubs()
+
+    for channel in channels:
+        channel.close()
+
+    connections.stop_all()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,8 +2,8 @@
 
 from concurrent import futures
 import threading
-
 from unittest import mock
+
 import grpc
 import pytest
 from xain_proto.fl import coordinator_pb2_grpc, hellonumproto_pb2_grpc
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 """XAIN FL conftest for cproto"""
 
 from concurrent import futures
+import threading
 
 import grpc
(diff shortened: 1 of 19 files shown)


tests/conftest.py
```diff
@@ -43,6 +47,37 @@ def coordinator_service():
...
+        fraction_of_participants=1.0,
+        aggregator=agg,
+        controller=ctrl,
+    )
+    coordinator_grpc = CoordinatorGrpc(coordinator)
+    coordinator_pb2_grpc.add_CoordinatorServicer_to_server(coordinator_grpc, server)
+    server.add_insecure_port("localhost:50051")
+    server.start()
+    terminate_event = threading.Event()
+    monitor_thread = threading.Thread(
```
little-dude: Maybe we need to `join()` this thread, just to make sure it terminated?

little-dude:
> adds simple Aggregator and Controller implementations for purposes of mocking

It seems these are needed only in the tests, so unless we plan to use them in `xain-fl`, I'd be in favor of moving them to the `tests` package.

tests/test_grpc.py
```diff
@@ -373,3 +374,31 @@ def test_end_training_denied(  # pylint: disable=unused-argument
...
+
+    Args:
+        mock_coordinator_service ([type]): [description]
+    """
+
+    init_weight = [np.arange(10), np.arange(10, 20)]
+    mock_coordinator_service.coordinator.weights = init_weight
+
+    # mock a local participant with a constant train_round function
+    with mock.patch("xain_sdk.participant.Participant") as mock_obj:
```
janpetschexain: the object should be mocked where it is called instead of where it is defined, that means
`with mock.patch("xain_sdk.participant_state_machine.Participant") as mock_obj:`

tests/conftest.py
finiteprods: yes, to confirm the expected response of setting `terminate_event`.

tests/test_grpc.py
finiteprods: I didn't realise that. Is it just coincidence that "xain_sdk.participant.Participant" works just as well?

tests/test_grpc.py
```diff
@@ -373,3 +374,31 @@ def test_end_training_denied(  # pylint: disable=unused-argument
...
+        mock_local_part = mock_obj.return_value
+        mock_local_part.train_round.return_value = init_weight, 1, {}
+
+        start_participant(mock_local_part, "localhost:50051")
```
wilk10: I have pulled your branch and run this test. `start_participant` is pretty slow, can we please mark this test as slow?

tests/test_grpc.py
```diff
@@ -373,3 +374,31 @@ def test_end_training_denied(  # pylint: disable=unused-argument
...
+
+        coord = mock_coordinator_service.coordinator
+        assert coord.state == coordinator_pb2.State.FINISHED
+        # coordinator set to 2 round for good measure, but the resulting
+        # aggregated weights are the same as a single round
+        assert coord.current_round == 2
+        np.testing.assert_equal(coord.weights, [np.arange(start=10, stop=29, step=2)])
```
wilk10: maybe can you please specify in the comment that we expect the initial weights to have been summed together? it was not extra obvious

little-dude:
> > > adds simple Aggregator and Controller implementations for purposes of mocking
> > 
> > 
> > It seems these are needed only in the tests, so unless we plan to use them in `xain-fl`, I'd be in favor of moving them to the `tests` package.
> 
> true that they're only used in the tests but e.g. `aggregate.py` also contains `IdentityAgg` which isn't used anywhere other than maybe testing, so the decision to have them there was just "following precedent".

IMO we should avoid having dead code in `xain-fl`, but I won't block the PR on this. I'll create a follow-up ticket for this.

tests/test_grpc.py
finiteprods: good point

tests/test_grpc.py
janpetschexain: i think in many cases it works coincidentally.

but it can also go wrong, yielding errors which look pretty strange on the first look, e.g. i had to change sth in the xain-proto repo in one of the last tickets, then merged the changes but couldn't publish it on pypi. ofcourse i still needed the changes in xain-proto for the followup tickets for xain-fl/sdk, therefor i adjusted their setup.py to reference to the dev-branch instead of the pypi package. so far so good, but then the mocking of one previously passing test failed, because it pointed to the xain-proto source for mocking instead of the xain-sdk's imported usage of the xain-proto stuff. long story short, as far as i have seen, it can happen that mocking the source can point to the "wrong" source, therefor the mocking happened at the source, but had no effect at the actual execution and the test/mocking failed.

finiteprods: thanks for sharing this insight!

PR #222, opened 2020-01-16
XP-480 revise message names

janpetschexain:
### References

- [XP-480](https://xainag.atlassian.net/browse/XP-480)

### Summary

- adjust the variable names and imports related to protobuf definitions according to the xain-proto changes

- none

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [ ] Passed scope checks.


.isort.cfg
```diff
@@ -1,7 +1,8 @@
 [settings]
-multi_line_output=3
-include_trailing_comma=True
 force_grid_wrap=0
-use_parentheses=True
-line_length=100
 force_sort_within_sections=True
+include_trailing_comma=True
+known_third_party=xain_proto,xain_sdk
```
finiteprods: just for reference, can you just comment what the problem you had with `isort` was? did this fix it?

docs/network_architecture.rst
```diff
@@ -192,7 +192,7 @@ either:
 
         # rest of the method logic
 
-*Periodically check if a the rcp call is active:*
+*Periodically check if the rcp call is active:*
```
finiteprods:
```suggestion
*Periodically check if the rpc call is active:*
```

docs/network_architecture.rst
```diff
@@ -376,14 +372,13 @@ this conversion.
...
 :math:`P` essentially kicks off the exchange. :math:`C` responds with the global
-model :math:`\theta` (and other data as specified in
-:code:`StartTrainingReply`). Then :math:`P` carries out the training locally.
-When complete, it sends the updated model :math:`\theta'` (and other metadata)
+model :math:`\weights` (and other data as specified in
+:code:`StartTrainingRoundResponse`). Then :math:`P` carries out the training locally.
+When complete, it sends the updated model :math:`\weights'` (and other metadata)
 back. :math:`C` responds with an acknowledgement.
 
 .. image:: _static/sequence2.png
```
finiteprods: ideally the sequence diagram should be updated with the new names too. It's unfortunate that the Sphinx markup doesn't support this kind of diagram (or it does IIRC but looks quite ugly in comparison)... I can help out with updating the image if you want.

tests/test_coordinator.py
```diff
@@ -77,15 +84,15 @@ def test_coordinator_state_standby_round():
     # are connected
     coordinator = Coordinator(minimum_participants_in_round=1, fraction_of_participants=1.0)
 
-    assert coordinator.state == coordinator_pb2.STANDBY
```
finiteprods: just a remark: I'm a little surprised that in the old version, `coordinator_pb2.STANDBY` seems to work just as well as `coordinator_pb2.State.STANDBY`...

.isort.cfg
janpetschexain: my ide is set up such that it can find my local repos via the `PYTHONPATH` env (basically pointing python to the local xain-fl, xain-sdk, etc). it seems that pip is looking up that env as well next to the actual installed packages. therefor isort then thought that e.g. xain-sdk is a local dependency instead of a 3rd party dependency, which conflicted with the ci settings.

pointing out xain-sdk and xain-proto as known 3rd party dependencies to isort, the imports are getting sorted correctly no matter what is happening on the local system's env vars.

docs/network_architecture.rst
little-dude: I agree with that but can we not block merging the PR because of this? Merging it as is would already fix the tests, and we could update the doc in a follow-up PR

tests/test_coordinator.py
janpetschexain: that's because the protobuf compiler basically generates two python thing out of an protobuf enum: 1) the actuall enum object and 2) a global constant for every enum value. calling it like `EnumName.EnumValue` is  just more readable that `EnumValue`.

setup.py
```diff
@@ -29,7 +29,7 @@
     "numproto==0.3",  # Apache License 2.0
     "structlog==19.2.0",  # Apache License 2.0
     # TODO: change xain-proto requirement to "xain-proto==0.2.0" once it is released
-    "xain-proto @ git+https://github.com/xainag/xain-proto.git@0e52b2fd1ceabbcccd443b05e2438a9ce0c65178#egg=xain_proto-0.2.0&subdirectory=python",  # Apache License 2.0
+    "xain-proto @ git+https://github.com/xainag/xain-proto.git@master#egg=xain_proto-0.1.0&subdirectory=python",  # Apache License 2.0
```
little-dude: IMO we should pin `xain-proto` to a specific revision to prevent breakage when it get updated.

setup.py
```diff
@@ -49,7 +49,7 @@
     "pytest==5.3.2",  # MIT license
     "pytest-cov==2.8.1",  # MIT
     "pytest-watch==4.2.0",  # MIT
-    "xain-sdk @ git+https://github.com/xainag/xain-sdk.git@development#egg=xain_sdk-0.2.0",
+    "xain-sdk @ git+https://github.com/xainag/xain-sdk.git@development#egg=xain_sdk-0.1.0",
```
little-dude: This seems weird, is it intentional?

docs/network_architecture.rst
finiteprods: yes i would not consider this a blocker, up to @janpetschexain whether to resolve this here or later!

.isort.cfg
finiteprods: good to know

docs/network_architecture.rst
janpetschexain: changed.

docs/network_architecture.rst
janpetschexain: replaced the diagram and added the source code as a comment.

setup.py
janpetschexain: it is. we increased the egg info to 0.2.0 a few tickets before because we didn't know if it was needed due to caching reasons. actually it was not needed, therefor i reversed it with the current update to be consistent with the package version again.

setup.py
janpetschexain: i agree with this in general. pointing to a branch `@ git+http...` is only a temporary solution anyway, would it be ok to leave as it is right now and fix this properly to `"xain-proto==0.2.0"` once it got released in the near future?

janpetschexain: done.

PR #246, opened 2020-01-30
PB-383 write aggregated weights after each round

little-dude:
This is based on #245, which should be merged first.

### References

https://xainag.atlassian.net/browse/PB-383

### Summary

We now store the weights at the end of *each* round instead of the end
of the final round.

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [ ] Passed scope checks.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,9 +42,11 @@ def coordinator_service():
     """
 
     server = grpc.server(futures.ThreadPoolExecutor(max_workers=1))
-    coordinator = Coordinator(minimum_participants_in_round=10, fraction_of_participants=1.0)
     store = FakeS3Store()
-    coordinator_grpc = CoordinatorGrpc(coordinator, store)
+    coordinator = Coordinator(
+        store=store, minimum_participants_in_round=10, fraction_of_participants=1.0
+    )
+    coordinator_grpc = CoordinatorGrpc(coordinator)
     coordinator_pb2_grpc.add_CoordinatorServicer_to_server(coordinator_grpc, server)
     server.add_insecure_port("localhost:50051")
     server.start()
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -42,9 +42,11 @@ def coordinator_service():
     """
 
     server = grpc.server(futures.ThreadPoolExecutor(max_workers=1))
-    coordinator = Coordinator(minimum_participants_in_round=10, fraction_of_participants=1.0)
     store = FakeS3Store()
```
finiteprods: should we now use `DummyStore` for tests rather than `FakeS3Store`? I understand you need the latter for `test_full_training_round` but for the rest, the former is probably better.

tests/conftest.py
```diff
@@ -62,16 +64,17 @@ def mock_coordinator_service():
...
     ctrl = IdController()
-    coordinator = Coordinator(
-        num_rounds=2,
-        minimum_participants_in_round=1,
-        fraction_of_participants=1.0,
-        aggregator=agg,
-        controller=ctrl,
-    )
-    with mock.patch("xain_fl.coordinator.coordinator_grpc.AbstractStore") as mock_obj:
+    with mock.patch("xain_fl.coordinator.store.AbstractStore") as mock_obj:
```
finiteprods: perhaps for consistency this should be whatever mock store is used above, rather than calling upon the mocking framework (at the time I wasn't sure how "sufficiently mocked" `FakeS3Store` was so did it this way).

tests/test_grpc.py
```diff
@@ -89,14 +89,15 @@ def test_participant_rendezvous_later(participant_stub):
     """
 
     # populate participants
-    coordinator = Coordinator(minimum_participants_in_round=10, fraction_of_participants=1.0)
+    coordinator = Coordinator(
+        store=FakeS3Store(), minimum_participants_in_round=10, fraction_of_participants=1.0
```
finiteprods: here you use `FakeS3Store` but is it necessary? also, in some other tests e.g. `test_monitor_heartbeats` you don't specify the store...

xain_fl/coordinator/store.py
```diff
@@ -42,6 +42,30 @@ def read_weights(self, participant_id: str, round: int) -> ndarray:
         """
 
 
+class DummyStore(AbstractStore):
```
finiteprods: perhaps more suggestive naming would be better, to distinguish `FakeS3Store` and `DummyStore` better. e.g. something like `LocalS3Store` and `MockStore`?

tests/conftest.py
little-dude: I removed it. By default the `Coordinator` uses the `DummyStore` that does nothing, so we don't need a mock indeed.

tests/conftest.py
little-dude: You're right, I changed that.

xain_fl/coordinator/store.py
little-dude: To me, "mock" sounds like something we use in tests, so I'd prefer not to rename `DummyStore` into `MockStore`. Following that logic, I did rename `FakeS3Store` into `MockS3Store` though. Is that better? @Robert-Steiner do you have an opinion regarding the names here?

tests/test_grpc.py
little-dude: It's not necessary indeed, I removed it. The default store is `DummyStore` which is enough for most tests, so we don't need to specify it.

xain_fl/coordinator/store.py
Robert-Steiner: Instead of `Dummy` it is sometimes called `NullObject` but for me it is fine to call it `DummyStore` and `MockStore`.

finiteprods: I think `NullStore` (or similar) would be more descriptive

little-dude: Let's go for `NullObjectStore`. I'll also make the change in `xain-sdk` for consistency.

little-dude: Done

PR #253, opened 2020-02-03
PB-366 flatten weights

janpetschexain:
### References

- [PB-366](https://xainag.atlassian.net/browse/PB-366)

### Summary

- use flat weights instead of list of weights
- update tests accordingly
- update documentation accordingly
- removed `hellonumproto` examples and tests

tested by running a dockerized coordinator and local participants on both sdk examples.

- none

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [x] Passed scope checks.


docs/network_architecture.rst
```diff
@@ -348,15 +348,15 @@ where the request and response data are given as the following protobuf messages
    message StartTrainingRoundRequest {}
 
    message StartTrainingRoundResponse {
-       repeated xain_proto.numproto.NDArray weights = 1;
+       xain_proto.np.NDArray weights = 1;
```
wilk10: i understand this has been implemented elsewhere in this way (`xain_proto` repo), but i would advise against using the `np` abbreviation because that is a convention reserved for the actual `numpy` package. but it doesn't quite fall under the scope of this PR, i guess

xain_fl/fl/coordinator/aggregate.py
```diff
@@ -18,17 +18,15 @@ def __init__(self):
...
     @abstractmethod
-    def aggregate(
-        self, mult_model_weights: List[List[ndarray]], aggregation_data: List[int]
-    ) -> List[ndarray]:
+    def aggregate(self, mult_model_weights: List[ndarray], aggregation_data: List[int]) -> ndarray:
         """Aggregate the weights of multiple models.
 
         Args:
-            mult_model_weights (List[List[ndarray]]): The weights of multiple models.
+            mult_model_weights (List[ndarray]): The weights of multiple models.
```
wilk10: i guess `multiple_model_weights` would be more explicit. it only saves 4 characters afterall

xain_fl/coordinator/store.py
```diff
@@ -13,73 +13,57 @@
 
 
 class AbstractStore(abc.ABC):
-
-    """An abstract class that defines the API a store must implement.
-
-    """
+    """An abstract class that defines the API a store must implement."""
```
wilk10: if the only changes to this file are about the docstrings and don't change any functionality, then i wouldn't include this in the PR, otherwise it may cause confusion in the commit history

xain_fl/fl/coordinator/aggregate.py
```diff
@@ -108,53 +100,42 @@ def aggregate(
...
-                    )
-                ],
-                axis=0,
-            )
-            for model_weights_per_idx in zip(*mult_model_weights)
-        ]
+        aggregated_model_weights: ndarray = np.sum(
+            [
+                model_weight * aggregation_weight
+                for model_weight, aggregation_weight in zip(mult_model_weights, aggregation_weights)
```
wilk10: if naming the variable `multiple_model_weights` as suggested above will cause this line to wrap to a new line, and make it less readable, then i would suggest to pass the zipped lists to a separate variable (just so that on this line it's shorter) (but still without using abbreviations), and perhaps separate the list comprehension from the numpy summation. but that's very much a detail. just minding readability only

docs/network_architecture.rst
janpetschexain: `np` was chosen explicitly as it stands for `numpy` (not `numproto`) to indicate directly that this is a protobuf implementation of numpy, as well as the associated utilites.

xain_fl/coordinator/store.py
janpetschexain: since it doesn't change the content/meaning of the docstring but only the formatting, i consider this simple cleanup which shouldn't require a dedicated ticket.

xain_fl/fl/coordinator/aggregate.py
janpetschexain: i thought about this as well, fine with both variants, i'll change it accordingly.

janpetschexain: changed.

xain_fl/fl/coordinator/aggregate.py
janpetschexain: changed.

xain_fl/coordinator/coordinator.py
```diff
@@ -100,7 +96,7 @@ def __init__(  # pylint: disable=too-many-arguments
         num_rounds: int = 1,
         minimum_participants_in_round: int = 1,
         fraction_of_participants: float = 1.0,
-        weights: Optional[List[ndarray]] = None,
+        weights: ndarray = np.empty(shape=(0,)),
```
finiteprods: it's not so obvious to me what `shape=(0,)` means. perhaps a short explanatory comment?

tests/test_grpc.py
```diff
@@ -513,4 +492,4 @@ def test_start_participant(mock_coordinator_service):
         assert coord.current_round == 1
 
         # expect weight aggregated by summation - see mock_coordinator_service
-        np.testing.assert_equal(coord.weights, [np.arange(start=10, stop=29, step=2)])
+        np.testing.assert_equal(coord.weights, init_weight)
```
finiteprods: now that weights are flat, there is no longer an interesting aggregate result to test for, which is a bit of a shame. but i would not necessarily consider that part of this PR!

xain_fl/coordinator/coordinator.py
janpetschexain: basically it's a 1-dimensional array, but it has no coordinates. this is probably the closest we can get to an empty vector without using `None`. this is somehow needed, since `None` is not available as a value for a protobuf field.

it's not ambiguous, since every proper model should have at least one weight. also we can use numpy utility via `np.empty(shape=(0,)).size` which evaluates to `0`. and this empty vector is serializable with our xain-proto lib, which doesn't work for example for `np.array(None, dtype=object)` which would be another way to represent `None` as a value for a protobuf message.

tests/test_grpc.py
finiteprods: I will create a separate ticket for this

xain_fl/coordinator/coordinator.py
finiteprods: makes sense, if not totally obvious!

PR #285, opened 2020-02-11
PB-398 Round resumption in Coordinator

finiteprods:
### References

[PB-398](https://xainag.atlassian.net/browse/PB-391)

### Summary

Adds proper support for round resumption in the Coordinator.

This merge request adds some resilience to the Coordinator in the face of Participant dropouts. In particular, when there is a dropout mid-round, the Coordinator pauses the round by going to standby. When enough Participants connect again, the round is resumed. While the idea of round resumption has been in the Coordinator for some time (and has recently been supported properly [in the Participant too](https://github.com/xainag/xain-sdk/pull/47)), up until now it has not been properly exercised, and this MR fixes the obvious faults, so that:

- `remove_participant` previously ignored the collection of selected Participants for a round; now it also takes that into account.
- `_handle_rendezvous` previously forced a complete re-selection each time a round resumes; the selection logic has been adjusted to only select from the connected pool while keeping the already selected survivors intact.

In addition:

- `test_select_outstanding` tests the above functionality.
- `test_remove_participant` has been split into 2 cases to properly test the above.
- tighter checks in the functions `is_finished` and `get_weight_updates` of the `Round` class.

This MR does not attempt to enable full fault-tolerance nor repair all the potential concurrency-related problems in the Coordinator e.g. ensuring various state collections are thread-safe, but these will be follow-up tasks going forward.

---

### Reviewer checklist

*Reviewer agreement:*

**Merge request checklist**

**Code checklist**

- [x] Passed scope checks.

### DIFF
--- a/xain_fl/coordinator/coordinator.py
+++ b/xain_fl/coordinator/coordinator.py
@@ -227,8 +227,10 @@ class Coordinator:  # pylint: disable=too-many-instance-attributes
             participant_id: The id of the participant to remove.
         """
 
-        self.participants.remove(participant_id)
         logger.info("Removing participant", participant_id=participant_id)
+        self.participants.remove(participant_id)
+        # remove from selected if necessary
+        self.round.remove_selected(participant_id)
 
         if self.participants.len() < self.minimum_connected_participants:
             self.state = State.STANDBY
@@ -239,6 +241,17 @@ class Coordinator:  # pylint: disable=too-many-instance-attributes
         self.controller.fraction_of_participants = self.fraction_of_participants
         selected_ids = self.controller.select_ids(self.participants.ids())
         self.round = Round(selected_ids)
+
+    def select_outstanding(self) -> List[str]:
+        """Selects participant ids."""
+
+        selected = set(self.round.participant_ids)
+        num_outstanding = self.minimum_participants_in_round - len(selected)
+        pool = set(self.participants.ids()) - selected
+        frac = num_outstanding / len(pool)
+
+        self.controller.fraction_of_participants = frac
+        return self.controller.select_ids(list(pool))
 
     def _handle_rendezvous(
         self, _message: RendezvousRequest, participant_id: str
@@ -265,7 +278,10 @@ class Coordinator:  # pylint: disable=too-many-instance-attributes
             # Select participants and change the state to ROUND if the latest added participant
             # lets us meet the minimum number of connected participants
             if self.participants.len() == self.minimum_connected_participants:
-                self.select_participant_ids_and_init_round()
+                # select enough to fill round if needed
+                if len(self.round.participant_ids) < self.minimum_participants_in_round:
+                    ids = self.select_outstanding()
+                    self.round.add_selected(ids)
 
                 self.state = State.ROUND
         else:
--- a/xain_fl/coordinator/coordinator.py
+++ b/xain_fl/coordinator/coordinator.py
@@ -243,7 +243,11 @@ class Coordinator:  # pylint: disable=too-many-instance-attributes
         self.round = Round(selected_ids)
 
     def select_outstanding(self) -> List[str]:
-        """Selects participant ids."""
+        """Selects participants outstanding for the round."""
+
+        # the following preconditions should hold
+        assert(len(self.round.participant_ids) < self.minimum_participants_in_round)
+        assert(self.participants.len() == self.minimum_connected_participants)
 
         selected = set(self.round.participant_ids)
         num_outstanding = self.minimum_participants_in_round - len(selected)
--- a/xain_fl/coordinator/coordinator.py
+++ b/xain_fl/coordinator/coordinator.py
@@ -246,8 +246,8 @@ class Coordinator:  # pylint: disable=too-many-instance-attributes
         """Selects participants outstanding for the round."""
 
         # the following preconditions should hold
-        assert(len(self.round.participant_ids) < self.minimum_participants_in_round)
-        assert(self.participants.len() == self.minimum_connected_participants)
+        assert len(self.round.participant_ids) < self.minimum_participants_in_round
+        assert self.participants.len() == self.minimum_connected_participants
 
         selected = set(self.round.participant_ids)
         num_outstanding = self.minimum_participants_in_round - len(selected)
--- a/xain_fl/coordinator/round.py
+++ b/xain_fl/coordinator/round.py
@@ -22,10 +22,30 @@ class Round:
         self.participant_ids = participant_ids
         self.updates: Dict[str, Dict] = {}
 
+    def add_selected(self, more_ids: List[str]) -> None:
+        """Add to the collection of selected participants.
+
+        Args:
+            more_ids: ids of participants to add.
+        """
+
+        self.participant_ids.extend(more_ids)
+
+    def remove_selected(self, participant_id: str) -> None:
+        """Remove from the collection of selected participants.
+
+        Args:
+            participant_id: id of participant to remove.
+        """
+
+        self.participant_ids = [
+            id for id in self.participant_ids if id != participant_id
+        ]
+
     def add_updates(
         self, participant_id: str, model_weights: ndarray, aggregation_data: int,
     ) -> None:
-        """Valid a participant's update for the round.
+        """Add a participant's update for the round.
 
         Args:
             participant_id: The id of the participant making the request.
@@ -57,7 +77,7 @@ class Round:
             round. `False` otherwise.
         """
 
-        return len(self.updates) == len(self.participant_ids)
+        return all(id in self.updates for id in self.participant_ids)
 
     def get_weight_updates(self) -> Tuple[List[ndarray], List[int]]:
         """Get a list of all participants weight updates.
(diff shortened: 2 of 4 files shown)


xain_fl/coordinator/coordinator.py
```diff
@@ -236,6 +238,17 @@ def select_participant_ids_and_init_round(self) -> None:
...
 
+    def select_outstanding(self) -> List[str]:
+        """Selects participant ids."""
+
+        selected = set(self.round.participant_ids)
+        num_outstanding = self.minimum_participants_in_round - len(selected)
+        pool = set(self.participants.ids()) - selected
+        frac = num_outstanding / len(pool)
+
+        self.controller.fraction_of_participants = frac
```
little-dude: Is it ok to override this? Isn't it a value that comes from the coordinator config file.

xain_fl/coordinator/coordinator.py
```diff
@@ -236,6 +238,17 @@ def select_participant_ids_and_init_round(self) -> None:
         selected_ids = self.controller.select_ids(self.participants.ids())
         self.round = Round(selected_ids)
 
+    def select_outstanding(self) -> List[str]:
+        """Selects participant ids."""
+
+        selected = set(self.round.participant_ids)
+        num_outstanding = self.minimum_participants_in_round - len(selected)
+        pool = set(self.participants.ids()) - selected
```
little-dude: Are we sure this is always strictly positive? Otherwise the `frac` calculation below might raise a `DivisionByZero` or be negative.

xain_fl/coordinator/coordinator.py
little-dude: I also don't see `fraction_of_participants` being used at all here. 
Say we have `fraction_of_participants = 0.5` and `minimum_participants_in_round = 10`.
Here is a scenario:

1. 20 participants connect => 10 are selected, and round start
2. 1 of the selected participant drops out, along with 9 of the non-selected participants => the selected participant is removed from the round and the participants list, and the 9 non-selected participants are removed from the participants list. Finally `select_outstanding()` is called
3. we have `selected = 9`, `num_outstanding = 1` and `pool = 1`, so `frac = 1` and we'll select 1 participant our of a pool of 1.

My question: shouldn't we only proceed to a new selection if we have enough connected participants such that `num_outstanding / pool <= fraction_of_participants`?

xain_fl/coordinator/coordinator.py
finiteprods: good question. here's an attempted mini-proof:

This function currently only gets called when there aren't enough participants selected _and_ there are enough connected to select from:
```python
len(self.round.participant_ids) < self.minimum_participants_in_round and
self.participants.len() == self.minimum_connected_participants             # (1)
```
and it should always be the case that
```python
self.minimum_participants_in_round <= self.minimum_connected_participants  # (2)
```
Now `num_outstanding` is necessarily positive due to `(1)` (assuming `minimum_participants_in_round` is, which should be!). Also `(1)` and `(2)` together imply
```python
len(self.round.participant_ids) < self.participants.len()
```
hence `pool` is also positive. Therefore `frac` is too.

xain_fl/coordinator/coordinator.py
finiteprods: yes, this is a bit unfortunate as the controller "API" isn't currently so great. It would have been easier if we could "select `n` from `m`" rather than insisting on setting a fraction. But I didn't want to tackle the controller API here, though it could be a separate issue.

xain_fl/coordinator/coordinator.py
little-dude: It may be worth assert a bunch of `assert` that checks our assumptions:

```python
assert(len(self.round.participant_ids) < self.minimum_participants_in_round)
# I think you meant >= here, instead of ==, right?
assert(self.participants.len() >= self.minimum_connected_participants   
# This one is so obvious I'm not sure this assertion is necessay
assert(self.minimum_participants_in_round <= self.minimum_connected_participants)
```

This would:

- catch a bug early if one of these invariants is violated
- help people reading/changing this function to understand under which condition it can be called

One additional note about (1):
```
len(self.round.participant_ids) < self.minimum_participants_in_round and
self.participants.len() == self.minimum_connected_participants
```

We must make sure that clients that sent their EndTraining message and that disconnected are kept in `self.round.participant_ids` but removed from `self.participants`. Can you confirm this is the case?

xain_fl/coordinator/coordinator.py
little-dude: Sounds good. What about my second question: is it ok when re-selecting new participants not to honor `num_outstanding / pool <= fraction_of_participants` ?

janpetschexain: actually, we must ignore this check. the reason is that both coordinator and controller have a `fraction_of_participants` attribute, but they refer to two different things. the coordinator uses `fraction_of_participants` as a ratio for all participants, but the controller uses this as a ratio for the participants relative to the respective set (which can for example be all participants, but also only outstanding participants).

i fully agree that we have to reorganize the controller api, as it's current usage is pretty obscure. but ofcourse this should be done in a dedicated merge request.

xain_fl/coordinator/round.py
```diff
@@ -22,10 +22,30 @@ def __init__(self, participant_ids: List[str]) -> None:
...
+    def remove_selected(self, participant_id: str) -> None:
+        """Remove from the collection of selected participants.
+
+        Args:
+            participant_id: id of participant to remove.
+        """
+
+        self.participant_ids = [
+            id for id in self.participant_ids if id != participant_id
+        ]
```
janpetschexain: list comprehension for identity comprehension is an anti pattern which shouldn't be used. i would suggest to do the following:

```
try:
    self.participant_ids.remove(participant_id)
except ValueError:
    pass
```

xain_fl/coordinator/coordinator.py
finiteprods: sorry I hadn't refreshed, missed the 2nd question.

In your scenario, 20 connect and 10 are selected. One of those selected drops (along with a bunch of the others), so the round is paused. Currently, the round resumes only when we have `minimum_connected_participants` again i.e. 20. At this point, we will be selecting 1 outstanding out of a pool of 11.

More generally, because we only re-select when `minimum_connected_participants`, even in the worst case (where we must re-select everyone) we're just back to `fraction_of_participants`.

xain_fl/coordinator/coordinator.py
finiteprods: yes I agree it would indeed be helpful to have some pre-conditions in there.

> `# I think you meant >= here, instead of ==, right?`

the naming of `minimum_connected_participants` is a little confusing at the moment as we shouldn't ever have more than this number connected. I believe the original idea was that there could potentially be more than this (and there is till a ticket for it), hence the naming.

> We must make sure that clients that sent their EndTraining message and that disconnected are kept in `self.round.participant_ids` but removed from `self.participants`.

Had thought about this as well, but decided against it for now after some deliberation. The thinking was, yes it would be better not to waste the work done by these guys, but I would consider this just as an optimisation over the more conservative approach of replacing them. Otherwise, it's an edge case that complicates the already quite complicated logic, and also the potential for `self.round.participant_ids` and `self.participants` to be out of sync might create issues down the line.

xain_fl/coordinator/coordinator.py
little-dude: Ah ok thanks for the explanation!

finiteprods: @janpetschexain I created a ticket for the controller API issue https://xainag.atlassian.net/browse/PB-433 cheers

Robert-Steiner:
> During the testing of your PR I got the following error:
> 
> ```
> xain-fl-dev_1      | Traceback (most recent call last):
> xain-fl-dev_1      |   File "/usr/local/bin/coordinator", line 11, in <module>
> xain-fl-dev_1      |     load_entry_point('xain-fl', 'console_scripts', 'coordinator')()
> xain-fl-dev_1      |   File "/app/xain_fl/__main__.py", line 52, in main
> xain-fl-dev_1      | ['ipv4:172.22.0.1:35554' 'ipv4:172.22.0.1:35544' 'ipv4:172.22.0.1:35548']
> xain-fl-dev_1      |     serve(coordinator=coordinator, server_config=config.server)
> xain-fl-dev_1      |   File "/app/xain_fl/serve.py", line 46, in serve
> xain-fl-dev_1      |     monitor_heartbeats(coordinator, terminate_event)
> xain-fl-dev_1      |   File "/app/xain_fl/coordinator/heartbeat.py", line 46, in monitor_heartbeats
> xain-fl-dev_1      |     coordinator.remove_participant(participant_id)
> xain-fl-dev_1      |   File "/app/xain_fl/coordinator/coordinator.py", line 233, in remove_participant
> xain-fl-dev_1      |     self.round.remove_selected(participant_id)
> xain-fl-dev_1      |   File "/app/xain_fl/coordinator/round.py", line 43, in remove_selected
> xain-fl-dev_1      |     self.participant_ids.remove(participant_id)
> xain-fl-dev_1      | AttributeError: 'numpy.ndarray' object has no attribute 'remove'
> ```
> 
> What I did is:
> 
>     * Set `min_participants = 3`
> 
>     * start 3 participants
> 
>     * kill one participant after the first round
> 
> 
> I will check why it happens.

The issues comes from `RandomController`
it returns a numpy array

PR #337, opened 2020-03-24
remove nix files

little-dude:
As discussed this morning on Slack, it would be nice to keep the repo as clean as possible. This is WIP because I'm experimenting with a setup based on git worktrees. The idea would be to have a `nix` branch that we can "mount" as a worktree, to provide the files for nix users.


little-dude:
For those interested I've found a pretty cool workflow with git worktree that allows me to keep these files under version control, in the same repo: I keep them in my own fork under a `nix` branch. Locally, I check that branch out in a worktree and call a script that deploys my files and update the local `.git/info/exclude` file so that they are not added by mistake later.

PR #339, opened 2020-03-24
PB-484 pet messages

janpetschexain:
**References**

- [PB-484](https://xainag.atlassian.net/browse/PB-484) including [PB-498](https://xainag.atlassian.net/browse/PB-498), [PB-499](https://xainag.atlassian.net/browse/PB-499), [PB-500](https://xainag.atlassian.net/browse/PB-500)

**Summary**

- compose and validate "sum", "update" and "sum2" messages of the PET protocol
- provide multi-layer buffers to serialize and deserialize the messages

**Open tasks**
- none

### DIFF
--- a/rust/src/pet/participant.rs
+++ b/rust/src/pet/participant.rs
@@ -1,17 +1,17 @@
 #![allow(dead_code)] // temporary
 
-use std::collections::HashMap;
+use std::{collections::HashMap, default::Default};
 
 use sodiumoxide::{
+    self,
     crypto::{box_, sealedbox, sign},
-    init,
     randombytes::randombytes,
 };
 
 use super::{utils::is_eligible, PetError};
 
 /// Tasks of a participant.
-pub enum Task {
+enum Task {
     Sum,
     Update,
     None,
@@ -35,27 +35,11 @@ pub struct Participant {
 }
 
 impl Participant {
-    pub fn new() -> Self {
-        let (encr_pk, encr_sk) = box_::gen_keypair();
-        let (sign_pk, sign_sk) = sign::gen_keypair();
-        let ephm_pk = box_::PublicKey([0_u8; box_::PUBLICKEYBYTES]);
-        let ephm_sk = box_::SecretKey([0_u8; box_::SECRETKEYBYTES]);
-        let certificate: Vec<u8> = Vec::new();
-        let signature_sum = sign::Signature([0_u8; sign::SIGNATUREBYTES]);
-        let signature_update = sign::Signature([0_u8; sign::SIGNATUREBYTES]);
-        let task = Task::None;
-        Self {
-            encr_pk,
-            encr_sk,
-            sign_pk,
-            sign_sk,
-            ephm_pk,
-            ephm_sk,
-            certificate,
-            signature_sum,
-            signature_update,
-            task,
-        }
+    pub fn new() -> Result<Self, PetError> {
+        // crucial: must be called before anything else in this module
+        sodiumoxide::init()
+            .and(Ok(Default::default()))
+            .or(Err(PetError::InvalidMessage))
     }
 
     /// Compute the "sum" and "update" signatures.
@@ -82,6 +66,31 @@ impl Participant {
     }
 }
 
+impl Default for Participant {
+    fn default() -> Self {
+        let (encr_pk, encr_sk) = box_::gen_keypair();
+        let (sign_pk, sign_sk) = sign::gen_keypair();
+        let ephm_pk = box_::PublicKey([0_u8; box_::PUBLICKEYBYTES]);
+        let ephm_sk = box_::SecretKey([0_u8; box_::SECRETKEYBYTES]);
+        let certificate: Vec<u8> = Vec::new();
+        let signature_sum = sign::Signature([0_u8; sign::SIGNATUREBYTES]);
+        let signature_update = sign::Signature([0_u8; sign::SIGNATUREBYTES]);
+        let task = Task::None;
+        Self {
+            encr_pk,
+            encr_sk,
+            sign_pk,
+            sign_sk,
+            ephm_pk,
+            ephm_sk,
+            certificate,
+            signature_sum,
+            signature_update,
+            task,
+        }
+    }
+}
+
 /// Buffer and wrap the asymmetrically encrypted part of a "sum/update/sum2" message.
 pub struct SealedBoxBuffer(Vec<u8>);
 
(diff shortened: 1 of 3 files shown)


rust/src/pet/participant.rs
```diff
@@ -1,122 +1,306 @@
...
+// signature_sum ----┤                 |
+// signature_update -┤                 |
+// mask_url ---------┤                 |
+//                   └-> Sum2BoxBuffer |
+//                        └-> Sum2Box -┤
+//                                     └-> MessageBuffer
+//                                          └-> Sum2Message
+
+/// Buffer and wrap the asymmetrically encrypted part of a "sum/update/sum2" message.
+struct SealedBoxBuffer<'a>(&'a [u8], &'a [u8], &'a [u8]);
```
little-dude: This works, but I think it would be slightly more user friendly to not require the lifetimes to be the same, to make things a bit more flexible. Also giving names to the lifetimes sometimes helps the reader to better understand what's going on.

```
struct SealedBoxBuffer<'sum, 'update, 'sum2>(&'sum [u8], &'update [u8], &'sum2 [u8]);
```

janpetschexain: i'll add lifetime naming, it definitely helps readability and adds flexibility. i would keep the tuple struct and not change it to to a struct, since it is a private object without an external api anyways.

little-dude:
I think this would be good to merge already. The comments help a lot understanding the hierarchy of the structs, and the code is understandable, even for someone who hasn't looked in much depth into the protocol.

There are a couple of points that I think could be improved. Note that what follows is somewhat a personal opinion so take it with a grain of salt.

My main gripe with this implementation is that we define many struct twice, once in `pet::coordinator` and once in `pet::participant`, with one version being used for parsing messages, and the other one for creating new ones. The fact that they don't have the same structure adds to the confusion. For instance, `SumBoxBuffer` is defined in `coordinator.rs` as:

```
struct SumBoxBuffer<'a>(&'a [u8]);
```

But in `participant.rs` as:

```
struct SumBoxBuffer<'a>(&'a [u8], &'a [u8], &'a [u8], &'a [u8], &'a [u8]);
```

I'd like to suggest the following `SumBoxBuffer` instead, which could be generalized to the other `Buffer`types. This is a pattern I've used in the past, that comes from this awesome article: https://lab.whitequark.org/notes/2016-12-13/abstracting-over-mutability-in-rust/

The idea is to have multiple `impl`s: one for mutable data (ie setters) and one for non mutable data (ie getters):


```rust
use super::PetError;
use sodiumoxide::{
    self,
    crypto::{box_, sign},
};
use std::ops::Range;

const SUM_RANGE: Range<usize> = 0..3;
const CERTIFICATE_RANGE: Range<usize> = 3..3;
const SIGNATURE_SUM_RANGE: Range<usize> = 3..67;
const SIGNATURE_UPDATE_RANGE: Range<usize> = 67..131;
const EPHM_PK_RANGE: Range<usize> = 131..163;
const MESSAGE_LENGTH: usize = 163;

#[derive(Debug, PartialEq, Eq, Clone)]
pub struct SumBoxBuffer<T> {
    inner: T,
}

impl<T: AsRef<[u8]>> SumBoxBuffer<T> {
    pub fn new(bytes: T) -> SumBoxBuffer<T> {
        SumBoxBuffer { inner: bytes }
    }

    pub fn new_checked(bytes: T) -> Result<SumBoxBuffer<T>, PetError> {
        let buffer = Self::new(bytes);
        buffer.check()?;
        Ok(buffer)
    }

    fn len(&self) -> usize {
        self.inner.as_ref().len()
    }

    fn check(&self) -> Result<(), PetError> {
        if self.len() != MESSAGE_LENGTH {
            return Err(PetError::InvalidMessage);
        }
        Ok(())
    }

    pub fn into_inner(self) -> T {
        self.inner
    }
}

impl<'a, T: AsRef<[u8]> + ?Sized> SumBoxBuffer<&'a T> {
    pub fn kind(&self) -> &'a [u8] {
        let data = self.inner.as_ref();
        &data[SUM_RANGE]
    }

    pub fn certificate(&self) -> &'a [u8] {
        let data = self.inner.as_ref();
        &data[CERTIFICATE_RANGE]
    }

    pub fn signature_sum(&self) -> &'a [u8] {
        let data = self.inner.as_ref();
        &data[SIGNATURE_SUM_RANGE]
    }

    pub fn signature_update(&self) -> &'a [u8] {
        let data = self.inner.as_ref();
        &data[SIGNATURE_UPDATE_RANGE]
    }

    pub fn ephm_pk(&self) -> &'a [u8] {
        let data = self.inner.as_ref();
        &data[EPHM_PK_RANGE]
    }
}

impl<T: AsRef<[u8]> + AsMut<[u8]>> SumBoxBuffer<T> {
    pub fn kind_mut(&mut self) -> &mut [u8] {
        let data = self.inner.as_mut();
        &mut data[SUM_RANGE]
    }

    pub fn certificate_mut(&mut self) -> &mut [u8] {
        let data = self.inner.as_mut();
        &mut data[CERTIFICATE_RANGE]
    }

    pub fn signature_sum_mut(&mut self) -> &mut [u8] {
        let data = self.inner.as_mut();
        &mut data[SIGNATURE_SUM_RANGE]
    }

    pub fn signature_update_mut(&mut self) -> &mut [u8] {
        let data = self.inner.as_mut();
        &mut data[SIGNATURE_UPDATE_RANGE]
    }

    pub fn ephm_pk_mut(&mut self) -> &mut [u8] {
        let data = self.inner.as_mut();
        &mut data[EPHM_PK_RANGE]
    }
}
```

This buffer type is straightforward and only deals with bytes. In our code, we should use a higher level representation, that leverages the buffer type for serialization and deserialization:

```rust
pub struct SumBox {
    pub certificate: Vec<u8>,
    pub signature_sum: sign::Signature,
    pub signature_update: sign::Signature,
    pub ephm_pk: box_::PublicKey,
}

impl SumBox {
    pub fn from_bytes(data: &[u8]) -> Result<Self, PetError> {
        let buf = SumBoxBuffer::new_checked(data)?;
        let certificate = buf.certificate().to_vec();
        let signature_sum =
            sign::Signature::from_slice(buf.signature_sum()).ok_or(PetError::InvalidMessage)?;
        let signature_update =
            sign::Signature::from_slice(buf.signature_update()).ok_or(PetError::InvalidMessage)?;
        let ephm_pk = box_::PublicKey::from_slice(buf.ephm_pk()).ok_or(PetError::InvalidMessage)?;
        Ok(Self {
            certificate,
            signature_sum,
            signature_update,
            ephm_pk,
        })
    }

    pub fn to_bytes(&self, data: &mut [u8]) -> Result<(), PetError> {
        let mut buf = SumBoxBuffer::new_checked(data)?;
        buf.kind_mut().copy_from_slice(b"sum");
        buf.signature_sum_mut()
            .copy_from_slice(&self.signature_sum.0[..]);
        buf.signature_update_mut()
            .copy_from_slice(&self.signature_update.0[..]);
        buf.ephm_pk_mut().copy_from_slice(&self.ephm_pk.0[..]);
        Ok(())
    }
}
```

That looks like a lot of boilerplate, but here is the nices part: the `SumBoxBuffer` implementation can almost entirely be generated with macros, which I can help with (I have written such a macro for another project already).

PR #340, opened 2020-03-24
Upgrade InfluxDB to 0.1.0

### LINKED ISSUE #318: Upgrade `influxdb`
1.0 was released recently and contains breaking changes:

```
error[E0599]: no function or associated item named `write_query` found for trait object `dyn influxdb::query::Query` in the current scope
  --> src/metric_store/influxdb.rs:36:28
   |
36 |                     Query::write_query(Timestamp::Now, metric_owner.to_string());
   |                            ^^^^^^^^^^^ function or associated item not found in `dyn influxdb::query::Query`
```

Robert-Steiner:
Closes #318

### DIFF
--- a/rust/src/bin/coordinator.rs
+++ b/rust/src/bin/coordinator.rs
@@ -9,7 +9,7 @@ use tracing_futures::Instrument;
 
 #[cfg(feature = "influx_metrics")]
 use xain_fl::{
-    common::metric_store::influxdb::{run_metricstore, InfluxDBMetricStore},
+    common::metric_store::influxdb::{run_metricstore, InfluxDBConnector},
     coordinator::settings::MetricStoreSettings,
 };
 
@@ -100,7 +100,7 @@ async fn _main(
     #[cfg(feature = "influx_metrics")]
     let metric_sender = if let Some(metric_store) = metric_store {
         // Start the metric store
-        let (influx_client, metric_sender) = InfluxDBMetricStore::new(
+        let (influx_client, metric_sender) = InfluxDBConnector::new(
             &metric_store.database_url[..],
             &metric_store.database_name[..],
         );
--- a/rust/src/common/metric_store/influxdb.rs
+++ b/rust/src/common/metric_store/influxdb.rs
@@ -71,7 +71,7 @@ pub async fn run_metricstore(mut influxdb_connector: InfluxDBConnector) {
                     Measurement::Round(round) => round.into_query("coordinator"),
                     Measurement::Counter(counter) => counter.into_query("coordinator"),
                 };
-                influxdb_connector
+                let _ = influxdb_connector
                     .client
                     .query(&query)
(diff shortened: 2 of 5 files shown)


rust/src/coordinator/core/service.rs
```diff
@@ -416,41 +414,23 @@ where
     #[cfg(feature = "influx_metrics")]
     fn write_counter_metrics(&self) {
         self.metrics_tx.as_ref().map(|tx| {
-            let _ = tx.send(Metric(
```
little-dude: Feels good :p

rust/src/bin/coordinator.rs
```diff
@@ -9,7 +9,7 @@ use tracing_futures::Instrument;
 
 #[cfg(feature = "influx_metrics")]
 use xain_fl::{
-    common::metric_store::influxdb::{run_metricstore, InfluxDBMetricStore},
+    common::metric_store::influxdb::{run_metricstore, InfluxDBConnector},
```
little-dude: Why the renaming? I'm don't mind it I'm just curious about the reason.

rust/src/common/metric_store/influxdb.rs
```diff
@@ -1,48 +1,79 @@
...
+    }
+}
 
-                // Submit the query to InfluxDB.
-                let _ = influx_client
+pub async fn run_metricstore(mut influxdb_connector: InfluxDBConnector) {
+    loop {
+        match influxdb_connector.receiver.recv().await {
+            Some(measurement) => {
+                let query = match measurement {
```
little-dude: As discussed on slack the could be extracted in an impl of [`InfluxDbWriteable`](https://docs.rs/influxdb/0.1.0/influxdb/trait.InfluxDbWriteable.html) for `Measurement` itself:

```rust
impl InfluxDbWriteable for Measurement {
    fn into_query<I: Into<String>>(self, name: I) -> WriteQuery {
        match self {
 		     Self::Round(round) => round.into_query(name),
 		     Self::Counters(counters) => counters.into_query(name)
        }
    }
}
```

I think that would be slightly cleaner, but that's perhaps just a matter of personal opinion so take this as a mere suggestion.

rust/src/bin/coordinator.rs
Robert-Steiner: I thought having the name metric store twice in the path is a bit too much. 
My plan was actually to do something like this:
```
common::metric_store::{
	run_metricstore,
	influxdb::Connector
}
```
But I wasn't sure if this change makes so much sense if we only have one metric store implementation.

little-dude: I think it makes sense. I don't know if we'll have other implementations but it doesn't hurt to plan for it.

Robert-Steiner: I have created a issue for that #347

PR #342, opened 2020-03-25
PB-584 Add LICENSE file at the root of the repository prior to release v0.7.0

rsaffi:


rsaffi:
Will proceed with merging as it's just the default Apache 2.0 license

PR #345, opened 2020-03-25
PB-519 Add swagger files

Robert-Steiner:
**References**
[PB-569](https://xainag.atlassian.net/browse/PB-569)

**Summary**
- added swagger files that describes the current API for the coordinator as well the aggregator
- added swagger ui in docker compose file

**Open tasks**
- create script that updates the versions within the swagger files

### DIFF
--- a/swagger/aggregator.yml
+++ b/swagger/aggregator.yml
@@ -1,7 +1,6 @@
 openapi: 3.0.1
 info:
   title: XAIN FL Aggregator API
-  description: ""
   contact:
     email: engineering@xain.io
   license:
@@ -15,7 +14,7 @@ servers:
   - url: http://localhost:8082/
 tags:
   - name: Aggregator
-    description: Aggregator API
+    description: API of the aggregator service
 paths:
   /{client_id}/{client_token}:
     get:
@@ -26,24 +25,26 @@ paths:
           in: path
           description: ID of the client
           required: true
+          example: 1fa2f908-83e2-4f92-80e1-4baee0bf25a5
           schema:
             type: string
         - name: client_token
           in: path
-          description: Aggregator token of the client
+          description: API token of the client
           required: true
+          example: 2e8700a6-390d-411f-977b-9aa3bebc13e9
           schema:
             type: string
       responses:
         200:
-          description: successful operation
+          description: weights of the global model
           content:
             application/octet-stream:
               schema:
                 type: string
                 format: binary
         404:
-          description: ""
+          description: client unknown
           content: {}
     post:
       tags:
@@ -53,16 +54,18 @@ paths:
           in: path
           description: ID of the client
           required: true
+          example: 1fa2f908-83e2-4f92-80e1-4baee0bf25a5
           schema:
             type: string
         - name: client_token
           in: path
-          description: Aggregator token of the client
+          description: API token of the client
           required: true
+          example: 2e8700a6-390d-411f-977b-9aa3bebc13e9
           schema:
             type: string
       requestBody:
-        required: true
+        description: weights of the local model
         content:
           application/octet-stream:
             schema:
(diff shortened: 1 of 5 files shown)


swagger/aggregator.yml
```diff
@@ -0,0 +1,81 @@
...
+      tags:
+        - Aggregator
+      parameters:
+        - name: client_id
+          in: path
+          description: ID of the client
+          required: true
+          example: 1fa2f908-83e2-4f92-80e1-4baee0bf25a5
+          schema:
+            type: string
```
little-dude: maybe:

```yaml
schema:
  type: string
  format: uuid
```

swagger/aggregator.yml
```diff
@@ -0,0 +1,81 @@
+openapi: 3.0.1
+info:
+  title: XAIN FL Aggregator API
+  contact:
+    email: engineering@xain.io
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+  version: 1.0.0
```
little-dude: This should be `0.0.7`. But I'm afraid we forget to update that when bumping the versions. Perhaps we should have a script.for this.

swagger/aggregator.yml
```diff
@@ -0,0 +1,81 @@
...
+  url: https://docs.xain.io/
+servers:
+  - url: http://localhost:8082/
+tags:
+  - name: Aggregator
+    description: API of the aggregator service
+paths:
+  /{client_id}/{client_token}:
+    get:
+      tags:
```
little-dude: I'm not sure we really need tags atm but it doesn't hurt so we can leave it.

swagger/aggregator.yml
```diff
@@ -0,0 +1,81 @@
...
+                type: string
+                format: binary
+        404:
+          description: client unknown
+          content: {}
+    post:
+      tags:
+        - Aggregator
+      parameters:
+        - name: client_id
```
little-dude: `client_id` comes back in many requests. I think there's a way to define a `ClientId` type that would be re-used no? Maybe it's overkill here since we don't have that many routes though.

swagger/coordinator.yml
```diff
@@ -0,0 +1,116 @@
...
+      parameters:
+        - name: client_id
+          in: path
+          description: ID of the client
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: client is selected for training round
```
little-dude: The description should be something like "get the  information for fetching the latest global model"

swagger/coordinator.yml
```diff
@@ -0,0 +1,116 @@
...
+    RendezVousResponse:
+      type: object
+      properties:
+        id:
+          type: string
+          description: client ID
+          example: 1fa2f908-83e2-4f92-80e1-4baee0bf25a5
+        ok:
+          type: boolean
+          description: ""
```
little-dude: Whether the rendez-vous was accepted (`true`) or rejected (`false`) by the coordinator. A participant that has been rejected should not try to recontact the coordinator.

swagger/coordinator.yml
```diff
@@ -0,0 +1,116 @@
...
+    HeartBeatResponse:
+      type: object
+      properties:
+        round:
+          minimum: 0
+          type: integer
+          description: current round of the coordinator
```
little-dude:
```
If non-null, this field indicates that the participant has been selected for the given round.
```

swagger/coordinator.yml
```diff
@@ -0,0 +1,116 @@
...
+          format: int32
```
little-dude: Are there unsigned integer? Because that field is always `null` or `>= 0`

PR #346, opened 2020-03-25
Mock the aggregator RPC client

little-dude:
## Description

The main tasks, _ie_ that tasks that implement most of the business logic are `aggregator::Service` for the aggregator and `coordinator::Service` for the coordinator. This is where we'll focus our effort for testing at the beginning. In order to test these two `Service`, we need to control their inputs and outputs.

Let's focus on the coordinator, since the situation is pretty similar for the aggregator anyway. The question is: what are the sources of IO. We have:

- an RPC server that receives requests from the network and sends back responses
- an RPC client that sends requests to the network and receives responses back
- an REST API that receives requests from the network, and sends back responses

However, two of these are not _direct_ sources of the `Service`. The API and RPC server layers are already isolated and run in their own task. They communicate with the `Service` via a `Handle` as illustrated below:

```
                                           +-------------+
                                           |     API     |
                                           | +------+    |
    +---------------------------+      +----->Handle|    <---+
    |           Service         |      |   | +------+    |   |
    | +----------+   +--------+ |      |   +-------------+   |
  +--->RPC client|   |Receiver<--------+   +-------------+   |
  | | +----------+   +--------+ |      |   |  RPC server |   |
  | +---------------------------+      |   | +------+    |   |
  |                                    +----->Handle|    <---+
  |                                        | +------+    |   |
  |                                        +-------------+   |
  |                                                          |
  |                   +---------+                            |
  |                   |         |                            |
  +-------------------> Network <----------------------------+
                      |         |
                      +---------+
```

We already control these handles and we can use them to send requests to the service. Thus, there is no need to mock the API and RPC server and our system under test can already be simplified:

```
                                           +------+
  +---------------------------+      +----->Handle|
  |           Service         |      |     +------+
  | +----------+   +--------+ |      |
+--->RPC client|   |Receiver<--------+
| | +----------+   +--------+ |      |
| +---------------------------+      |     +------+
|                                    +----->Handle|
|                                          +------+
|
|
|                   +---------+
|                   |         |
+-------------------> Network |
                    |         |
                    +---------+
```

The RPC client is the only remaining piece, and the idea of the PR is to mock it away so that our system under test is finally fully under our control:

```
                                         +------+
+---------------------------+      +----->Handle|
|           Service         |      |     +------+
| +----------+   +--------+ |      |
| |  Mock    |   |Receiver<--------+
| +----------+   +--------+ |      |
+---------------------------+      |     +------+
                                   +----->Handle|
                                         +------+
```

--------------------------------

## Implementation

For the mocking itself, we use the `mockall` crate. The idea is to leverage rust's conditional compilation to replace the `rpc::Client` by our mock when running `cargo test`. For this, we use the `#[cfg(test)]` and `#[cfg(not(test))]` attributes.

The mock itself doesn't expose the exact same API than the `rpc::Client`: the types in some of the function signatures are slightly simpler. That's works well in Rust because most APIs are designed around traits.

For instance the real `rpc::Client::select` method looks like this:

```
pub fn select(
    &mut self,
    ctx: Context,
    credentials: Credentials
) -> impl Future<Output = Result<Result<(), ()>>> + '_
```

Thus, the return type is _some type that implements `Future<Output=io::Result<Result<(), ()>>>`, and an anonymous lifetime (the `'_` part). But in our mock, the return type is an actual concrete type:

```
fn select(&mut self, ctx: Context, credentials: Credentials) -> future::Ready<io::Result<Result<(), ()>>>;
```

It works, because that concrete type satisfies the `Future<Output = Result<Result<(), ()>>> + '_` trait bound.

### DIFF
--- a/rust/src/aggregator/rpc.rs
+++ b/rust/src/aggregator/rpc.rs
@@ -28,7 +28,42 @@ mod inner {
     }
 }
 
-pub use inner::{Rpc, RpcClient as Client};
+pub use inner::Rpc;
+
+#[cfg(test)]
+mod mocks {
+    pub use super::*;
+
+    use futures::future;
+    use mockall::mock;
+    use std::io;
+    use tarpc::{client::Config, context::Context, rpc::Transport};
+
+    mock! {
+        pub NewClient {
+            fn spawn(self) -> io::Result<MockRpcClient>;
+        }
+    }
+
+    mock! {
+        pub RpcClient {
+            fn new<T: Transport<(), ()> + 'static>(config: Config, transport: T) -> MockNewClient;
+            fn select(&mut self, ctx: Context, credentials: Credentials) -> future::Ready<io::Result<Result<(), ()>>>;
+            fn aggregate(&mut self, ctx: Context) -> future::Ready<io::Result<Result<(), ()>>>;
+        }
+    }
+
+    impl Clone for MockRpcClient {
+        fn clone(&self) -> Self {
+            Self::default()
+        }
+    }
+}
+
+#[cfg(not(test))]
+pub use inner::RpcClient as Client;
+#[cfg(test)]
+pub use mocks::MockRpcClient as Client;
 
 /// A server that serves a single client. A new `Server` is created
 /// for each new client.
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -265,6 +265,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "difference"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "524cbf6897b527295dff137cec09ecf3a05f4fddffd7dfcd1585403449e74198"
+
+[[package]]
 name = "digest"
 version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


rust/src/aggregator/rpc.rs
```diff
@@ -28,7 +28,43 @@ mod inner {
...
+    use tarpc::{client::Config, context::Context, rpc::Transport};
+
+    mock! {
+        pub NewClient {
+            fn spawn(self) -> io::Result<MockClient>;
+        }
+    }
+
+    mock! {
+        pub Client {
```
janpetschexain: for readability, i would suggest to name the mocked object exactly the same as the real ones and assign an alias later by `use`. i.e.
```
mock! { pub RpcClient {...} }

#[cfg(test)]
pub use mocks::MockRpcClient as Client;
```
this makes it clearer what is to be mocked here. same for `NewRpcClient` instead of `RpcClient`.

rust/src/aggregator/rpc.rs
```diff
@@ -28,7 +28,43 @@ mod inner {
...
+
+    impl Clone for MockClient {
+        fn clone(&self) -> Self {
+            Self::default()
+        }
+    }
+}
+
+#[cfg(test)]
+pub use mocks::MockClient as Client;
```
janpetschexain: for readability, can we place the two cfg-dependent inputs of the `RpcClient` directly one below the other? this way it's clearer that they stand for a xor import of the same object under different configurations.

rust/src/aggregator/rpc.rs
little-dude: Good point. I'll do that.

rust/src/aggregator/rpc.rs
little-dude: Also a good point. I'll do that.

little-dude:
Closing in favor of https://github.com/xainag/xain-fl/pull/352

PR #348, opened 2020-03-26
Improve the docker infra

little-dude:
The image we currently produce is over 4GB large. In order to reduce
this, we perform the compilation in a build container, and just copy
the coordinator and aggregator binaries to the final image.

This commit also removes the dependency vendoring. The main reason for
having it was to cache the dependencies. But that doesn't cache the
builds. Instead we use `cargo fetch` and `cargo build` on an empty
project that contains all our dependencies to force cargo to cache
them.

Finally, we split the images:

- an image to upload to docker hub, that contains a release build
- an image for development that contains a debug build as well as some
  dev tools (Valgrind)


docker/Dockerfile.release
```diff
@@ -0,0 +1,21 @@
+FROM python:3.7.6-slim-buster AS builder
```
dkravetz: Is there a reason to let go of the build argument and using 2 dockerfiles? (I think I'm missing something here :smile: )

little-dude: We could keep the args but there are a couple small differences in several places between the two images so I needed to check the args in multiple places:

- in the debug image, we install dev tools (only `valgrind` for now, but I'm planning to also install `perf` real soon)
- in the debug image we use a trick to cache the dependencies before building the project, which we don't use for the release image. We could but I think there isn't really a need for it: 99% of the time, we'll build the release image via CI, so there won't be any caching possible and if there's no caching "trick" makes builds slightly longer.
- of course we use have different cargo flags for debug and build images

Given that, and the fact that conditional logic is rather limited in docker (it basically defers to bash), I thought it was just simpler to just use two different Dockerfiles.

dkravetz: Good enough for me

PR #351, opened 2020-03-27
Examples improvements

little-dude:
This is work I did last week but forgot to merge to the new repo. I think it will be very useful for the next testing session.

- use human readable units for the `--model-size` argument of the dummy example
- make the heartbeat period configurable: this can speed the examples up significantly, but also help putting more pressure on the coordinator, or force a high dropout rate for example
- make logging more configurable
- add a script to start multiple participants for the keras example
- keep track of the performance metrics in the keras example


python/sdk/xain_sdk/__init__.py
```diff
@@ -4,21 +4,20 @@
 
 from .interfaces import TrainingInputABC, TrainingResultABC
 from .participant import ParticipantABC, ParticipantError
+from .utils import configure_logging  # pylint: disable=import-outside-toplevel
```
rsaffi: In theory the `pylint` hint could be removed now.

little-dude: done

PR #353, opened 2020-03-27
PB-569 first aggregator test

Robert-Steiner:
**Summary**

- add first aggregator io test

### DIFF
--- a/rust/src/tests/lib/aggregator.rs
+++ b/rust/src/tests/lib/aggregator.rs
@@ -3,7 +3,7 @@ use crate::{
     common::client::Credentials,
 };
 use bytes::Bytes;
-use std::{future::Future, pin::Pin};
+use futures::future;
 
 #[derive(Clone)]
 pub struct ServiceHandle(InnerServiceHandle);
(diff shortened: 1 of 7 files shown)


rust/src/tests/mocks/aggregator.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    type AddWeightsFut = Pin<Box<dyn Future<Output = Result<(), ()>> + Send>>;
+    type AggregateFut = Pin<Box<dyn Future<Output = Result<Bytes, ()>> + Send>>;
+
+    fn add_weights(&mut self, weights: Bytes) -> Self::AddWeightsFut {
+        self.weights.push(weights);
+        Box::pin(async move { Ok(()) })
+    }
+
+    fn aggregate(&mut self) -> Self::AggregateFut {
```
little-dude: That may a bit too fancy. It would be nice to have an aggregator that returns something unambiguous, easily predictable and verifiable.

Perhaps `add_weights` could remain as it is and append the incoming bytes, and `aggregate` could sort the resulting vec. That way when we have tests where multiple participants upload weights at the same time, we can easily determine the expected value. If we have the participants upload different sequences of bytes with distinct values (like `[0, 0, 0, 0]`, `[1, 1, 1, 1]`, up to `[0xff, 0xff, 0xff, 0xff]`) the expectd result would be `[0, 0, 0, 0, 1, 1, 1, 1, ..., 16, 16, 16, 16]` for example.

rust/src/tests/lib/aggregator.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn new() -> ByteAggregator {
+        ByteAggregator { weights: vec![] }
+    }
+}
+
+impl Aggregator for ByteAggregator {
+    type Error = ();
+
+    type AddWeightsFut = Pin<Box<dyn Future<Output = Result<(), ()>> + Send>>;
+    type AggregateFut = Pin<Box<dyn Future<Output = Result<Bytes, ()>> + Send>>;
```
little-dude: I don't think we need a `Pin<Box<_>>` here. Since the computation is actually synchronous a `futures::future::Ready` should work.

rust/src/tests/lib/aggregator.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl Aggregator for ByteAggregator {
+    type Error = ();
+
+    type AddWeightsFut = future::Ready<Result<(), ()>>;
+    type AggregateFut = future::Ready<Result<Bytes, ()>>;
+
+    fn add_weights(&mut self, weights: Bytes) -> Self::AddWeightsFut {
+        weights.into_iter().for_each(|el| self.weights.push(el));
```
little-dude: Sorry for not seeing that earlier but I think this could be simplified:

```
self.weights.extend(weights.into_iter())
```

or something like that.

Robert-Steiner: Fixed 😊

PR #354, opened 2020-03-27
PB-537 pet tests

janpetschexain:
**References**
- [PB-537](https://xainag.atlassian.net/browse/PB-537) including [PB-551](https://xainag.atlassian.net/browse/PB-551), [PB-552](https://xainag.atlassian.net/browse/PB-552), [PB-553](https://xainag.atlassian.net/browse/PB-553), [PB-554](https://xainag.atlassian.net/browse/PB-554), [PB-555](https://xainag.atlassian.net/browse/PB-555), [PB-556](https://xainag.atlassian.net/browse/PB-556), [PB-557](https://xainag.atlassian.net/browse/PB-557)

**Summary**
- add unit tests for message composition
- add unit tests for message validation
- add unit tests for utils
- add message wrappers

**Open Tasks**
- follow up task: unify the message buffer structure ([PB-588](https://xainag.atlassian.net/browse/PB-588))

### DIFF
--- a/rust/src/pet/participant.rs
+++ b/rust/src/pet/participant.rs
@@ -40,7 +40,7 @@ impl Participant {
         // crucial: init must be called before anything else in this module
         sodiumoxide::init()
             .and(Ok(Default::default()))
-            .or(Err(PetError::InvalidMessage))
+            .or(Err(PetError::InsufficientSystemEntropy))
     }
 
     /// Compute the "sum" and "update" signatures.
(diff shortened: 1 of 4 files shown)


rust/src/pet/participant.rs
```diff
@@ -206,12 +220,12 @@ impl<'tag, 'cert, 'sign_, 'url, 'dict> UpdateBoxBuffer<'tag, 'cert, 'sign_, 'url
...
-            certificate,             // 0 bytes (dummy)
-            &signature_sum.0[..],    // 64 bytes
-            &signature_update.0[..], // 64 bytes
-            model_url,               // 32 bytes (dummy)
-            dict_seed,               // 112 * dict_sum.len() bytes
+            b"update",                 // 6 bytes
+            certificate,               // 0 bytes (dummy)
+            signature_sum.as_ref(),    // 64 bytes
+            signature_update.as_ref(), // 64 bytes
+            model_url,                 // 32 bytes (dummy)
```
little-dude: This is fine as a very short term solution, but we should define a proper encoding for variable length fields. A simple way to do that is to use TLV encoding. I guess certificates have a variable length as well? If so we should put it at the end of the message, with the other variable length fields.

janpetschexain: the `dummy` values are just placeholders for now, as soon as they become meaningful it definitely makes sense to encode them properly. certificates are most likely going to be variable in size. regarding the full message length, it also depends on if we are going to use padding for a constant message length (for attack vector reasons) or not.

PR #356, opened 2020-03-30
PB-546 Swagger PET API

Robert-Steiner:
**Reference:**

- [PB-546](https://xainag.atlassian.net/browse/PB-546)

**Summary:**

- created a first API version that includes the PET protocol


swagger/coordinator.yml
```diff
@@ -0,0 +1,263 @@
...
+      type: object
+      properties:
+        coord_pk:
+          description: The public key of the coordinator.
+          type: string
+          format: byte
+          minLength: 32
+          maxLength: 32
+        sum:
+          description: The sum threshold of the current round.
```
little-dude, line 224: The description seems a bit unclear to me. Maybe we could directly adapt this from the PET protocol document:

> fraction of participants to be selected in this round for computing the unmasking sum

And something similar for `update`

swagger/coordinator.yml
```diff
@@ -0,0 +1,214 @@
...
+        format: byte
+    PublishEncryptedSeedsMessage:
+      type: object
+      additionalProperties:
+        type: array
+        items:
+          type: string
+          format: byte
+          minLength: 32
+          maxLength: 32
```
janpetschexain, line 214: this should be of min/max length `80`.

this is because the encryption of the seed produces a header of 48 bytes (32 bytes for the public part of an ephemeral key and 16 bytes for a MAC). the ciphertext of the seed itself stays 32 bytes, but without the header it is undecryptable even for the intended receiver.

Robert-Steiner: Fixed

PR #357, opened 2020-03-31
PB-588 unify buffers

janpetschexain:
**References**
- [PB-588](https://xainag.atlassian.net/browse/PB-588) and [PB-592](https://xainag.atlassian.net/browse/PB-592)

**Summary**
- implement an unified `RoundBoxBuffer` for `RoundBox` buffering
- define a `MessageBoxBuffer` trait for `MessageBox` buffering and implement it for the unified `SumBoxBuffer`, `UpdateBoxBuffer` and `Sum2BoxBuffer`
- implement an unified `RoundBox` for round message encryption and decryption
- define a `MessageBox` trait for sum/update/sum2 message encryption and decryption and implement it for the unified `SumBox`, `UpdateBox` and `Sum2Box`
- implement a `MessageBuffer` for `Message` buffering and a unified `Message` wrapper for encryption and decryption
- compose messages on participant side and validate messages on coordinator side
- add methods to proceed in the phases of the pet protocol

**Open Tasks**
- temporary removal of tests, adjust tests in [PB-590](https://xainag.atlassian.net/browse/PB-590)


rust/src/pet/coordinator.rs
```diff
@@ -1,887 +1,314 @@
...
-        } else {
-            // unknown message type
-            Err(PetError::InvalidMessage)
-        }
-    }
+    masks: Counter<Vec<u8>>,
 }
 
 impl Default for Coordinator {
     fn default() -> Self {
```
little-dude, line 50: In Rust there's an implicit convention that `Default::default` always returns the same thing. This implementation breaks this convention because `seed`, and `{encr,sign}_{pk,sk}` are random. I think it would be best to move this initialization into a `new()` method instead.

Unrelated nit: I think a shorter version would be slightly more readable.

```rust
    fn default() -> Self {
        let (encr_pk, encr_sk) = box_::gen_keypair();
        let (sign_pk, sign_sk) = sign::gen_keypair();
        Self {
            encr_pk,
            encr_sk,
            sign_pk,
            sign_sk,
            sum: 0.01,
            update: 0.1,
            seed: randombytes(32),
            min_sum: 1,
            min_update: 3,
            phase: Phase::Idle,
            dict_sum: HashMap::new(),
            dict_seed: HashMap::new(),
            masks: Counter::new(),
        }
    }
```

rust/src/pet/participant.rs
```diff
@@ -21,29 +25,55 @@ enum Task {
...
     signature_sum: sign::Signature,    // 64 bytes
     signature_update: sign::Signature, // 64 bytes
 
-    // other
+    // round parameters
     task: Task,
 }
 
+impl Default for Participant {
+    fn default() -> Self {
```
little-dude, line 43: Like for `Coordinator`, I think this should be move to a `new()` method instead of using the `Default` trait.

rust/src/pet/message/round.rs
```diff
@@ -0,0 +1,142 @@
...
+        Self {
+            bytes: vec![0_u8; len],
+        }
+    }
+}
+
+impl<B: AsRef<[u8]>> RoundBoxBuffer<B> {
+    /// Create a round box buffer from `bytes`. Fails if the `bytes` don't conform to the expected
+    /// round box length `exp_len`.
+    fn from(bytes: B, exp_len: usize) -> Result<Self, PetError> {
```
little-dude, line 31: Why do we have to pass `exp_len` here? Don't we already know that it's 65 bytes? It seems a bit weird to have the caller pass the expected length, because if they already know it, they do the check themself.

rust/src/pet/message/round.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /// Encrypt the round box.
+    pub fn seal(&self, pk: &box_::PublicKey) -> Vec<u8> {
+        let bytes = self.serialize();
+        sealedbox::seal(&bytes, pk)
+    }
+}
+
+impl RoundBox<box_::PublicKey, sign::PublicKey> {
```
little-dude, line 105: As far as I can see, this impl is never used. Do we really need it?

rust/src/pet/message/round.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use std::ops::Range;
+
+use sodiumoxide::crypto::{box_, sealedbox, sign};
+
+use super::{ROUND_TAG, TAG_RANGE};
+use crate::pet::PetError;
+
+// round box field ranges
+const ENCR_PK_RANGE: Range<usize> = 1..33; // 32 bytes
```
little-dude, line 11: Usually we try to align fields on 4 or 8 bytes boundaries. I don't know if it really has an impact here, but it would be interesting to benchmark it, and potentially add some padding.

rust/src/pet/message/sum.rs
```diff
@@ -0,0 +1,147 @@
...
+
+// sum box field ranges
+const EPHM_PK_RANGE: Range<usize> = 65..97; // 32 bytes
+
+/// Mutable and immutable buffer access to sum box fields.
+struct SumBoxBuffer<B> {
+    bytes: B,
+}
+
+impl SumBoxBuffer<Vec<u8>> {
```
little-dude, line 18: nit: I don't know if it's really necessary to have an impl for `Vec<u8>` specifically. Maybe:

```rust
impl<B> for SumBoxBuffer<B> {
    fn new(bytes: B) -> Self {
        Self { bytes }
    }
}
```

would be enough? That's personal taste though, so feel free to ignore.

rust/src/pet/message/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    fn signature_sum_mut(&mut self) -> &mut [u8] {
+        &mut self.bytes_mut()[SIGN_SUM_RANGE]
+    }
+}
+
+#[allow(clippy::len_without_is_empty)]
+/// Encryption of message boxes.
+pub trait MsgBoxEncr {
+    /// Get the length of the serialized message box.
+    fn len(&self) -> usize;
```
little-dude, line 77: It doesn't feel right to make implementors implement this method, while it's actually only used in `serialize()` which is _also_ implemented manually. It would make more sense if `len()` was called in some blanket method impl, or if `MsgBoxEncr` was used somewhere ("used" as in "used a a trait bound in some struct, trait, or impl).

Edit: oops. Just as I was finishing this comment I noticed `MsgBoxEncr` _is_ being used, and in particular, `MsgBoxEncr::len` is used in `Message::len`. So it does make sense to have that method.

rust/src/pet/message/round.rs
janpetschexain: it's used in the `Message` wrapper (https://github.com/xainag/xain-fl/blob/PB-588-unify_buffers/rust/src/pet/message/mod.rs#L218-L271), which is in turn used in the message validation methods of the pet coordinator.

rust/src/pet/message/round.rs
janpetschexain: good idea, we should keep this in mind.

rust/src/pet/message/round.rs
janpetschexain: my intention was to have a similar buffer interface for for `round` as for `sum`/`update`/`sum2`. the former could go without the `exp_len`, but the latter need this (update already now, sum/sum2 as soon as certificates are not dummy anymore).

since the `Buffer`s are only privately available to the `Box`es, there is no risk in wrongly calling them from the outside. not sure what's the best approach here.

rust/src/pet/participant.rs
janpetschexain: i'll split that.

rust/src/pet/coordinator.rs
janpetschexain: i'll split that.

rust/src/pet/message/sum.rs
janpetschexain: with a generic buffer of type `B` there is no guarantee that any of the accessors can actually be used. what i would like to express is that `B` should be trait-bounded by at least one of the bounds `AsRef<[u8]> + ?Sized` or `AsMut<[u8]>`, but i'm not sure how to do this without having separate methods `new` and `new_mut` which i would like to avoid.

therefore i went with the type specific `Vec<u8>` impl. not sure what's the best approach here.

rust/src/pet/message/mod.rs
janpetschexain: one of the overall roots of complexity right now is the combined use of `sealedbox` and `box_` resulting in every message being a composition of a `RoundBox` and a `MessageBox`. a future simplification i have in mind would be to use only `sealedbox` and no `box_` anymore, this would effectively lead to a combination of both boxes into a single box and `Message` would then contain only one struct.

the main trade-off is that messages could never be decrypted by the sender when using a `sealedbox`, only by the receiver. but since we don't require the sender to decrypt its own message again, i guess this is ok. the current implementation follows the pet protocol quite strictly in this regard, therefore i want to check this with Michal before making the simplification.

rust/src/pet/message/sum.rs
little-dude: No need for any trait bound. If someone makes a buffer out of something that does not impl `AsRef<[u8]>` or `AsMut<[u8]>`, and then tries to use a getter or setter, the compiler is going to yell at them, that the method is not implemented for that type.

janpetschexain: i just noticed that we are mixing up two different impl here. the `from` method is already generic over `B`:
```
impl<B: AsRef<[u8]>> SumBoxBuffer<B> {
    /// Create a sum box buffer from `bytes`. Fails if the `bytes` don't conform to the expected
    /// sum box length `exp_len`.
    fn from(bytes: B, exp_len: usize) -> Result<Self, PetError> {
        (bytes.as_ref().len() == exp_len)
            .then_some(Self { bytes })
            .ok_or(PetError::InvalidMessage)
    }
}
```
whereas the `new` method cannot be generic over `B` since it has no input (aside from the length parameter):
```
impl SumBoxBuffer<Vec<u8>> {
    /// Create an empty sum box buffer of size `len`.
    fn new(len: usize) -> Self {
        Self {
            bytes: vec![0_u8; len],
        }
    }
}
```

PR #362, opened 2020-04-06
cleanup README files

little-dude:
- use a separate README for the repo and for readthedocs
- remove the implementation details from the repo's main README. That
  part of the documentation belong to the Rust documentation and was
  slightly outdated anyway
- remove the keras example README, which was outdated and redundant
  with the instructions in the repo's README


README.md
```diff
@@ -1,36 +1,61 @@
...
+## Resources
 
-This aggregator handles very different types of requests than the
-coordinator itself: weight distribution involves streaming high
-volumes of data over the network, and aggregation involves periodic
-CPU intensive tasks.
+- [Whitepaper](https://www.xain.io/federated-learning-technology)
+- [Source code](https://github.com/xainag/xain-fl/)
+- [Rest API reference](https://xain-fl.readthedocs.io/en/latest/api)
+- [docker images](https://hub.docker.com/repository/docker/xain/xain-fl/)
```
Robert-Steiner, line 18: When I open this link, the browser redirects me to the login page.
Can you try this link `https://hub.docker.com/r/xain/xain-fl`.

Robert-Steiner, line 18:
```suggestion
- [Docker images](https://hub.docker.com/repository/docker/xain/xain-fl/)
```

little-dude: Done. Btw we should have a README for the dockerhub page. Right now it just takes the README from the repo's root.

little-dude:
> I'm just not sure if we should have two separate READMEs. Personally I prefer that both are the same.

It would be nice not to duplicate anything I agree. But where should the instructions for compiling and troubleshooting go then? I think they belong to the repo, while xain-fl.readthedocs.io should be a sort of entrypoint for our documentation, that just points to the relevant resource for each area (SDK, API, backend, whitepaper, etc.)

little-dude:
Maybe the README at the root of the repo should start with a link to xain-fl.readthedocs.io, and then contain the instructions for the repo, instead of repeating what's in `docs/README.md`?

little-dude:
@Robert-Steiner what do you think of that last commit that moves the repo specific instructions to a CONTRIBUTING.md, and adds a direct link to the rtd doc? Just a suggestion.

Robert-Steiner:
I like that 
What do you think @janpetschexain @finiteprods @rsaffi ?

janpetschexain:
> I like that
> What do you think @janpetschexain @finiteprods @rsaffi ?

me too

Robert-Steiner:
But I'm totally fine going with the first version

finiteprods:
Yeah I’m in favour of providing a link over repeating ourselves 👍

little-dude:
I don't mind either way, so if you have even a slight preference for the CONTRIBUTING.md version let's go with that :)

PR #364, opened 2020-04-06
PB-571 First dropout test

Robert-Steiner:
**References:**

- [PB-571](https://xainag.atlassian.net/browse/PB-571)

**Summary:**

- added first dropout test

### DIFF
--- a/rust/src/tests/coordinator.rs
+++ b/rust/src/tests/coordinator.rs
@@ -133,7 +133,7 @@ async fn dropout_1_participant_during_training() {
         .returning(|_| future::ready(Ok(Ok(()))));
     service_handle.end_training(id_1, true).await;
 
-    // Create a third client. The third customer should be selected by the coordinator and
+    // Create a third client. The third client should be selected by the coordinator and
     // be able to participate in the training session.
     let id_3 = service_handle.rendez_vous_accepted().await;
     let round = service_handle.heartbeat_selected(id_3).await;
--- a/rust/src/tests/coordinator.rs
+++ b/rust/src/tests/coordinator.rs
@@ -85,7 +85,7 @@ async fn dropout_1_participant_during_training() {
     };
     let (rpc_client, service_handle, _join_handle) = start_service(settings);
 
-    // Create one client. Since min_clients is 2, the heartbeat response should be `StandBy`.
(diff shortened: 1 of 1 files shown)


rust/src/tests/coordinator.rs
```diff
@@ -74,3 +74,89 @@ async fn full_cycle_1_round_1_participant() {
...
     }
 }
+
+#[tokio::test]
+async fn dropout_1_participant_during_training() {
+    let settings = FederatedLearningSettings {
+        rounds: 1,
+        participants_ratio: 1.0,
+        min_clients: 2,
+        heartbeat_timeout: 5,
```
little-dude, line 84: We could probably use a much lower value. I guess it need to be an integer though, so maybe `1`?

rust/src/tests/coordinator.rs
```diff
@@ -74,3 +74,89 @@ async fn full_cycle_1_round_1_participant() {
...
+    sleep_ms(2500).await;
+    service_handle.heartbeat(id_1).await;
+    sleep_ms(2500).await;
+    service_handle.heartbeat(id_1).await;
+    // The second client should be dropped.
+
+    // pretend the first client trained and sent its weights to the
+    // aggregator. The aggregator now sends an end training requests
+    // to the coordinator RPC server that we fake with the
+    // service_handle. The service should then trigger the aggregation
```
little-dude, line 128: If only one participant uploaded their weights, aggregation shouldn't be triggered. Or am I missing something?

rust/src/tests/coordinator.rs
```diff
@@ -74,3 +74,93 @@ async fn full_cycle_1_round_1_participant() {
...
+    // The second client should be accepted but not re-selected (for current round)
+    // by the coordinator.
+    let hb_resp = service_handle.heartbeat(id_2).await;
+    assert_eq!(hb_resp, HeartBeatResponse::StandBy);
+
+    // The third client finished training.
+    service_handle.end_training(id_3, true).await;
+
+    // Trigger aggregation.
+    rpc_client
```
little-dude, line 152: I'm surprised this works: `service_handle.end_training` should trigger the aggregation, and since you only set `rpc_client.expect_aggregate()` _after_, I would expect this to fail. I'll do more tests on this after lunch if you don't mind, there might be something wrong with the mock.

PR #369, opened 2020-04-08
add a to facilitate releasing a new version

little-dude:
This is an interactive script that automates some of the checks, and is able create a tag locally. The user still has to push manually.

It's not particularly beautiful because I'm not well versed in bash or in git porcelain commands, but `shellcheck` doesn't emit any warning, so it's not _that_ bad.

### DIFF
--- a/scripts/bump_version.sh
+++ b/scripts/bump_version.sh
@@ -1,0 +1,295 @@
+#!/usr/bin/env bash
+
+WORKDIR="$(git rev-parse --show-toplevel)"
+# Save the git HEAD before running the script
+HEAD=
+# Latest tag
+PREV_TAG=
+# Commit that corresponds to the latest tag
+PREV_TAGGED_COMMIT=
+
+# Latest version numbers
+PREV_MAJOR=
+PREV_MINOR=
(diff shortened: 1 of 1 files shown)


scripts/bump_version.sh
```diff
@@ -0,0 +1,261 @@
...
+        git --no-pager diff "${PREV_TAGGED_COMMIT}" HEAD CHANGELOG.md
+    }
+
+    if [ "$(diff | wc -l)" -eq 0 ] ; then
+        echo "error: the CHANGELOG has not been updated since ${PREV_TAG}" 2>&1
+    fi
+
+    echo "The CHANGELOG has been updated since ${PREV_TAG}"
+    diff
+    echo "Do the change above look correct for v$(version)"
```
finiteprods, line 142:
```suggestion
    echo "Does the change above look correct for v$(version)"
```

finiteprods: or maybe it's "Do the changes above ..."

little-dude: Thanks I made the change locally :)

PR #375, opened 2020-04-09
remove `then_some`

little-dude:
`bool.then_some()`'s future is uncertain so let's remove them.
`bool.then()` will likely be stabilized but is a bit more verbose, and
I personally think the imperative form is more readable in our case
because:

- we're not interested in the `Some` case, but more in the `None`
  case that we turn into an error
- we don't want an `Option<T>` but a `Result<(), E>`, so we have to
  chain two functions, which feels unnatural: `b.then(Some).ok_or(MyErr)`

To give a concrete example:

```rust
is_eligible(signature_sum, self.sum))
        .then_some(())
        .ok_or(PetError::InvalidMessage)
```

It took me a small effort to understand when an error is returned: when `is_eligible` returns `true`, or when it returns `false`? I think is a bit more straighforward:

```rust
if !is_eligible(signature_sum, self.sum) {
	return Err(PetError::InvalidMessage)
}
Ok(())
```

### DIFF
--- a/rust/src/pet/coordinator.rs
+++ b/rust/src/pet/coordinator.rs
@@ -99,9 +99,9 @@ impl Coordinator {
 
     /// Validate and handle a sum message.
     pub fn validate_sum_message(&mut self, message: &[u8]) -> Result<(), PetError> {
-        (self.phase == Phase::Sum)
-            .then_some(())
-            .ok_or(PetError::InvalidMessage)?;
+        if self.phase == Phase::Sum {
+            return Err(PetError::InvalidMessage);
+        }
         let msg = SumMessage::open(
             message,
             &self.encr_pk,
@@ -144,9 +144,10 @@ impl Coordinator {
 
     /// Validate a certificate (dummy).
     fn validate_certificate(certificate: &[u8]) -> Result<(), PetError> {
-        (certificate == b"")
-            .then_some(())
-            .ok_or(PetError::InvalidMessage)
+        if certificate == b"" {
+            return Ok(());
+        }
+        Err(PetError::InvalidMessage)
     }
 
     /// Validate a sum signature and its implied task.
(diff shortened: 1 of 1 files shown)


little-dude:
I hope I didn't change the logic here.

I understand that this is perhaps controversial because it's about style (I do want to remove `then_some`, but we could just use `then` instead)

rust/src/pet/coordinator.rs
```diff
@@ -211,27 +225,41 @@ impl Coordinator {
...
-            });
-        })
-        .ok_or(PetError::InvalidMessage)
+        if dict_seed.len() != self.dict_sum.len() {
+            return Err(PetError::InvalidMessage);
+        }
+
+        if dict_seed.keys().all(|pk| self.dict_sum.contains_key(pk)) {
+            return Err(PetError::InvalidMessage);
+        }
```
janpetschexain, line 234: logic was reverted, should be `if !dict_seed.keys().all(|pk| self.dict_sum.contains_key(pk)) { return Err(...) }`

alternative, this can be rewritten to `if dict_seed.keys().any(|pk| !self.dict_sum.contains_key(pk)) { return Err(...) }`

rust/src/pet/coordinator.rs
```diff
@@ -211,27 +225,41 @@ impl Coordinator {
...
+
+        for (pk, seed) in dict_seed.iter() {
+            self.dict_seed
+                .get_mut(pk)
+                // UNWRAP_SAFE: It is fine to unwrap because:
+                //
+                //   - we checked just above that self.dict_sum has
+                //     all the keys
+                //   - self.dict_seed has the same keys than
+                //     self.dict_sum, since free_dict_sum()
```
janpetschexain, line 244: typo: `freeze_...` instead of `free_...`

janpetschexain:
> I hope I didn't change the logic here.
> 
> I understand that this is perhaps controversial because it's about style (I do want to remove `then_some`, but we could just use `then` instead)

i like the `(bool_expr).then_some(..)` approach because it collects all conditions in one place such that one can quickly see whats needed to succeed/fail. in case of
```
if a {
    return Err(err)
}
if b {
    return Err(err)
}
if c {
    return Err(err)
}
Ok(())
```
that information is quite scattered. maybe we can instead use
```
if a && b && c {
    return Err(err)
}
Ok(())
```
this makes all conditions available in one place, gets rid of `then_some()` and is still no overhead in evaluation of the conditions, since `&&` and `||` are short-circuiting anyways.

regarding the use of `then()` over `then_some()`, it feels somehow strange to me to use a closure if i only want to return a fixed value and not an output of a function, but that wouldn't be a major issue.

PR #380, opened 2020-04-16
Coordinator refactoring

little-dude:
Each commit can be reviewed separately. The main change is 3ea765f.

I'd like to _maybe_ refactor further at some point, taking inspiration from this blog post https://hoverbear.org/blog/rust-state-machine-pattern/

### DIFF
--- a/rust/src/coordinator.rs
+++ b/rust/src/coordinator.rs
@@ -3,6 +3,7 @@
 use std::{
     collections::{HashMap, HashSet},
     default::Default,
+    iter,
 };
 
 use counter::Counter;
@@ -242,9 +243,7 @@ impl Coordinator {
 
     /// Update the mask dictionary.
     fn update_dict_mask(&mut self, mask_hash: &[u8]) {
-        self.dict_mask += mask_hash
-            .chunks_exact(mask_hash.len())
-            .map(|mask| mask.to_vec());
+        self.dict_mask.update(iter::once(mask_hash.to_vec()))
     }
 
     /// Freeze the mask dictionary. Returns a unique mask. Fails due to insufficient sum
--- a/rust/src/coordinator.rs
+++ b/rust/src/coordinator.rs
@@ -199,17 +199,13 @@ impl Coordinator {
         self.dict_sum.insert(*encr_pk, *ephm_pk);
     }
 
-    /// Freeze the sum dictionary. Fails due to insufficient sum participants.
-    fn freeze_dict_sum(&mut self) -> Result<(), PetError> {
-        (self.dict_sum.len() >= self.min_sum)
-            .then(|| {
-                self.dict_seed = self
-                    .dict_sum
-                    .keys()
-                    .map(|pk| (*pk, HashMap::<box_::PublicKey, Vec<u8>>::new()))
-                    .collect();
-            })
-            .ok_or(PetError::InsufficientParticipants)
+    /// Freeze the sum dictionary.
+    fn freeze_dict_sum(&mut self) {
+        self.dict_seed = self
+            .dict_sum
+            .keys()
+            .map(|pk| (*pk, HashMap::<box_::PublicKey, Vec<u8>>::new()))
+            .collect();
     }
 
     /// Update the seed dictionary.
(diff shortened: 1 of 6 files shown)


little-dude:
> minor request: can you also see if your refactoring applies to `Participant.rs` to include as part of this MR? e.g. I've seen some `url` / `hash` occurrences there.

Good point, will do

rust/src/coordinator.rs
```diff
@@ -230,31 +227,24 @@ impl Coordinator {
...
-        .ok_or(PetError::InsufficientParticipants)
-    }
-
     /// Update the mask dictionary.
-    fn update_dict_mask(&mut self, mask_url: &[u8]) {
-        self.dict_mask += mask_url
-            .chunks_exact(mask_url.len())
-            .map(|mask| mask.to_vec());
+    fn update_dict_mask(&mut self, mask_hash: &[u8]) {
+        self.dict_mask.update(iter::once(mask_hash.to_vec()))
```
janpetschexain, line 232: what's the reason for the change from `+=` to `.update()`? they are equivalent for `Counter<_>`, but the `+=` better conveys the semantics of counting, imo.

rust/src/coordinator.rs
```diff
@@ -230,31 +227,24 @@ impl Coordinator {
...
     fn freeze_dict_mask(&self) -> Result<Vec<u8>, PetError> {
         let counts = self.dict_mask.most_common();
-        (counts.iter().map(|(_, count)| count).sum::<usize>() >= self.min_sum
-            && (counts.len() == 1 || counts[0].1 > counts[1].1))
-            .then(|| counts[0].0.clone())
-            .ok_or(PetError::InsufficientParticipants)
+
+        // it is an error to call freeze_dict_mask if we didn't
+        // receive a single mask
+        assert!(!counts.is_empty());
```
janpetschexain, line 241: this is already covered by `has_enough_masks()`, the `mask_count` yields `0` in this case.

rust/src/coordinator.rs
```diff
@@ -317,32 +317,60 @@ impl Coordinator {
         self.phase = Phase::Sum;
     }
 
-    /// End the sum phase and proceed to the update phase. Fails due to insufficient sum
-    /// participants.
-    fn proceed_phase_update(&mut self) -> Result<(), PetError> {
-        self.freeze_dict_sum()?;
+    /// Check whether there are enough sum participants for starting
+    /// the update phase.
+    fn has_enough_sum_participants(&self) -> bool {
```
janpetschexain, line 322: i would suggest a more consistent naming for those three methods, like
```
has_enough_sums()
has_enough_updates()
has_enough_masks()
```
or alternatively
```
has_enough_sum_messages()
has_enough_update_messages()
has_enough_sum2_messages()
```

rust/src/lib.rs
```diff
@@ -20,4 +20,5 @@ pub enum PetError {
     InsufficientSystemEntropy,
     InvalidMessage,
     InsufficientParticipants,
+    MaskSelectionFailed,
```
janpetschexain, line 23: maybe `AmbiguousMasks` would fit better here

rust/src/service.rs
```diff
@@ -60,10 +60,10 @@ impl Service {
     //
     // when we fail to handle a message, we should send back and error
     // so that the participant is informed. One problem currently is
-    // that the `validate` methods don't return specific errors.
+    // that the `handle` methods don't return specific errors.
```
janpetschexain, line 63: on an important side note: the coordinator **must never** inform a participant about the specific reason why message validation failed. that would open gates to various attacks. instead the coordinator must ignore that message, but it can/should log the failure internally.
it is ok to, for example, send an ACK to the participant if the message was received. it might also send a generic failure message if validation fails, but there is no reason to do so and it's better to not open up to potential attacks.

rust/src/coordinator.rs
little-dude: Having a `+` operator with operands of that of different types was very confusing, especially with types that usually don't work with `+`: `Dict<K,V> + `Iterator<Item=V>` doesn't make much sense I think.

rust/src/coordinator.rs
little-dude: Yeah that's why I made is an `assert!`. But you're right, maybe it's useless I'll remove it.

rust/src/coordinator.rs
little-dude: ah very nice. I have a slight preference the first version. How about you (or @finiteprods) ?

rust/src/lib.rs
little-dude: Thanks!

rust/src/coordinator.rs
finiteprods: same here

janpetschexain: same

little-dude: alright it's settled then.

rust/src/coordinator.rs
janpetschexain: mhh ok, that's probably quite subjective. i'm ok going with `.update()`.

PR #381, opened 2020-04-17
coordinator: introduce type aliases & some renaming

little-dude:
This is based on https://github.com/xainag/xain-fl/pull/380, I split it to make the changes easier to review.

-------

Type aliases help the reader understanding the nature of the data
being manipulated. For instance in

```rust
type UpdateDict = HashMap<SumParticipantPublicKey, HashMap<UpdateParticipantPublicKey, EncryptedMaskingSeed>>;
```

it is clear that this is a dictionary where the keys identify sum participant, and the values are nested dictionaries where the keys identify update participants. While in the following, it is not so clear, and one has to look for a place where the map is being manipulated to try to understand what the keys refer to:

```rust
HashMap<box_::PublicKey, HashMap<box_::PublicKey, Vec<u8>>>
```

This commit also renames the `Coordinator.dict_seed` attribute to
`Coordinator.dict_update`. There are two types of seed dictionaries:
the dictionaries sent by the update participant, and the dictionary
built by the coordinator from these updates. It is easy to confuse
them, so we chose to name the dictionaries sent by the participants
"seed dictionary" and the global dictionary built by the coordinator
during the update phase "update dictionary".

Finally, the renaming aforementioned lead to renaming
`Coordinator.update_dict_seed` to `Coordinator.update_dict_update`
which sounds weird, so we renamed these methods as well:

- update_sum_dict => add_sum_participant
- update_dict_seed => add_seed_dict
- update_dict_mask => add_mask_hash


little-dude:
I'm not sure about the terminology choice about `update_dict` and `seed_dict`. Maybe we should reverse it: `seed_dict` should be the global dictionary that we build out of `update_dict` (or just `update` even??) ? I don't really mind, I just want to have a clear naming distinction between the two types of dictionaries. Currently "seed" is used for both which is a little confusing.

rust/src/coordinator.rs
```diff
@@ -574,85 +612,82 @@ mod tests {
...
-        let updates = iter::repeat_with(|| {
-            let seed = randombytes(32);
-            let upd_encr_pk = box_::PublicKey::from_slice(&randombytes(32)).unwrap();
-            let upd_dict_seed = dict_sum
+        let updates = iter::repeat_with(|| generate_update(&dict_sum))
+            .take(min_update)
+            .collect::<Vec<(UpdateParticipantPublicKey, SeedDict)>>();
+
+        let mut dict_update = UpdateDict::new();
+        for sum_encr_pk in dict_sum.keys() {
```
little-dude, line 657: @janpetschexain sorry, I know you like combinators, but that was just too much for my brain so I converted the outer loop into an actual `for` loop :smile:

rust/src/coordinator.rs
```diff
@@ -19,6 +19,41 @@ use super::{
...
+/// Dictionary created during the first phase of the protocol. It maps
+/// the public key of every sum participant to the ephemeral public
+/// key that participant generated.
+type SumDict = HashMap<SumParticipantPublicKey, SumParticipantEphemeralPublicKey>;
+
+/// Public key that identifies an update participant
+type UpdateParticipantPublicKey = box_::PublicKey;
+
+/// Seed dictionaries are sent by update participants. They contain
+/// the participant's seed, encrypted with the public key of each sum
```
janpetschexain, line 39: `encrypted with the ephemeral public key` (the distinction between the public key and the ephemeral public key is important here)

rust/src/coordinator.rs
```diff
@@ -19,6 +19,41 @@ use super::{
...
+/// all the update participants to each sum participant.
+type UpdateDict =
+    HashMap<SumParticipantPublicKey, HashMap<UpdateParticipantPublicKey, EncryptedMaskingSeed>>;
+
+/// Masking seed generated by an update participant, encrypted with
+/// the ephemeral public key of a sum participant
+pub type EncryptedMaskingSeed = Vec<u8>;
+
+/// A 32 bytes hash of that identifies a model mask computed by an
+/// update participant.
```
janpetschexain, line 54: the masks are computed by the sum participants

rust/src/coordinator.rs
janpetschexain: no worries, i probably overdid it in some places^^

rust/src/coordinator.rs
little-dude: oops good catch, thanks

rust/src/coordinator.rs
little-dude: Thank you. I fixed it.

janpetschexain:
> I'm not sure about the terminology choice about `update_dict` and `seed_dict`. Maybe we should reverse it: `seed_dict` should be the global dictionary that we build out of `update_dict` (or just `update` even??) ? I don't really mind, I just want to have a clear naming distinction between the two types of dictionaries. Currently "seed" is used for both which is a little confusing.

i have no preferences either. maybe another distinction could be `seed_dict` for the global dict and `seed_update` or `local_seed_dict` for the local dict. in the end both local and global dicts contain the seeds, so it makes sense to keep that in their names somehow.

little-dude:
I pushed a commit with the "local" terminology suggestion. With this we would have:

```rust
/// Seed dictionaries are sent by update participants. They contain
/// the participant's seed, encrypted with the ephemeral public key of
/// each sum participant.
type LocalSeedDict = HashMap<SumParticipantPublicKey, EncryptedMaskingSeed>;

/// The update dictionary is built from the seed dictionaries sent by
/// the update participants. It maps the encrypted masking seeds of
/// all the update participants to each sum participant.
type SeedDict = HashMap<SumParticipantPublicKey, HashMap<UpdateParticipantPublicKey, EncryptedMaskingSeed>>;
```

I also renamed `has_enough_updates()` into `has_enough_seeds` but perhaps I shouldn't have? I wonder if there's not too much "seeds" flying around now although I agree conceptually it makes sense to have the word "seed" for both dictionaries.

So yeah I'm not sure here. @finiteprods @Robert-Steiner what's clearer to you? Before or after 6cef001  (please ignore the failure for now, this is just to see what the renaming would look like)

finiteprods:
> So yeah I'm not sure here: on one hand, it makes sense to have the word "seed" in both names, but on the other hand, that makes the distinctions between both types less obvious. @finiteprods @Robert-Steiner what's clearer to you? Before or after [6cef001](https://github.com/xainag/xain-fl/commit/6cef001b7f7b8b7f25124066b4918c7d5b785a53) (please ignore the failure for now, this is just to see what the renaming would look like)

after is slightly better than before, but not by much. i have 2 suggestions:

1. since you like the naming convention of a `Dict` suffix for `HashMap`s, I would call these `SeedDict` and `SeedDictDict` respectively.
2. just call them what they are in the spec. so `El` and `C`.

rust/src/coordinator.rs
```diff
@@ -19,6 +19,41 @@ use super::{
...
+type UpdateParticipantPublicKey = box_::PublicKey;
+
+/// Seed dictionaries are sent by update participants. They contain
+/// the participant's seed, encrypted with the ephemeral public key of
+/// each sum participant.
+type LocalSeedDict = HashMap<SumParticipantPublicKey, EncryptedMaskingSeed>;
+
+/// The update dictionary is built from the seed dictionaries sent by
+/// the update participants. It maps the encrypted masking seeds of
+/// all the update participants to each sum participant.
```
finiteprods, line 45: whatever we end up calling these, reminder to update comment accordingly!

rust/src/coordinator.rs
```diff
@@ -19,6 +19,41 @@ use super::{
...
+type SeedDict =
+    HashMap<SumParticipantPublicKey, HashMap<UpdateParticipantPublicKey, EncryptedMaskingSeed>>;
+
+/// Masking seed generated by an update participant, encrypted with
+/// the ephemeral public key of a sum participant
+pub type EncryptedMaskingSeed = Vec<u8>;
+
+/// A 32 bytes hash of that identifies a model mask computed by a sum
```
finiteprods, line 53:
```suggestion
/// A 32-byte hash that identifies a model mask computed by a sum
```

rust/src/coordinator.rs
```diff
@@ -45,9 +80,12 @@ pub struct Coordinator {
...
 
     // round dictionaries
-    dict_sum: HashMap<box_::PublicKey, box_::PublicKey>,
-    dict_seed: HashMap<box_::PublicKey, HashMap<box_::PublicKey, Vec<u8>>>,
-    dict_mask: Counter<Vec<u8>>,
+    /// Dictionary built during the sum phase.
+    dict_sum: SumDict,
+    /// Dictionary built during the update phase.
+    dict_seed: SeedDict,
+    /// Dictionary built during the sum2 phase.
```
finiteprods, line 87: is this naming right? is a `Counter` a "dictionary"?

janpetschexain:
> > So yeah I'm not sure here: on one hand, it makes sense to have the word "seed" in both names, but on the other hand, that makes the distinctions between both types less obvious. @finiteprods @Robert-Steiner what's clearer to you? Before or after [6cef001](https://github.com/xainag/xain-fl/commit/6cef001b7f7b8b7f25124066b4918c7d5b785a53) (please ignore the failure for now, this is just to see what the renaming would look like)
> 
> after is slightly better than before, but not by much. i have 2 suggestions:
> 
> 1. since you like the naming convention of a `Dict` suffix for `HashMap`s, I would call these `SeedDict` and `SeedDictDict` respectively.
> 2. just call them what they are in the spec. so `El` and `C`.

although the double-`Dict` is a bit unwieldy, it conveys it's content quite accurately. i wouldn't go with the second suggestion, actually i challenge all of you to tell me the meaning of the `El` and `C` abbreviations without looking them up in the pet protocol ;)

rust/src/coordinator.rs
janpetschexain: it is very closely related: https://docs.rs/counter/0.4.3/counter/#treat-it-like-a-map, basically behaves like a hashmap from `T` to `int` with some added functionality.

Robert-Steiner:
> ```rust
> SeedDict
> ```

I'm fine with that name. For me, the good comment on variable is more important.
We could also name it `SeedLookupTable` or just `SeedMatrix`.

rust/src/coordinator.rs
little-dude: thanks, that's done!

little-dude:
I'm not convinced by `SeedDictDict`. I'd vote for `SeedMatrix`. But since there's not major disagreement on the current name, maybe we should just merge so that @janpetschexain can rebase his changes. We can always go back to this, the change is pretty minor.

little-dude:
Cool. Let's get this in then so that nobody is blocked. When the dust settles we can make come back to this.

PR #383, opened 2020-04-17
PB-662 simplify pet messages

janpetschexain:
**References**
- [PB-662](https://xainag.atlassian.net/browse/PB-662)

**Summary**
- simplify messages: the old structure was roughly speaking a two-fold construction `SumMessage {RoundBox, SumBox}` using a `sealedbox` and a `box_`, each with their respective buffers. this is now simplified to a `SumMessage` using only a `sealedbox` (same for update and sum2). the only minor drawback is that each message must now contain its receiver (the coordinator pk) to avoid surreptitious forwarding (but that information is available anyway and only contributes to an additional 32 bytes, which is way less than the previous construction).
- simplify/unify the buffer trait and structures
- adopt the new type naming: extends the type aliases which were introduced with #380/#381. some things are still dummies like the masks and models, which will change when masking is implemented. also make the types available where necessary by moving them to the crate root.
- adjust the tests accordingly
- remove the encryption keys from the participant: the simplification allows us to go with signature keys only for the participant.
- remove the signature keys from the coordinator: a change in the round parameter update allows us to go with encryption keys only for the coordinator. we maybe have to reintroduce them later, depending on the certificate solution we will choose (but these signature keys are going to remain the same over the course of the rounds then.)
- consistify variable naming in some places
- remove use of `then_some` and `then` from the unstable `bool_to_option` feature
- add a length-value encoding for variable-sized buffer fields: this applies currently to the `certificate`, `masked_model`, `local_dict_seed` and `mask` fields
- implement and use the `From`/`Into` and `TryFrom`/`TryInto` traits instead of custom `from` implementations

**Notes**
- the sum2 message goes with the naming of `mask` instead of `mask_hash`: we are submitting whole masks directly via the messages for the first implementation, uploading to buckets might come later. the coordinator still stores only hashed masks in the `mask_dict` for efficiency.
- same holds for `masked_model` instead of `model_url`
- access to the buffer fields is not yet time-constant wrt the number of fields, this should be part of future optimization


rust/src/message/sum.rs
```diff
@@ -24,14 +21,21 @@ impl SumMessageBuffer<Vec<u8>> {
...
         }
     }
+}
 
-    /// Create a sum message buffer from `bytes`. Fails if the `bytes` don't conform to the expected
-    /// sum message length `exp_len`.
-    fn try_from(bytes: Vec<u8>, exp_len: usize) -> Result<Self, PetError> {
-        if bytes.len() != exp_len {
-            return Err(PetError::InvalidMessage);
+impl TryFrom<Vec<u8>> for SumMessageBuffer<Vec<u8>> {
```
little-dude, line 26: Couldn't this be made generic? `T: AsRef[u8]`?

rust/src/message/sum2.rs
```diff
@@ -25,14 +59,22 @@ impl Sum2MessageBuffer<Vec<u8>> {
...
         }
     }
+}
 
-    /// Create a sum2 message buffer from `bytes`. Fails if the `bytes` don't conform to the
-    /// expected sum2 message length `exp_len`.
-    fn try_from(bytes: Vec<u8>, exp_len: usize) -> Result<Self, PetError> {
-        if bytes.len() != exp_len {
-            return Err(PetError::InvalidMessage);
+impl TryFrom<Vec<u8>> for Sum2MessageBuffer<Vec<u8>> {
```
little-dude, line 64: Same question than above: couldn't this be implemented for any `T: AsRef<[u8]>`?

rust/src/coordinator.rs
```diff
@@ -116,10 +121,15 @@ impl Coordinator {
 
     /// Validate and handle an update message.
     fn handle_update_message(&mut self, bytes: &[u8]) -> Result<(), PetError> {
-        let msg = UpdateMessage::open(bytes, &self.pk, &self.sk, self.sum_dict.len())?;
+        let msg = UpdateMessage::open(bytes, &self.pk, &self.sk)?;
+        if msg.local_seed_dict().keys().collect::<HashSet<_>>()
```
little-dude, line 125: For the future, we should try to see if there's a cheaper way to do that. This can quickly become expensive if there are many participants.

rust/src/message/sum.rs
janpetschexain: good idea, i'll try that.

rust/src/coordinator.rs
janpetschexain: an equivalent way to compare the key sets would be
```
if dict1.len() == dict2.len() && dict1.keys().all(|key| dict2.contains_key(key)) {
    // do sth
} else {
    // do other
}
```
this might me more efficient due to boolean operations only.

janpetschexain: and i've just seen a built-in way to that even more efficiently:
```
if dict1.keys.eq(dict2.keys()) {
    // do sth
} else {
    // do other
}
```
that's probably the way to  go.

little-dude: Are we guaranteed that the two iterators produced by `.keys()` iterate through the keys in the same order? Otherwise it's not strictly equivalent.

janpetschexain: i just tested this and unfortunately the `eq()` depends on the order whereas `keys()` produces an iterator of arbitrary order. i guess we have to go with the other approach then.

rust/src/message/sum.rs
janpetschexain: for the records, since that doesn't work due to potential impl conflicts, these are the links @little-dude found regarding this topic:
- https://github.com/rust-lang/rust/issues/50133
- https://old.reddit.com/r/rust/comments/c1yubv/why_this_code_does_not_compile/ergj6hw/

janpetschexain:
> looks good. i like the use of `_range` functions defined in terms of the previous range in the sequence. just as a minor suggestion: maybe you could define a helper e.g.
> 
> ```rust
> fn range(prev_range, width) {
>     prev_range.end .. prev_range.end + width
> }
> ```
> 
> since you use this pattern a lot, and currently you are making 2 function calls (to the previous range) per `range` call.

i'm currently working on some optimizations to make the access to the message buffer fields time-constant wrt to the number of variable-sized fields. the suggested change will be included in the next mr. the range checks in the `try_from` impl will change as well.

PR #384, opened 2020-04-20
use Borrow to implement messages

little-dude:
Instead of having two blocks:

```rust
impl Foo<&A, &B, &C> {}
impl Foo<A, B, C> {}
```

we can use the `Borrow` trait, which has this precise purpose. To
quote the [book](https://doc.rust-lang.org/1.27.2/book/first-edition/borrow-and-asref.html#which-should-i-use):

> Choose Borrow when you want to abstract over different kinds of
> borrowing, or when you’re building a data structure that treats
> owned and borrowed values in equivalent ways

For convenience, this commit also introduces a `Certificate`
placeholder, just so that we don't have `Vec<u8>` hanging around.

Finally, this commit renames various `Foo:from` methods into
`Foo::new`. By convention, `from()` is kind of reserved for From
implementations. Perhaps we could have `from_parts()` and
`into_parts()` instead.

### DIFF
--- a/rust/src/message/sum.rs
+++ b/rust/src/message/sum.rs
@@ -1,8 +1,8 @@
-use std::ops::Range;
+use std::{borrow::Borrow, ops::Range};
 
 use sodiumoxide::crypto::{box_, sealedbox, sign};
 
-use super::{MessageBuffer, CERTIFICATE_BYTES, SUM_TAG, TAG_BYTES};
+use super::{Certificate, MessageBuffer, CERTIFICATE_BYTES, SUM_TAG, TAG_BYTES};
 use crate::{
     CoordinatorPublicKey, CoordinatorSecretKey, ParticipantTaskSignature, PetError,
     SumParticipantEphemeralPublicKey, SumParticipantPublicKey, SumParticipantSecretKey,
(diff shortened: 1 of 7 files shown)


rust/src/message/sum.rs
```diff
@@ -61,16 +61,28 @@ impl<B: AsRef<[u8]> + AsMut<[u8]>> SumMessageBuffer<B> {
 
 #[derive(Clone, Debug, PartialEq)]
 /// Encryption and decryption of sum messages.
-pub struct SumMessage<K, C, S, E> {
+pub struct SumMessage<K, C, S, E>
+where
+    K: Borrow<SumParticipantPublicKey>,
```
little-dude, line 66: I think it makes sense to have the trait bounds at the struct level, because it makes the compiler errors clearer when trying to create structs with inappropriate types.

rust/src/message/update.rs
```diff
@@ -95,9 +102,16 @@ pub struct UpdateMessage<K, C, S, M, D> {
...
+where
+    K: Borrow<UpdateParticipantPublicKey>,
+    C: Borrow<Certificate>,
+    S: Borrow<ParticipantTaskSignature>,
+    M: Borrow<Vec<u8>>,
+    D: Borrow<LocalSeedDict>,
+{
     /// Create an update message from its parts.
-    pub fn from(
+    pub fn new(
```
little-dude, line 114: An alternative would be to name this `from_parts`

little-dude:
I haven't updated the tests yet. Let's discuss the changes first.

rust/src/message/sum.rs
janpetschexain: i agree. it makes calls to the fields a bit more verbose, like `self.pk.borrow().as_ref()` instead of `self.pk.as_ref()`, but that's definitely worth it.

rust/src/message/sum.rs
```diff
@@ -61,16 +61,28 @@ impl<B: AsRef<[u8]> + AsMut<[u8]>> SumMessageBuffer<B> {
...
+impl<K, C, S, E> SumMessage<K, C, S, E>
+where
+    K: Borrow<SumParticipantPublicKey>,
+    C: Borrow<Certificate>,
+    S: Borrow<ParticipantTaskSignature>,
+    E: Borrow<SumParticipantEphemeralPublicKey>,
+{
     /// Create a sum message from its parts.
-    pub fn from(pk: K, certificate: C, sum_signature: S, ephm_pk: E) -> Self {
+    pub fn new(pk: K, certificate: C, sum_signature: S, ephm_pk: E) -> Self {
```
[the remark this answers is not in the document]
little-dude: Yeah that sounds good to me too

rust/src/message/sum.rs
```diff
@@ -134,7 +142,7 @@ impl
     fn deserialize(buffer: SumMessageBuffer<Vec<u8>>) -> Self {
         // safe unwraps: lengths of `buffer` slices are guaranteed by constants
         let pk = sign::PublicKey::from_slice(buffer.part_pk()).unwrap();
-        let certificate = buffer.certificate().to_vec();
+        let certificate = Certificate(vec![]);
```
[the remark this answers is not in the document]
little-dude: That's a mistake

PR #386, opened 2020-04-20
S3 store

Robert-Steiner:


rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+            s3_client: S3Client::new(region),
+            buckets: Buckets::new(),
+        }
+    }
+
+    /// Upload a [`Mask`]
+    pub async fn upload_mask(&self, mask_hash: &MaskHash, mask: &Mask) -> Result<(), StorageError> {
+        // mask_hash is stored in the mask_dict
+        let se_mask = bincode::serialize(mask).unwrap();
+        let key = hex::encode(&mask_hash);
```
little-dude, line 102: Question: must keys be ascii? Or is it just user friendliness?

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+        Self {
+            s3_client: S3Client::new(region),
+            buckets: Buckets::new(),
+        }
+    }
+
+    /// Upload a [`Mask`]
+    pub async fn upload_mask(&self, mask_hash: &MaskHash, mask: &Mask) -> Result<(), StorageError> {
+        // mask_hash is stored in the mask_dict
+        let se_mask = bincode::serialize(mask).unwrap();
```
little-dude, line 101: We should probably comment about why it's safe to unwrap here.

rust/src/storage/s3/store.rs
Robert-Steiner: No it doesn't have to. The only requirement is that the key is a string.

rust/src/storage/s3/store.rs
Robert-Steiner: Good catch! I think I should remove the unwrap here because I'm not sure if this is save.

little-dude: it's a bit complicated to understand what's susceptible to error out or not: https://github.com/xainag/xain-fl/pull/394/files#diff-3270ef1cc4780644df11318b22dd06ccR121

But I think in this case it's pretty safe to assume it works. We should have a few tests perhaps to confirm.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+                .list_objects(bucket, continuation_token)
+                .await
+                .map_err(|_| StorageError::S3Error)?;
+
+            if let Some(identifiers) = self.unpack_object_identifier(&list_obj_resp) {
+                self.delete_objects(bucket, identifiers)
+                    .await
+                    .map_err(|_| StorageError::S3Error)?;
+            } else {
+                break Ok(());
```
little-dude, line 366: nit: I know this is a matter of taste, I suggest we just `break;` and return `Ok(())` at the end of the function. This confused me for a second.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+
+    // Unpack the next_continuation_token of the ListObjectsV2Output response
+    fn unpack_next_continuation_token(
+        &self,
+        list_obj_resp: &ListObjectsV2Output,
+    ) -> Option<String> {
+        // https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingObjectKeysUsingJava.html
+        if let Some(is_truncated) = list_obj_resp.is_truncated {
+            if is_truncated {
+                list_obj_resp.next_continuation_token.clone()
```
little-dude, line 443: Is it possible that `is_truncated` is `true` but `next_continuation_token` is `false`? That won't affect the returned value here, but if that occurs, we might silently abort iterating through a truncated object in the caller.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+    fn new(s3_client: S3Client, bucket: String) -> Self {
+        Self {
+            s3_client: s3_client.clone(),
+            bucket: bucket.clone(),
+            list_object_ids_future: Some(ListObjectIdentifierFuture::new(s3_client, bucket, None)),
+        }
+    }
+
+    fn poll_object_identifiers(&mut self, cx: &mut Context) -> Poll<Option<Vec<ObjectIdentifier>>> {
+        if let Some(ref mut fut) = self.list_object_ids_future {
```
little-dude, line 528: nit: this is shifting everything to the right. One way to avoid that is:

```rust
let mut fut = if let Some(ref mut fut) = self.list_object_ids_future {
    fut
} else {
    return Poll::Ready(None);
}
// ...
```

But it is debatable whether this is better or not to be honest so I'll leave this as a mere suggestion.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+            match ready!(Pin::new(fut).poll(cx)) {
+                Ok((keys, next_continuation_token)) => {
+                    self.list_object_ids_future = match next_continuation_token {
```
little-dude, line 531: the `match` statement can be simplified a bit to avoid the `None => None` arm:

```rust
            match ready!(Pin::new(fut).poll(cx)) {
                Ok((keys, next_continuation_token)) => {
                    self.list_object_ids_future = next_continuation_token.map(|token| {
                        ListObjectIdentifierFuture::new(
                            self.s3_client.clone(),
                            self.bucket.clone(),
                            Some(token),
                        )
                    });
                    return Poll::Ready(keys);
                }
                Err(()) => return Poll::Ready(None),
            }
```

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+            let req = ListObjectsV2Request {
+                bucket: bucket.clone(),
+                continuation_token: continuation_token.clone(),
+                max_keys: Some(10),
+                ..Default::default()
+            };
+
+            let list_obj_resp = s3_client.list_objects_v2(req).await.map_err(|_| ())?;
+
+            let next_continuation_token = {
```
little-dude, line 481: It seems that we already have some helpers to ease the continuation token management (`unpack_next_continuation_token` for instance). Is something preventing us from using them here as well?

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+            // Minio is not limited to 1000
+            max_keys: Some(1000),
+            ..Default::default()
+        };
+
+        self.s3_client.list_objects_v2(req).await
+    }
+
+    // Unpack the next_continuation_token of the ListObjectsV2Output response
+    fn unpack_next_continuation_token(
```
little-dude, line 436: this doesn't need `&self` does it?

rust/src/storage/s3/store.rs
Robert-Steiner: yeah that's true. We should add some more specific error types to the StorageError enum later. For now I just used it as a placeholder.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,658 @@
...
+            None
+        }
+    }
+}
+
+// experimental stream implementation of the list object endpoint
+struct ListObjectIdentifierFuture(
+    Pin<
+        Box<
+            dyn Future<Output = Result<(Option<Vec<ObjectIdentifier>>, Option<String>), ()>> + Send,
```
little-dude, line 457: Some explanations about this type would be nice: what do `Option<Vec<ObjectIdentifier>>` and `Option<String>` represent? It might even be worth making this tuple a named struct (?).

rust/src/storage/s3/store.rs
Robert-Steiner: That should not be possible but we should test this😁.
In the [rusoto](https://docs.rs/rusoto_s3/0.43.0/rusoto_s3/struct.ListObjectsV2Output.html#structfield.next_continuation_token) documentation they wrote: 

> NextContinuationToken is sent when isTruncated is true, which means there are more keys in the bucket that can be listed. The next list requests to Amazon S3 can be continued with this NextContinuationToken. NextContinuationToken is obfuscated and is not a real key

rust/src/storage/s3/store.rs
Robert-Steiner: I had some issues putting it into the impl block of `ListObjectIdentifierFuture`. I can't remember the error. I will give it another try

rust/src/storage/s3/store.rs
Robert-Steiner: That is a good idea. I will change that

rust/src/storage/s3/store.rs
little-dude: Perhaps is worth checking that this holds true and just log an error if it doesn't then.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,734 @@
...
+use tokio::{io::AsyncReadExt, stream::Stream};
+
+#[derive(Debug, Error)]
+pub enum StorageError {
+    #[error("S3Error")]
+    S3Error,
+}
+
+#[derive(Clone)]
+struct Buckets(HashMap<&'static str, &'static str>);
```
little-dude, line 45: Since this is global and immutable, I think it could even be made `const`. I'm not sure what can be done these days with `const` (there has been a lot of work lately in the compiler in that area), so you might need either [`phf`](https://docs.rs/phf/0.8.0/phf/) or [`lazy_static`](https://docs.rs/lazy_static/1.4.0/lazy_static/index.html), not sure.

Maybe we can keep that for another PR though.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,734 @@
...
+        }
+    }
+
+    /// Upload a [`MaskedModel`].
+    pub async fn upload_masked_model(
+        &self,
+        key: &str,
+        masked_model: &MaskedModel,
+    ) -> Result<(), StorageError> {
+        // we don't need the keys of the masked models. we can retrieve all keys via `list objects`.
```
little-dude, line 126: > we don't need the keys of the masked models

Do you mean that we never need to retrieve a specific mask? That's true, but I think there's still some value to be able to identify a mask. Perhaps we could use the update participant public key digest for that. But anyway it doesn't change anything for this method.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,734 @@
...
+        Self {
+            s3_client: S3Client::new_with(dispatcher, credentials_provider, region),
+            buckets: Buckets::new(),
+        }
+    }
+
+    /// Upload a [`Mask`].
+    pub async fn upload_mask(&self, key: &str, mask: &Mask) -> Result<(), StorageError> {
+        match self
+            .upload(self.buckets.masks(), key, mask.serialize())
```
little-dude, line 112: The masks and models are pretty big, and the route I'm taking right now in the masking refactoring is to store them as `Vec<BigUint>` (for masked models) and `Vec<Ratio<BigInt>>` (for models), which will make them even bigger. So (de)serialization might take a non-negligeable amount of time, to the point that it might be worth considering using `block_on` for these operations.

Do you know if rusoto has a streaming API? Like can we upload a `tokio::Stream<Item=u8>`? If so I think it's worth investigating that area as well.

This can't be addressed now of course since we need benchmarks. Just something to keep in mind for the future.

rust/src/storage/s3/store.rs
```diff
@@ -0,0 +1,734 @@
...
+        // we don't need the keys of the masked models. we can retrieve all keys via `list objects`.
+        // maybe we need the number of objects in the bucket.
+        // the key is a random string, we could calculate the hash of a masked model and use this
+        // as a key but this would cost more time to calculate
+        match self
+            .upload(self.buckets.masked_models(), &key, masked_model.serialize())
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(_) => Err(StorageError::S3Error),
```
little-dude, line 135: It seems that we never use the returned value and that we always convert the error into a `StorageError`, so perhaps this could be done in the `upload` method itself? That way we could just do:

```rust
self.upload(self.buckets.masked_models(), &key, masked_model.serialize())
    .await?
```

rust/src/storage/s3/store.rs
Robert-Steiner: The value of the `body` in the `PutObjectRequest` is a [`StreamingBody`](https://docs.rs/rusoto_s3/0.43.0/rusoto_s3/struct.PutObjectRequest.html#structfield.body) which is an alias for [`ByteStream`](https://docs.rs/rusoto_core/0.43.0/rusoto_core/struct.ByteStream.html). I think this is what you mean right?

rust/src/storage/s3/store.rs
Robert-Steiner: Nice! I will take a look

rust/src/storage/s3/store.rs
little-dude: Yeah something like this. But `ByteStream` still requires the entire `Vec<u8>`. I was hoping we could serialize on the fly instead.

rsaffi:
Rebased against the updated `master`!

Robert-Steiner:
#544

PR #388, opened 2020-04-22
DO-105 :pencil: Remove telemetry flag from release image. Update compose

dkravetz:
This PR fixes `docker-compose-release.yaml` and removes the `telemetry` flag from the release builds

### DIFF
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -10,7 +10,7 @@ RUN bash ./install_rust.sh
 COPY rust/Cargo.lock .
 COPY rust/Cargo.toml .
 COPY rust/src src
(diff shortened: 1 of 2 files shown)


rsaffi:
If you guys don't mind @dkravetz and @little-dude I'd like to start using `hostname` on all services. It may sound silly at first, but when one has to debug/investigate something and effectively needs to jump into a shell in the container, it really helps having clear view of where you're typing stuff on. 
Been there enough times to see how useful it is. 

I sometimes set `container_name` as well, but that's not so useful, but more aesthetic.

WDYT?

little-dude:
That sounds good to me.

PR #391, opened 2020-04-23
DO-89 :wheel_of_dharma: Add k8s manifests

dkravetz:
This PR adds initial kubernetes manifests for both the aggregator and the coordinator. 

InfluxDB was installed beforehand by @rsaffi . SSL/TLS will be added as part of DO-96.

The current domains are [http://dev-aggregator.xain.io](http://dev-aggregator.xain.io) and [http://dev-coordinator.xain.io](http://dev-coordinator.xain.io).

This PR also updates the `docker-compose-release.yaml` to reflect the current status of the project, as well as removing Jaeger from the build features, as discussed with @rsaffi and @little-dude


k8s/aggregator/development/config-release-aggregator.toml
```diff
@@ -0,0 +1,14 @@
+[logging]
+filter = "info,xain_fl=trace"
+telemetry = { jaeger_endpoint = "jaeger:6831", service_name = "aggregator" }
```
rsaffi, line 3: We should be good to get rid of this, at least on `k8s`'s scope...

dkravetz: Good catch, I've already forgot about it. Done!

k8s/coordinator/development/config-release-coordinator.toml
```diff
@@ -0,0 +1,21 @@
+aggregator_url = "http://localhost:8082"
```
rsaffi, line 1: This won't work, because it means that the `coordinator` will inform participants that in order to reach the `aggregator`, they need to connect to `http://localhost:8082`.

rsaffi: I believe it should be `http://dev-aggregator.xain.io`, right?

dkravetz: Yes, but I'm also adding the port (80 in this case) :)

PR #392, opened 2020-04-24
wrap libsodium

little-dude:
We'd like to implement some traits on libsodium types, but this is
prevented by the orphan rule. To work around this, this commit
introduce newtypes around the libsodium crypto primitives that we use
in our codebase.


rust/src/crypto.rs
```diff
@@ -0,0 +1,265 @@
+use derive_more::{AsMut, AsRef, From};
+use sodiumoxide::crypto::{box_, sealedbox, sign};
+
+pub trait ByteObject: Sized {
+    /// Create a new object with all the bytes initialized to `0`.
+    fn null() -> Self;
```
janpetschexain, line 6: i would avoid the use of `null()` as a function name here, because i immediately associated it with `None`. maybe `init()` / `new()` / `zeroed()` suits better.

rust/src/crypto.rs
```diff
@@ -0,0 +1,265 @@
...
+    sealedbox::seal(m, recipient_pk.as_ref())
+}
+
+/// Decrypt the ciphertext `c` using the key pair `(pk, sk)` and
+/// returns the decrypted message.
+///
+/// If decryption fails `Err(())` is returned.
+pub fn decrypt(c: &[u8], pk: &PublicKey, sk: &SecretKey) -> Result<Vec<u8>, ()> {
+    sealedbox::open(c, pk.as_ref(), sk.as_ref())
+}
```
janpetschexain, line 129: we could tie encryption and decryption to the respective keys to make these calls similar to the signing. sth along the lines of
```
impl PublicKey {
    pub fn encrypt(&self, m: &[u8]) -> Vec<u8> {
        sealedbox::seal(m, self.as_ref())
    }
}

impl SecretKey {
    pub fn decrypt(&self, c: &[u8], pk: &PublicKey) -> Result<Vec<u8>, ()> {
        sealedbox::open(c, pk.as_ref(), self.as_ref())
    }
}
```
the second one could even have a shorter function signature like
```
pub fn decrypt(&self, c: &[u8]) -> Result<Vec<u8>, ()> {
    sealedbox::open(c, &self.as_ref().public_key(), self.as_ref())
}
```
but be aware that the call to `public_key()` on a secret key actually recomputes the public key from the secret key which involves a multiplication on the elliptic curve. this can become a limitation if we need to do this very often.

rust/src/crypto.rs
little-dude: I like `zeroed()`, I'll change that

rust/src/crypto.rs
little-dude: I initially did it that way but it was unclear whether I should implement `decrypt` on the public or the secret key. It felt quite arbitrary to implement it on the secret key. Wdyt?

little-dude:
re: naming

I hesitate, but I didn't really like abbreviating `Encr`, and I felt like `Encryption` was a bit long so ultimately I decided not to choose and go without a prefix :p

Now that I think about it more, I think I prefer the full `Encryption` version. What do you think about:

```
PublicEncryptionKey
SecretEncryptionKey
PublicSigningKey
SecretSigningKey
```

rust/src/crypto.rs
janpetschexain: it's not arbitrary at all. encryption is always done with the public key. decryption is always done with the secret key.
actually i'm not quite sure why the public key is even provided to the decrypt function in libsodium. maybe some internal consistency checks or just a relic. but regarding the mathematics, it is not needed there.

janpetschexain:
i would go with (3) or (4), slightly favoring (3).

finiteprods:
i actually quite like (2) as `Encr` and `Sign` are both 4 chars, so line up nicely (as do `Public` and `Secret`). i'll throw out another suggestion, which is compromise of (2) and (3)

```
# suggestion (5)
PublicEncryptKey, SecretEncryptKey
PublicSigningKey, SecretSigningKey
```
i don't mind whether it's `Public`/`Secret` prefix (as above) or `Encrypt`/`Signing`

janpetschexain:
> i actually quite like (2) as `Encr` and `Sign` are both 4 chars, so line up nicely (as do `Public` and `Secret`). i'll throw out another suggestion, which is compromise of (2) and (3)
> 
> ```
> # suggestion (5)
> PublicEncryptKey, SecretEncryptKey
> PublicSigningKey, SecretSigningKey
> ```
> 
> i don't mind whether it's `Public`/`Secret` prefix (as above) or `Encrypt`/`Signing`

that's another good suggestion. having equally long names was my original intention behind `Encr`.

we're tackling one of the hardest problems in computer science, no easy solution :D
https://hackernoon.com/the-art-of-naming-variables-52f44de00aad

Robert-Steiner:
I like (2) and (5)

little-dude:
rebased. I like (5) to and it seems to make everyone happy so I'm gonna go with that. I don't understand what having the same length for the type names is important though :smile:

janpetschexain:
> rebased. I like (5) to and it seems to make everyone happy so I'm gonna go with that. I don't understand what having the same length for the type names is important though 😄

the answer to that is OCD ^^

little-dude:
@janpetschexain I addressed all your comments, and updated the tests. This is ready for another round of review.

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Generate a new random key pair
+pub fn generate_encrypt_key_pair() -> (PublicEncryptKey, SecretEncryptKey) {
+    let (pk, sk) = box_::gen_keypair();
+    (PublicEncryptKey(pk), SecretEncryptKey(sk))
+}
+
+/// Deterministically derive a new key pair from the given seed
+pub fn derive_key_pair_from_seed(seed: &KeySeed) -> (PublicEncryptKey, SecretEncryptKey) {
```
janpetschexain, line 34: i would suggest a naming consistent to the key types:
`derive_encrypt_key_pair_from_seed()`

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    fn as_slice(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+/// A seed that can be used for signing key pair generation. When
+/// `KeySeed` goes out of scope, its contents will be zeroed out.
+#[derive(AsRef, AsMut, From, Serialize, Deserialize, Eq, PartialEq, Clone)]
+pub struct SignKeySeed(sign::Seed);
```
janpetschexain, line 61: i would suggest a naming consistent to the key types:
`SigningKeySeed`

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Generate a new random signing key pair
+pub fn generate_sign_key_pair() -> (PublicSigningKey, SecretSigningKey) {
```
janpetschexain, line 64: i would suggest a naming consistent to the key types:
`generate_signing_key_pair()`

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+    let (pk, sk) = sign::gen_keypair();
+    (PublicSigningKey(pk), SecretSigningKey(sk))
+}
+
+/// Deterministically derive a new signing key pair from the given
+/// seed
+pub fn derive_sign_key_pair_from_seed(seed: &SignKeySeed) -> (PublicSigningKey, SecretSigningKey) {
```
janpetschexain, line 71: i would suggest a naming consistent to the key types:
`derive_signing_key_pair_from_seed()`

little-dude:
I think I'd prefer to remove the `use crypto::*` and move the type aliases to `lib.rs` actually. I'll do that tonight.

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+/// Deterministically derive a new key pair from the given seed
+pub fn derive_key_pair_from_seed(seed: &KeySeed) -> (PublicEncryptKey, SecretEncryptKey) {
+    let (pk, sk) = box_::keypair_from_seed(seed.as_ref());
+    (PublicEncryptKey(pk), SecretEncryptKey(sk))
+}
+
+/// A seed that can be used for key pair generation. When `KeySeed`
+/// goes out of scope, its contents will be zeroed out.
+#[derive(AsRef, AsMut, From, Serialize, Deserialize, Eq, PartialEq, Clone)]
+pub struct KeySeed(box_::Seed);
```
janpetschexain, line 42: i would suggest a naming consistent to the key types:
`EncryptKeySeed`

rust/src/crypto.rs
janpetschexain, line 34: an optional suggestion: this could become a function of `EncryptKeySeed` like the following
```
impl EncryptKeySeed {
    pub fn derive_encrypt_key_pair(&self) -> (PublicEncryptKey, SecretEncryptKey) {
        let (pk, sk) = box_::keypair_from_seed(self.as_ref());
        (PublicEncryptKey(pk), SecretEncryptKey(sk))
    }
}
```
and similarly for signing seeds

little-dude: good idea

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+pub type SumParticipantEphemeralSecretKey = SecretEncryptKey;
+
+/// A public signature key that identifies an update participant.
+pub type UpdateParticipantPublicKey = ParticipantPublicKey;
+
+/// A secret signature key that belongs to the public key of an update
+/// participant.
+pub type UpdateParticipantSecretKey = ParticipantSecretKey;
+
+/// A signature to prove a participant's elligibility for a task.
```
finiteprods, line 313:
```suggestion
/// A signature to prove a participant's eligibility for a task.
```

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+/// participant.
+pub type SumParticipantSecretKey = ParticipantSecretKey;
+
+/// A public encryption key generated by a sum participant. It is used
+/// by the update participants to encrypt their masking seed for each
+/// sum participant.
+pub type SumParticipantEphemeralPublicKey = PublicEncryptKey;
+
+/// A secret signature key that belongs to the public key of a sum
+/// participant.
```
finiteprods, line 303: check this docstring

rust/src/crypto.rs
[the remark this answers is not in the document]
little-dude: I picked "derive" here because I wanted to highlight that unlike `generate_key_pair` above, this function was deterministic: given the same input it will always output the same thing. I don't know if it really matters though.

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+    Deserialize,
+    Hash,
+    Eq,
+    Ord,
+    PartialEq,
+    Copy,
+    Clone,
+    PartialOrd,
+    Debug,
+)]
```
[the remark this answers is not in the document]
little-dude: yeah, but that's probably configurable.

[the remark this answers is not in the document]
little-dude: If I remove the last one, `rustfmt` puts it back on a single line.

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    /// Compute the corresponding public key for this secret key
+    pub fn public_key(&self) -> PublicEncryptKey {
+        PublicEncryptKey(self.0.public_key())
+    }
+}
+
+/// Secret key for asymmetric authenticated encryption
+#[derive(AsRef, AsMut, From, Serialize, Deserialize, Eq, PartialEq, Clone, Debug)]
+pub struct SecretEncryptKey(box_::SecretKey);
```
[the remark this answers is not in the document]
little-dude: I'm not sure what you're referring to here.

rust/src/crypto.rs
little-dude: Oops, indeed.

rust/src/crypto.rs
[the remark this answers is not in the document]
finiteprods: just saying that at the moment, `impl SecretEncryptKey {...}` comes before this line, and the other way round would be more conventional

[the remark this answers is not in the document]
little-dude: Ah sure I'll move it.

rust/src/crypto.rs
```diff
@@ -0,0 +1,314 @@
...
+        sealedbox::seal(m, self.as_ref())
+    }
+}
+
+impl SecretEncryptKey {
+    /// Decrypt the ciphertext `c` using this secret key and the
+    /// associated public key, and return the decrypted message.
+    ///
+    /// If decryption fails `Err(())` is returned.
+    pub fn decrypt(&self, c: &[u8], pk: &PublicEncryptKey) -> Result<Vec<u8>, ()> {
```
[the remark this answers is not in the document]
janpetschexain: when using the `public_key()` method then we don't have to pass it. but calling `public_key()` on a secret key recomputes the public key from the secret key which involves a multiplication on the elliptic curve. this can become quite costly if it is done very often. the alternative is to pass the precomputed public key by reference as an extra argument.

rust/src/crypto.rs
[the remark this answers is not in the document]
janpetschexain: i'm afraid that changing rustfmt would affect all formating of function-like argument lists (see `attr_fn_like_width ` in https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#attr_fn_like_width). would it be valid rust syntax to add a manual line break here like the following?
```
#[derive(AsRef, AsMut, From, Serialize, Deserialize, Hash, Eq)]
#[derive(Ord, PartialEq, Copy, Clone, PartialOrd, Debug)]
struct ...
```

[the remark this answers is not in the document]
little-dude: That would work but this is even weirder imo

little-dude:
I've addressed all the comments except https://github.com/xainag/xain-fl/pull/392#discussion_r414627968

I also:

- moved the type aliases back to `lib.rs`
- split the `crypto` module into two files: `crypt/sign.rs` and `crypto/encrypt.rs`
- removed the `use crypto::*` that I had put in all the modules. That created tons of errors, and I went through all of them, replacing the generic `crypto::*` types by the appropriate alias. In the tests there were places where it was unclear what was required though, especially here: https://github.com/xainag/xain-fl/pull/392/commits/3cd671b2bcf077a792e2dccc24d880477bdd9450#diff-7047ef3ec6efbf0f4f070fc70c1eb404R300

rust/src/crypto.rs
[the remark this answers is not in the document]
janpetschexain: on one hand the deterministic nature of this function is already implicitly given by `..._from_seed()`, on the other hand it probably doesn't do any harm to have a `derive_...` as an explicit action. no strong preference for either one.

rust/src/participant.rs
```diff
@@ -296,7 +294,10 @@ mod tests {
                 .iter()
                 .map(|seed| {
                     (
-                        sign::PublicKey::from_slice(&randombytes(32)).unwrap(),
+                        // Or `UpdateParticipantPublicKey` ??? I don't
+                        // understand if this is supposed to be a
+                        // local seed dict or the global one.
+                        SumParticipantPublicKey::from_slice(&randombytes(32)).unwrap(),
```
janpetschexain, line 300: `UpdateParticipantPublicKey` instead of `SumParticipantPublicKey` is correct.

this is actually neither a `SeedDict` nor a `LocalSeedDict` but the inner dict of a `SeedDict`. basically the inner `collect()` creates a bunch of `HashMap<UpdateParticipantPublicKey, EncrMaskSeed>`, then the outer `collect()` creates the `SeedDict`. sorry for this monstrous double-collect :/

rust/src/crypto/encrypt.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Compute the corresponding public key for this secret key
+    pub fn public_key(&self) -> PublicEncryptKey {
+        PublicEncryptKey(self.0.public_key())
+    }
+}
+
+/// Secret key for asymmetric authenticated encryption
+#[derive(AsRef, AsMut, From, Serialize, Deserialize, Eq, PartialEq, Clone, Debug)]
+pub struct SecretEncryptKey(box_::SecretKey);
```
janpetschexain, line 75: it's probably better to have the definition of `SecretEncryptKey` before the impl in line 58

rust/src/crypto.rs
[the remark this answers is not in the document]
little-dude: Note that the `_from_seed()` suffix is gone now that it's a method of `EncryptKeySeed`.

@finiteprods what do you think of the arguments above?

little-dude:
> also sorry for the sometimes unclear types in the tests, i often took the lazy way and just mocked keys where ever they are just collected but not used, this makes type inference not easy.

Ah no problem at all! And thank you for the thorough reviews, this PR is definitely looking much better than the initial version.

I've addressed your last round of comment, so I'll wait for @finiteprods feedback and then squash everything.

PR #396, opened 2020-04-30
PB-610 masking

janpetschexain:
**References**
- [PB-610](https://xainag.atlassian.net/browse/PB-610)

**Summary**
- add a `Model<N>` type which represents its parameters internally as a `Vec<N>` of numerical values. currently, models for `f32`, `f64`, `i32` and `i64` are implemented and this is easily extendable. models can be masked wrt a `MaskConfig`.
- add a `MaskIntegers` trait which represents masked numbers. they can be aggregated and serialized. this is implemented for `MaskedModel` as well as `Mask`.
- implement a `MaskedModel` type which represents its parameters as a `Vec<BigUint>` and can be unmasked by providing a `Mask`. unmasking is currently implemented into `Model<N>`s of `f32`, `f64, `i32` and `i64` and can easily be extended.
- implement a `Mask` type which has the same representation as `MaskModel` and can be used to unmask a `MaskedModel`.
- implement `crypto::ByteObject` for `MaskSeed` and `EncryptedMaskSeed` and their respective encryption/decryption methods.
- implement a `MaskConfig` which holds the information needed to shift a model parameter from the primitive data type into the respective finite group and reverse.
- add a catalog of available `MaskConfigs` depending on group type, data type, bound type and model type.
- adjust the participant's messages accordingly.
- restructure the coordinator via traits to use the `Model<N>`, `MaskedModel` and `Mask` and change the corresponding functions.
- implement `crypto::ByteObject` for `Certificate` and `RoundSeed`
- add tests accordingly
- add macros to derive field accessors for traits and structs

**Notes**
- the available mask configurations are currently represented as a catalog. it may be part of a future optimization to derive the mask configuration programmatically, but right now this is lacking some utility, namely to calculate base-2 logarithm on big integers for power2 group orders and to check big integers for being a prime number for prime group orders.
- the prime check is also a time constraint if done repeatedly. a naive implementation to check a number `n` will have time complexity `O(n^1/2)` and even optimized probabilistic tests (with `k` rounds) still are `Õ(k log(n)^2)`. since we have to deal with arbitrary large integers this might be an issue and we of course need an implementation of efficient checks in the first place.
- the coordinator doesn't make the aggregated model available yet. also the service defaults to a `Coordinator<f32>` currently. both have to be tackled soon.


rust/src/model.rs
```diff
@@ -0,0 +1,160 @@
...
+        MaskIntegers,
+        MaskedModel,
+    },
+    utils::generate_integer,
+    PetError,
+};
+
+#[derive(Clone, Debug, PartialEq)]
+/// A model with parameters represented as a vector of numbers.
+pub enum Model {
```
little-dude, line 20: Why using an enum here? My first impression is that generics would fit nicely:

```rust
struct Model<T>(Vec<T>)
```

The only method that must be implemented manually for each type afaict is `xxx_as_ratio` so we could have:

```
trait AsRatios {
    fn as_ratios(&self) -> Option<Vec<Ratio<BigInt>>>;
}

impl AsRatios for Model<f64> { ... }
impl AsRatios for Model<f32> { ... }
impl AsRatios for Model<i32> { ... }
// etc.
```

Then:

```impl<T> Model<T>
where
    Model<T>: AsRatios,
{
    fn mask(&self, ...,) {
	    // less repetition now
    }
}
```

I may be missing something though.

janpetschexain: i started with generics here, but failed. the issue is that the methods from the `num` crate impose some specific trait bounds on their method arguments. for example methods on floats are bounded by `Float`/`FloatCore` whereas methods on integers are bounded by `Integer`/`PrimInt` and those are disjunct. any generic type `struct Model<T>(Vec<T>)` must impose either one of the trait bounds on `T` at the same time. my workaround was to implement it as an enum without generics.

little-dude: Ah then maybe what we need is different impls for floats and ints. Basically `Model<T: Float>` and `Model<T: Integer>` ? I'd need to take a closer look though.

janpetschexain: i refactored it using traits now. i was basically blocked by having an unnecessary extra layer of generics in my mental model, like `primitive_type: FloatCore/PrimInt: Model<N>`. by resolving that with the new traits and removing most of the occurrences of `FloatCore`/`PrimInt` this works now. thanks for pointing me in the right direction!

rust/src/certificate.rs
```diff
@@ -1,49 +1,54 @@
...
 }
 
-impl AsRef<[u8]> for Certificate {
-    /// Get a reference to the certificate.
-    fn as_ref(&self) -> &[u8] {
-        self.0.as_slice()
+#[allow(clippy::len_without_is_empty)]
+impl Certificate {
+    /// Get the number of bytes of a certificate.
+    pub const BYTES: usize = 32;
```
little-dude, line 29: fwiw I named that `LENGTH` in https://github.com/xainag/xain-fl/pull/398. I think it's a bit clearer. But does it make sense to have this for certificates? Their size is not fixed.

rust/src/lib.rs
```diff
@@ -76,18 +85,13 @@ type SumDict = HashMap<SumParticipantPublicKey, SumParticipantEphemeralPublicKey
 
 /// Local seed dictionaries are sent by update participants. They contain the participant's masking
 /// seed, encrypted with the ephemeral public key of each sum participant.
-type LocalSeedDict = HashMap<SumParticipantPublicKey, EncrMaskSeed>;
+type LocalSeedDict = HashMap<SumParticipantPublicKey, EncryptedMaskSeed>;
```
little-dude, line 88: :+1:

rust/src/model.rs
```diff
@@ -0,0 +1,569 @@
...
+/// Masking of models.
+pub trait MaskModels<N> {
+    /// Get a reference to the weights.
+    fn weights(&self) -> &Vec<N>;
+
+    /// Cast the weights as ratios. Must handle non-finite weights.
+    fn as_ratios(&self) -> Vec<Ratio<BigInt>>;
+
+    /// Mask the model wrt the mask configuration. Enforces bounds on the scalar and weights.
+    fn mask(&self, scalar: f64, config: &MaskConfig) -> (MaskSeed, MaskedModel) {
```
little-dude, line 22: Could the seed be an argument to the function? It would make testing this function easier if we can control the seed.

rust/src/model.rs
```diff
@@ -0,0 +1,569 @@
...
+                (shifted + generate_integer(&mut prng, config.order())) % config.order()
+            })
+            .collect();
+        // safe unwrap: masked weights are guaranteed to conform to the mask configuration
+        let masked_model = MaskedModel::from_parts(masked_weights, config.clone()).unwrap();
+        (mask_seed, masked_model)
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
```
little-dude, line 47: We usually derive `Eq` along with `PartialEq`. As long as the numbers are not float that should work (equality is not reflexive for floats).

rust/src/participant.rs
```diff
@@ -113,7 +115,17 @@ impl Participant {
 
     /// Compose an update message.
     pub fn compose_update_message(&self, pk: &CoordinatorPublicKey, sum_dict: &SumDict) -> Vec<u8> {
-        let (mask_seed, masked_model) = Self::mask_model();
+        let model = Model::try_from(vec![0_f32, 0.5, -0.5]).unwrap(); // dummy
```
little-dude, line 118: Is this for testing or should we hardcode this for now?

rust/src/participant.rs
```diff
@@ -221,29 +248,29 @@ mod tests {
     #[test]
     fn test_check_task() {
         let mut part = Participant::new().unwrap();
-        let elligible_signature = Signature::from_slice_unchecked(&[
-            229, 191, 74, 163, 113, 6, 242, 191, 255, 225, 40, 89, 210, 94, 25, 50, 44, 129, 155,
-            241, 99, 64, 25, 212, 157, 235, 102, 95, 115, 18, 158, 115, 253, 136, 178, 223, 4, 47,
-            54, 162, 236, 78, 126, 114, 205, 217, 250, 163, 223, 149, 31, 65, 179, 179, 60, 64, 34,
-            1, 78, 245, 1, 50, 165, 47,
+        let eligible_signature = Signature::from_slice_unchecked(&[
```
little-dude, line 251: Why was this changed? Also, would it be possible to add a comment to explain how these values were found?

rust/src/utils.rs
```diff
@@ -15,42 +18,100 @@ pub fn is_eligible(signature: &ParticipantTaskSignature, threshold: f64) -> bool
     } else if threshold > 1_f64 {
         true
     } else {
```
little-dude, line 20: nit: I would remove one level of indentation here:

```rust
    if threshold < 0_f64 {
        return false;
    } else if threshold > 1_f64 {
        return true;
    }
    // ...
```

rust/src/utils.rs
```diff
@@ -15,42 +18,100 @@ pub fn is_eligible(signature: &ParticipantTaskSignature, threshold: f64) -> bool
...
+            .to_bigint()
+            .unwrap();
+        // safe unwrap: `threshold` is guaranteed to be finite
+        Ratio::new(numer, denom) <= Ratio::from_float(threshold).unwrap()
     }
 }
 
+/// Generate a secure pseudo-random integer. Draws from a uniform distribution over the integers
+/// between zero (included) and `max_int` (excluded).
+pub fn generate_integer(prng: &mut ChaCha20Rng, max_int: &BigUint) -> BigUint {
```
little-dude, line 35: Maybe this could go in the `crypto` module?

rust/src/utils.rs
```diff
@@ -15,42 +18,100 @@ pub fn is_eligible(signature: &ParticipantTaskSignature, threshold: f64) -> bool
...
         ]);
         assert!(!is_eligible(&sig, 0.5_f64));
     }
+
+    #[test]
+    fn test_generate_integer() {
+        let mut prng = ChaCha20Rng::from_seed([0_u8; 32]);
+        let max_int = BigUint::from(u128::max_value()).pow(2_usize);
+        assert_eq!(
+            generate_integer(&mut prng, &max_int).to_bytes_le(),
```
little-dude, line 82: I'm not sure it's really useful to test the `to_bytes_le()` part, because it makes the tests quite verbose. We can probably assume it works correctly since it comes from `num`.

rust/src/model.rs
little-dude, line 22: Could we document how the mask is computed?

rust/src/model.rs
```diff
@@ -0,0 +1,569 @@
...
+    }
+}
+
+impl MaskModels<f32> for Model<f32> {
+    /// Get a reference to the weights.
+    fn weights(&self) -> &Vec<f32> {
+        &self.weights
+    }
+
+    /// Cast the weights as ratios. Positve/negative infinity is mapped to max/min and NaN to zero.
```
little-dude, line 99: s/Positve/Positive

rust/src/model.rs
```diff
@@ -0,0 +1,569 @@
...
+                Ratio::from_integer(BigInt::from(-1_i64)),
+                Ratio::<BigInt>::zero(),
+                Ratio::from_integer(BigInt::from(1_i64)),
+            ],
+        );
+    }
+
+    // for float types the error depends on the order of magnitude of the weights => raise the
+    // tolerance or bound the random weights if this test fails to often
+    macro_rules! test_masking {
```
little-dude, line 344: Could we document what this test does? It seems important but I don't understand what it does.

rust/src/model.rs
```diff
@@ -0,0 +1,569 @@
...
+                Ratio::<BigInt>::from_float(f32::MAX).unwrap(),
+            ],
+        );
+    }
+
+    #[test]
+    fn test_model_f64() {
+        // try from
+        let weights = vec![-1_f64, 0_f64, 1_f64];
+        let neg_inf_weights = vec![f64::NEG_INFINITY, 0_f64, 1_f64];
```
little-dude, line 250: I think it would make sense to split these cases in different test functions, it wouldn't add much code, and would make the tests simpler to read.

rust/src/model.rs
```diff
@@ -0,0 +1,569 @@
...
+        },
+        seed::MaskSeed,
+        MaskIntegers,
+    };
+
+    #[test]
+    fn test_model_f32() {
+        // try from
+        let weights = vec![-1_f32, 0_f32, 1_f32];
+        let neg_inf_weights = vec![f32::NEG_INFINITY, 0_f32, 1_f32];
```
little-dude, line 190: Same comment than below: could we have one test for each case instead?

rust/src/certificate.rs
janpetschexain: i've seen the changes for `LENGTH`, makes sense to adopt this. regarding this constant for certificates, it's just a dummy for now and will probably change once we implement/use real certificates.

rust/src/model.rs
janpetschexain: i didn't put the seed as an argument to the method to avoid cases of "forgotten" seeding or seeding with non-random values. but i agree that it would help testing. what do you think would have more benefits?

rust/src/model.rs
janpetschexain: i'll add that.

rust/src/participant.rs
janpetschexain: it's mainly for integrating the masking, since it needs a (dummy) model to operate on. i would leave it in for now, i guess it's going to be replaced soon anyways when Kwok finishes his work on the participant.

rust/src/participant.rs
janpetschexain: i changed it here to make it consistent with the changed tests in `utils.rs`, other than that there was no reason.
the values were found by generating a random key pair, calculating the corresponding signature and checking by hand if the eligibility criterium was fulfilled.

rust/src/utils.rs
janpetschexain: i'll change that.

rust/src/participant.rs
little-dude: sounds good

rust/src/utils.rs
janpetschexain: i'll move it.

rust/src/utils.rs
janpetschexain: i'll change that to a string-radix representation for the `BigUint`.

rust/src/model.rs
janpetschexain: definitely. i would like to link the document Michael once drafted for that, but this is gone from confluence, not sure why. for now i will write some lines in the docstring. just didn't have time yet.

rust/src/utils.rs
janpetschexain: i'll also move the `is_eligible()` to `crypto::sign`.

rust/src/model.rs
janpetschexain: fixed.

rust/src/model.rs
janpetschexain: i'll do that. i still want to increase test coverage and will include it there.

rust/src/model.rs
janpetschexain: i'll split it.

janpetschexain:
> Looks good to me overall. But are seemingly unrelated changes to the `message` module which will conflict with #398. Are these necessary for this PR?

there are some necessary and some invertible changes:
- deserialization changed because deserialization of certificates can fail now. this will be handled by your changes as well, so it can be ignored, but should be kept in mind.
- apart from that deserialization of the local seed dictionary was refactored.
- the macro usage for the struct field accessors can be removed/ignored since it is handled differently by your changes.
- the tests were updated to use the masking, this applies to the `update` and `sum2` module.
- the tests were updated to adopt the new `ByteObject` trait. this can be ignored for now, but we should use it in the refactored module.

rust/src/model.rs
janpetschexain: Michael provided the document and i added it to confluence: https://xainag.atlassian.net/wiki/spaces/FP/pages/542408769/Masking

little-dude:
I think this is ready to be merged.

To make it easier for me to rebase my PR after this is merged here is what I suggest:

> deserialization changed because deserialization of certificates can fail now. this will be handled by your changes as well, so it can be ignored, but should be kept in mind.

Ignore

> deserialization of the local seed dictionary was refactored.

I did it as well: https://github.com/xainag/xain-fl/pull/398/files#diff-af6390119e17f2da639293e1e7a007c1R306

Need to check in details how our implementations differ to see which one to keep


> the macro usage for the struct field accessors can be removed/ignored since it is handled differently by your changes.

Ignore

> the tests were updated to use the masking, this applies to the update and sum2 module.

Keep

> the tests were updated to adopt the new ByteObject trait. this can be ignored for now, but we should use it in the refactored module.

Agree. I also wanted to do that.


What do you think?

janpetschexain:
> What do you think?

sounds good to me.

janpetschexain:
> > deserialization of the local seed dictionary was refactored.
> 
> I did it as well: https://github.com/xainag/xain-fl/pull/398/files#diff-af6390119e17f2da639293e1e7a007c1R306
> 
> Need to check in details how our implementations differ to see which one to keep

can be ignored, see https://github.com/xainag/xain-fl/pull/398#discussion_r420667538

PR #401, opened 2020-05-07
PB-734 masking tests

janpetschexain:
**References**
- [PB-734](https://xainag.atlassian.net/browse/PB-734)

**Summary**
- update tests for masking
- add tests for aggregation
- add tests for masking combined with aggregation
- fix `I32` and `I64` data types bug: missing precision for integer models in the latter test case
- fix `Integer` group type bug: the order was underestimated by `1`
- use `order - 1`, which is the max possible number under a mask config, instead of `order` for `bytes_per_digit()`: this saves one byte per serialized model weight in case of those `Power2` group types which happen to be `2**(8*n), n in N`


janpetschexain:
> Could we hold this off until I make the refactoring proposal for the masking code? I've a lot of local changes so I'd like to avoid having to rebase, and I think this would be easier to backport than me rebasing.

sure

rust/src/mask/masking.rs
```diff
@@ -614,97 +607,93 @@ mod tests {
...
-                        }
-                    )
-                    .unwrap();
+                    let tolerance = Ratio::from_integer(BigInt::from($count as usize))
+                        / Ratio::from_integer(config.exp_shift());
                     assert!(
                         averaged_model.iter()
                             .zip(unmasked_model.iter())
                             .all(|(averaged_weight, unmasked_weight)| {
+                                println!("{}, {}", averaged_weight, unmasked_weight);
```
little-dude, line 616: I think you forgot this

janpetschexain: well, indeed

PR #403, opened 2020-05-12
PB-718 Update CI

Robert-Steiner:
**References**

* [PB-718](https://xainag.atlassian.net/browse/PB-718)
* [PB-717](https://xainag.atlassian.net/browse/PB-717)

**Summary**

1. Updated docker-compose file:
   - Removed all existing services
   - Added minio service (I took the same setup we had in the old python repo)
2. Updated Rust CI:
	 - Added a smoke test job which starts all external dependencies (in the moment it is only minio) via docker-compose.
	 - The smoke test job runs tests that are marked with the `[#ignore]` annotation only.
	 - The `[#ignore]` annotation is used to distinguish between simple unit tests and tests that depend on an external service.
	    	- It is the best solution I have been able to find so far. We can also remove the `[#ignore]`  annotation and run them together in one job but I think it will make development easier if we can run just unit test without having to run any external services via docker-compose.
	- Add caching
		- I noticied that the ci is fast with caching enabled but I havn't tested it much.
		- I'm fine removing the cache if it creates to much visual noice.

**Notes:** 
(just some findings I want to share)

### [**`jobs.id.container`**](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer)

`Steps` can run in a docker container.

An example:

Instead of installing rust nightly via a `step` like:

```
      - name: install nightly
        uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: nightly
          components: clippy

```

we can specify that we want to run the actual step (e.g. `cargo test`) in the `rustlang/rust:nightly` container.

```diff
 test:
    name: rust-test
    runs-on: ubuntu-latest
+     container: rustlang/rust:nightly
    steps:
      - uses: actions/checkout@master
-     - name: install nightly
-       uses: actions-rs/toolchain@v1
-       with:
-         profile: minimal
-         toolchain: nightly
-         components: clippy

      - name: run unit tests
        working-directory: ./rust
        env:
          RUSTFLAGS: "-D warnings"
        run: cargo test

```

The reason why I decided against this feature is that the pulling of the docker image takes longer than installing the toolchain.

### [**`jobs.id.services`**](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices)

> Used to host service containers for a job in a workflow. Service  containers are useful for creating databases or cache services like  Redis. The runner  automatically creates a Docker network and manages  the life cycle of the service containers.

Github workflow has a feature to specify external docker containers as dependencies directly in the workflow file, so that we don't need to call docker-compose manually. An example for minio:

```yml
  test:
    name: rust-test
    runs-on: ubuntu-latest
    container: rustlang/rust:nightly
    # Service containers to run with `container-job`
    services:
      # Label used to access the service container
      minio:
        # Docker Hub image
        image: minio/minio
        env:
          MINIO_ACCESS_KEY: minio
          MINIO_SECRET_KEY: minio123
        ports:
          - 9000:9000
        options: --health-cmd "curl -f http://localhost:9000/minio/health/live" --health-interval 10s --health-timeout 5s --health-retries 5
```

I really like that feature but unfortunately it does not work with minio. Minio requires the command `server [volume]` to  run in server mode. Without that command the minio container will immediately quit. [The Github workflow does not yet support the specification of a command argument.]( https://github.community/t5/GitHub-Actions/How-do-I-properly-override-a-service-entrypoint/td-p/44129)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -127,7 +127,6 @@ jobs:
         with:
           profile: minimal
           toolchain: nightly
-          components: rustfmt, clippy
 
       - name: Cache cargo registry
         uses: actions/cache@v1
@@ -173,7 +172,6 @@ jobs:
         with:
           profile: minimal
           toolchain: nightly
-          components: rustfmt, clippy
 
       - name: Cache cargo registry
         uses: actions/cache@v1
@@ -204,10 +202,3 @@ jobs:
       - name: Stop docker-compose
         working-directory: ./docker
         run: docker-compose stop
-
-  test-container:
-    name: rustfmt
-    runs-on: ubuntu-latest
-    container: rustlang/rust:nightly
-    steps:
-      - uses: actions/checkout@master
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -3,6 +3,39 @@ on: [push]
 name: Rust-CI
 
 jobs:
+  rustfmt:
+    name: rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+
+      - name: install nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: rustfmt
+
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo index
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/git
+          key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo build
+        uses: actions/cache@v1
+        with:
+          path: rust/target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: cargo fmt
+        working-directory: ./rust
+        run: cargo fmt --all -- --check
+
   check:
     name: rust-build
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -38,62 +87,120 @@ jobs:
...
 
       - name: install nightly
         uses: actions-rs/toolchain@v1
         with:
-            profile: minimal
-            toolchain: nightly
-            components: rustfmt
+          profile: minimal
+          toolchain: nightly
+          components: rustfmt, clippy
```
little-dude, line 176: I don't think we need these two components here do we?

.github/workflows/rust.yml
```diff
@@ -3,6 +3,39 @@ on: [push]
 name: Rust-CI
 
 jobs:
+  rustfmt:
+    name: rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+
+      - name: install nightly
```
little-dude, line 12: As noted in the PR description, this could be replaced by a `runs-on:` item that specify a rust-nightly container no?

.github/workflows/rust.yml
```diff
@@ -16,17 +49,33 @@ jobs:
       - name: install nightly
```
little-dude, line 49: Same as above: do we still need this step?

.github/workflows/rust.yml
Robert-Steiner: Right, it can be replaced with the item `container: rustlang/rust:nightly`.
The `runs-on:` item only describes the the type of the VM. It is a bit confusing. If you only specify the `runs-on:` item, the steps will run on the VM. If you additionally specify a `container` item, the steps will run in that docker container.
The `container` feature makes the workflow easier to read but it is also a bit slower 10-20 sec. Shall we use it anyway?

Robert-Steiner:
> Very nice. Definitely an improvement. Do you know if there's a way to group a sequence of steps? It would be really nice to be able to re-use the steps for caching instead of having to duplicate them:
> 
> ```yaml
>       - name: Cache cargo registry
>         uses: actions/cache@v1
>         with:
>           path: ~/.cargo/registry
>           key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
>       - name: Cache cargo index
>         uses: actions/cache@v1
>         with:
>           path: ~/.cargo/git
>           key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
>       - name: Cache cargo build
>         uses: actions/cache@v1
>         with:
>           path: rust/target
>           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
> ```

Thanks for the review👍
[Just searched in the github community and it seems that it is currently not possible to group and share steps between jobs.](https://github.community/t5/GitHub-Actions/reusing-sharing-inheriting-steps-between-jobs-declarations/m-p/37849#M3100)

.github/workflows/rust.yml
little-dude: But we would also shave time off by skipping the install step no?

Robert-Steiner: Right, but the installation step takes less time.
I just tested it again. The `Initialize containers` step takes 39 sec and the `install nightly` 17 sec.

little-dude: I see. Well since you've been working on improving the build time with the cache I think it makes sense to follow the same road and optimize for time as well.

.github/workflows/rust.yml
Robert-Steiner: fixed

Robert-Steiner:
> looks good to me (given my limited knowledge on CIs and github actions)

Thanks for the review

PR #405, opened 2020-05-18
PB-724 Add redis

Robert-Steiner:
**References**

- [PB-724](https://xainag.atlassian.net/browse/PB-724)

**Summary**

- added redis to docker-compose file

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     volumes:
       - minio-data:/data
     networks:
-      - xain-fl-rs
+      - xain-fl
     environment:
       MINIO_ACCESS_KEY: minio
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -13,8 +13,19 @@ services:
     ports:
       - "9000:9000"
 
+  redis:
+    image: redis:6.0.2 # Stable redis
+    hostname: redis
+    volumes:
+      - redis-data:/data
+    networks:
+      - xain-fl-rs
```
janpetschexain, line 22: i noticed that we still have the `-rs` rust suffix in our docker files and should probably rename that without using a suffix at one point (maybe not now).

Robert-Steiner: Fixed 👍

PR #408, opened 2020-05-19
Redis integration (Async coordinator poc)

Robert-Steiner:
First try to build an async version of the coordinator. The async version uses redis to maintain its state. In addition, the aggregation part was removed to simplify the implementation. I took some ideas from the [mini-redis project](https://github.com/tokio-rs/mini-redis/blob/master/src/server.rs). 

I will add some more comments later.


rust/src/coordinator.rs
```diff
@@ -105,1045 +119,590 @@ impl RoundSeed {
...
+    async fn proceed_sum_phase(&mut self) {
         info!("going to sum phase");
         self.gen_round_keypair();
-        *self.phase_mut() = Phase::Sum;
-        self.emit_event(ProtocolEvent::StartSum(self.round_parameters()));
+        self.state.phase = Phase::Sum;
+        self.set_state().await;
+
+        // We want to process as many messages as possible at the same time but we don't want to
+        // process more than is necessary (min_sum).
```
janpetschexain, line 232: we need to be careful about two things here:
1) if the coordinator only takes the bare minimum of `sum` participants in the `Sum` phase (meaning around `min_sum` many) and then too many of those drop out and are unavailable during the `Sum2` phase, then the round will abort. therefore we should choose a safety margin, maybe a certain percentage, or split the parameters into `min_sum` and `min_sum2` where `min_sum > min_sum2`.
2) if the coordinator selects only a subset of available participants, then selection bias must be avoided. for example only picking the first `m` out of `n` available participants or having a too short phase. not only does this mean a selection bias for faster devices/hardware but it might also lead to attacks where eligible participants gain an advantage by having better latency or processing power. a way to do this might be to have a reasonably large time frame for the phases (i know that this is a vague term, it should be configurable in the end) and to select a random subset of all available participants for that phase if there are to many of them.

if we play strictly by the rules of the protocol we wouldn't do any of the above but just abort the round and adjust the `sum` and `update` fractions for the next round accordingly. this avoids the issues explained above at the potential cost of some overhead, but in the end selecting good initial parameters for the protocol is the responsibility of the people setting up a specific federated learning use case.

Robert-Steiner: Thanks for the comment.

1. If I understand it correctly, the coordinator can still unmask the global model, even if not all sum participants (who responded in the sum phase) do not respond in the sum2 phase?

janpetschexain: exactly. technically a single valid `sum2` message suffices to unmask the global model. all sum participants are doing just redundant work. the main reasons for this are resiliency against sum participant dropouts and against malicious sum participants.

Robert-Steiner:
closed in favor of #413

PR #409, opened 2020-05-22
Redis Examples

Robert-Steiner:
I have created some small examples that uses Redis features that may be of interest for future development. 

# Pub/Sub

A publisher sends messages to a subscriber via the Pub/Sub feature of Redis. 

1. `cargo run --bin publisher`
2. `cargo run --bin subscriber`

# Lua scripts

An example that performs the `add_local_seed_dict` function directly on Redis. 
The advantage is, that the whole function is now performed as an atomic operation on Redis. 
In addition, it makes the usage of the cache in the [L572](https://github.com/xainag/xain-fl/pull/408/files#diff-7da06f52711366c73f68d43cd22e7a93R572) redundant. However, I don't know if this script is causing performance problems.

```rust
...
if local_seed_dict.keys().len() == self.sum_dict().keys().len()
    && local_seed_dict.keys()all(|pk| self.sum_dict().contains_key(pk))
    && self.seed_dict().values().next().map_or(true, |dict| !dict.contains_key(pk)) {
            for (sum_pk, seed) in local_seed_dict {
                self.seed_dict_mut()
                    .get_mut(sum_pk)
                    .ok_or(PetError::InvalidMessage)?
                    .insert(*pk, seed.clone());
            }
 }
...
```

`cargo run --bin lua_script`

# Client side caching

An example that shows, how Redis notifies a client of an updated value. Client-side caching could be a solution to keep services that depend on the coordinator state in sync with the coordinator.

`cargo run --bin client_side_caching`

---

I have added some more description in the code.

--- 

# Coordinator in Redis (Just for fun)

An example in which the entire coordinator state machine (without the crypto stuff) is executed in Redis via lua scripts.
The performance was not very good when I ran Redis via Docker. The update phase took over a minute for 1k sum and 1k update participants. However, when I ran Redis directly on my machine, the update phase only took 6-7 seconds.

`cargo run --release --bin coordinator`


rust/src/bin/coordinator_in_redis.rs
```diff
@@ -0,0 +1,265 @@
...
+        end
+        
+        redis.call("INCR", "round")
+        
+        local sum_keys = redis.call("HKEYS", "sum_dict")
+        
+        redis.call("DEL", "sum_dict")
+        redis.call("DEL", "update_participants")
+        
+        for i, sum_pk in pairs(sum_keys) do
```
little-dude, line 148: Do you know if redis optimizes the whole script?
Does it executes each `redis.call` one after another sequentially? Or is it able to parallelize  some calls, or batch them, or be lazy (for instance instead of downloading all the keys into `sum_keys` and then iterate through it, create an iterator and retrieve each key when `"DEL"` is called)

Robert-Steiner: I'm not sure if you can parallelize function calls to be honest but I know that `pairs` creates a stateless iterator. But I'm sure there is a lot that can be optimized. 😁

Just one point that I want to highlight about scripts because I don't know if it's so clear. Scripts run directly on the data of the redis instance. There are no network calls involved in `redis.call`, it basically corresponds to the call `self.sum_dict_mut().clear();` in rust.
So the client/ in our case the coordinator sends the script only once (redis caches the script in memory and returns an id that can be used to call that script again). Every further call only contains the id of the script and the parameters (for the update seed dict function, it would be the local_seed_dict and the update_pk).

PR #410, opened 2020-05-25
re-add unmasking logic

little-dude:
This was removed by mistake in
https://github.com/xainag/xain-fl/pull/398

### DIFF
--- a/rust/src/coordinator.rs
+++ b/rust/src/coordinator.rs
@@ -2,6 +2,7 @@ use std::{
     cmp::Ordering,
     collections::{HashMap, VecDeque},
     default::Default,
+    mem,
 };
 
 use sodiumoxide::{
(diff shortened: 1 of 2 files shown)


rust/src/coordinator.rs
```diff
@@ -496,12 +508,27 @@ impl Coordinator {
...
-        self.emit_event(ProtocolEvent::EndRound(outcome));
         self.start_new_round();
     }
 
+    fn end_round(&mut self) -> Result<Model, RoundFailed> {
+        let global_mask = self.freeze_mask_dict()?;
+        self.aggregation
+            .validate_unmasking(&global_mask)
+            .map_err(RoundFailed::from)?;
+        let aggregation = mem::replace(&mut self.aggregation, Aggregation::new(self.mask_config));
```
Robert-Steiner, line 528: First I was a bit confused about `mem::replace` but is a really nice solution here

little-dude: I'm not sure about it tbh.

On one hand, we could make `Aggregation.unmask` take `&mut` instead of `mut` and we wouldn't need these convolutions with `mem::replace`. But on the other hand I like the idea that the aggregation consumes the masked model to produce the model.

janpetschexain: the replacement of the old aggregate with `Aggregation::new()` happens twice now, here and in `start_new_round()`. also, this introduces different behavior for the aggregate compared to all other round parameters, therefore i would suggest to clear the aggregate next to the other round parameters in `start_new_round()`. this means that the aggregate doesn't get consumed to produced a model, but the same holds true for example for the dicts, which are also used to produce the model in the end, nevertheless they are not consumed but cleared after the round has finished.

janpetschexain: if we want to have a behavior somewhere "in between", we could probably treat the aggregate similar to the dicts and freeze it if needed.

little-dude: > therefore i would suggest to clear the aggregate next to the other round parameters in `start_new_round()`. this means that the aggregate doesn't get consumed to produced a model, but the same holds true for example for the dicts, which are also used to produce the model in the end, nevertheless they are not consumed but cleared after the round has finished.

Yeah I'm ok with that. I think it's pretty easy to do so I'll take a stab at it tomorrow.

little-dude:
> i don't want to block the testing session, therefore approval. we should implement a unified approach to all round parameters later on.

Thanks. I'm merging that then and I'll implement your suggestion tomorrow.

PR #412, opened 2020-05-29
PB-670 C API for FL

janpetschexain:
**References**
- [PB-670](https://xainag.atlassian.net/browse/PB-670)

**Summary**
- ~the `Client` introduced here is just a mock for now, it should be replaced by the networking client or a wrapper around that when that is merged;~ also some of the logic needs to be implemented then marked as `TODO`s
- generics as well as methods can't be exposed to C, therefore most of this code exposes functions for each primitive data type
- implement getters for the global model: the global model is converted to a model of primitive type and cached, a "mutable slice" is returned to the caller
- implement setters for the local model: take a "mutable slice" and convert it's data to a model, ~provides a new iterator for that if the slice doesn't point to the cached model~
- implement a callback: this is currently only in `new_client()` as an experiment/example, but is already quite general in that it is stateful and generic regarding the input (as long as both sides across the FFI-boundary agree on the same structs, the test simulates such a callback)

### DIFF
--- a/rust/src/coordinator.rs
+++ b/rust/src/coordinator.rs
@@ -107,6 +107,7 @@ pub struct Coordinator {
     seed: RoundSeed,
     min_sum: usize,
     min_update: usize,
+    expected_participants: usize,
     phase: Phase,
 
     // round dictionaries
@@ -137,7 +138,7 @@ pub enum ProtocolEvent {
 
     /// The sum phase finished and produced the given sum
     /// dictionary. The coordinator is now in the update phase.
-    StartUpdate(SumDict),
+    StartUpdate(SumDict, f64),
 
     /// The update phase finished and produced the given seed
     /// dictionary. The coordinator is now in the sum2 phase.
@@ -157,6 +158,7 @@ impl Default for Coordinator {
         let seed = RoundSeed::zeroed();
         let min_sum = 1_usize;
         let min_update = 3_usize;
+        let expected_participants = 10_usize;
         let phase = Phase::Idle;
         let sum_dict = SumDict::new();
         let seed_dict = SeedDict::new();
@@ -177,6 +179,7 @@ impl Default for Coordinator {
             seed,
             min_sum,
             min_update,
+            expected_participants,
             phase,
             sum_dict,
             seed_dict,
@@ -525,7 +528,8 @@ impl Coordinator {
         info!("going to update phase");
         self.freeze_sum_dict();
         self.phase = Phase::Update;
-        self.emit_event(ProtocolEvent::StartUpdate(self.sum_dict.clone()));
+        let scalar = 1_f64 / (self.expected_participants as f64 * self.update);
+        self.emit_event(ProtocolEvent::StartUpdate(self.sum_dict.clone(), scalar));
     }
 
     /// End the update phase and proceed to the sum2 phase.
@@ -803,7 +807,7 @@ mod tests {
         coord.try_phase_transition(); // finish the sum phase
         assert_eq!(
             coord.next_event().unwrap(),
-            ProtocolEvent::StartUpdate(sum_dict.clone())
+            ProtocolEvent::StartUpdate(sum_dict.clone(), 1.0)
         );
         assert!(coord.next_event().is_none());
         assert_eq!(coord.phase, Phase::Update);
@@ -963,7 +967,7 @@ mod tests {
         coord.try_phase_transition();
         assert_eq!(
             coord.next_event().unwrap(),
-            ProtocolEvent::StartUpdate(sum_dict)
+            ProtocolEvent::StartUpdate(sum_dict, 1.)
         );
         assert!(coord.next_event().is_none());
         assert_eq!(coord.phase, Phase::Update);
--- a/rust/src/sdk/api.rs
+++ b/rust/src/sdk/api.rs
@@ -76,24 +76,27 @@ PrimModel! {f64, c_double}
 PrimModel! {i32, c_int}
 PrimModel! {i64, c_long}
 
-/// Generates a consuming iterator for the primitive model. The iterator is wrapped in a private
-/// submodule, because safe traits and their safe methods can't be implemented as `unsafe` and this
-/// way we can ensure that the implementation is only ever used in an `unsafe fn`. The argument
-/// `$prim` is the corresponding Rust primitive data type.
-macro_rules! PrimIter {
-    ($prim:ty $(,)?) => {
-        paste::item! {
-            mod [<prim_iter_ $prim>] {
-                use std::{mem, os::raw::c_ulong};
+/// The iterators are wrapped in a private submodule, because safe traits and their safe methods
+/// can't be implemented as `unsafe` and this way we can ensure that the implementation is only
+/// ever used in an `unsafe fn`.
+mod iter {
+    use std::{mem, os::raw::c_ulong};
 
+    /// Generates a consuming iterator for the primitive model.  The argument
+    /// `$prim` is the corresponding Rust primitive data type.
+    macro_rules! PrimIter {
+        ($prim:ty $(,)?) => {
+            paste::item! {
                 use super::[<PrimitiveModel $prim:upper>];
 
+                #[doc(hidden)]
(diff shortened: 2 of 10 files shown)


rust/src/sdk/api.rs
```diff
@@ -0,0 +1,448 @@
...
+                }
+            }
+
+            impl Iterator for [<IntoIter $prim_rust:upper>] {
+                type Item = $prim_rust;
+
+                /// Advances the iterator and returns the next primitive value. Returns `None` when
+                /// the iteration is finished.
+                ///
+                /// # Safety
```
little-dude, line 74: I don't think we can do that even if we document it unfortunately, because this breaks the unsafe contract: a function that can result in UB if fed with the wrong input must be marked as unsafe.

janpetschexain: we can't alter the trait's definition, but we can wrap this in a private module such that nobody can accidentally access this. then this is only called in functions which are marked as unsafe anyways.

rust/src/sdk/api.rs
```diff
@@ -0,0 +1,448 @@
...
+///
+/// # Safety
+/// The method dereferences from the raw pointer arguments. Therefore, the behavior of
+/// the method is undefined if the arguments don't point to valid objects.
+pub unsafe extern "C" fn is_next_round(client: *mut Client) -> bool {
+    if client.is_null() {
+        // TODO: add error handling
+        panic!("invalid client");
+    }
+    let client = &mut *client;
```
little-dude, line 180: converting a raw pointer to a reference is unsafe. Doesn't rustc complain that this should be within an unsafe block?

janpetschexain: it actually is inside an unsafe block, because the whole function is declared unsafe.

little-dude: I see. I'd still put it within an `unsafe` block to make it more obvious which part of the function is actually unsafe. It makes auditing the code easier.

rust/src/sdk/api.rs
```diff
@@ -0,0 +1,448 @@
...
+///
+/// # Safety
+/// The method dereferences from the raw pointer arguments. Therefore, the behavior of
+/// the method is undefined if the arguments don't point to valid objects.
+pub unsafe extern "C" fn is_update_participant(client: *mut Client) -> bool {
+    if client.is_null() {
+        // TODO: add error handling
+        panic!("invalid client");
+    }
+    let client = &mut *client;
```
little-dude, line 207: We don't need mutability in this function.

rust/src/sdk/api.rs
janpetschexain: i was thinking about this as well, i definitely see the benefit of a pinpointing `unsafe` block. but in case of multiple `unsafe` blocks per function, this might become harder to read, hence i chose the `unsafe fn`. also, `unsafe fn` makes this very clear to the caller, otherwise it would look like a safe `fn` on first glance.

if we use both `unsafe fn` and `unsafe` blocks at the same time, we get a redundancy warning:
```
unnecessary `unsafe` block
`#[warn(unused_unsafe)]` on by defaultrustc(unused_unsafe)
api.rs: because it's nested under this `unsafe` fn
```
we might just disable the warning here for the sake of having `unsafe` blocks as well. what do you think?

little-dude: I'm not sure actually because I have very little experience with unsafe. We should ask the community about this. The rust-users forum seems like a good place to get feedback.

rust/src/sdk/api.rs
```diff
@@ -0,0 +1,448 @@
...
+            /// are violated.
+            pub unsafe extern "C" fn [<update_model_ $prim>](
+                client: *mut Client,
+                model: [<PrimitiveModel $prim:upper>],
+            ) {
+                if !client.is_null()
+                    && !model.ptr.is_null()
+                    && model.len != 0
+                    && model.len <= usize::MAX as c_ulong
+                    && model.len as usize * mem::size_of::<$prim>() <= usize::MAX
```
little-dude, line 307: This can overflow if `len` is set to an absurd value.

rust/src/sdk/api.rs
janpetschexain: good catch, i'll change it. i'm wondering why the compiler didn't say anything.

rust/src/sdk/api.rs
janpetschexain: i'll merge line 306 and 307 to
```
&& model.len <= (usize::MAX / mem::size_of::<$prim>()) as c_ulong
```
line 306 was only meant as a pre-check for 307, but as you noted it is necessary but not sufficient.

rust/src/sdk/api.rs
little-dude: This seems relevant: https://github.com/rust-lang/rust/issues/69173

rust/src/sdk/api.rs
little-dude: That sounds like a good idea to me :+1:

rust/src/sdk/api.rs
janpetschexain: thanks for the link. i would suggest that we then use `unsafe fn` in combination with explicit `unsafe` blocks. also, we shouldn't disable the warning and just keep it emitting until the lint change gets merged.

janpetschexain:
> This looks fine to me but I'm not really qualified to review `unsafe` code. So I think we should be very careful here:
> 
> * when the dust settles we should have this code audited
> * in the meantime, we should have a disclaimer in each module that contains unsafe code, stating that this code has not undergone an audit yet, and should not be used in production as is
> 
> As discussed in the comments, maybe we should add `unsafe` blocks within the unsafe functions, to visually highlight the unsafe portions and make them greppable. Each unsafe block should come with a comment explaining why what we're doing is safe.
> 
> Note: I haven't reviewed the whole thing yet

i included the suggested changes up until here.

janpetschexain:
the `Iterator` impl on the raw slices is removed and can't leak anymore. it got replaced by an iteration over a slice build from raw parts which is completely contained in the unsafe functions.

docs are updated as well, but the code for documenting the output of expanded macros is quite ugly. the main issue is that `paste::item!` doesn't work with `stringify!`/`concat!`, if anybody knows how to do this in a better way i'm happy to hear.

rust/src/sdk/api.rs
```diff
@@ -0,0 +1,542 @@
...
+pub unsafe extern "C" fn is_next_round(client: *mut Client) -> bool {
+    if client.is_null() {
+        // TODO: add error handling
+        panic!("invalid client");
+    }
+    let client = unsafe {
+        // safe if the raw pointer `client` comes from a valid allocation of a `Client`
+        &mut *client
+    };
+    let is_next_round = client.has_new_coord_pk_since_last_check;
```
Robert-Steiner, line 189: Just an idea: We could also use Option<()> here and write:
```
let is_next_round = client.has_new_coord_pk_since_last_check.take();
```

Robert-Steiner: Or another solution 
```
let is_next_round = std::mem::replace(client.has_new_coord_pk_since_last_check, false);
```

rust/src/coordinator.rs
```diff
@@ -525,7 +528,8 @@ impl Coordinator {
         info!("going to update phase");
         self.freeze_sum_dict();
         self.phase = Phase::Update;
-        self.emit_event(ProtocolEvent::StartUpdate(self.sum_dict.clone()));
+        let scalar = 1_f64 / (self.expected_participants as f64 * self.update);
```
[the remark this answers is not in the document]
janpetschexain: i decided against making the `scalar` part of the general round parameters, because it is only of use to update participants and it is only used during the composition of the update message. it could in theory be part of the general round parameters because its value is known at the beginning of a round though.

i suggest that we keep it as part of the update phase data for now, and once the extension to the pet protocol for arbitrary scalars is going to be implemented, this scalar is not going to be communicated at all.

rust/src/sdk/api.rs
janpetschexain: good ideas, i'll change that.

PR #413, opened 2020-06-02
Async Coordinator

Robert-Steiner:
Implementation of an async coordinator by using the state machine pattern described in https://blog.yoshuawuyts.com/state-machines/.

In the current version of the coordinator, the state is maintained by the `phase field` which an enum with the variants `Sum`,  `Update`, `Sum2` and `Idle`.

The idea described in the above blog post is, to utilise the rust type system so that only valid state transitions are performed.  Any invalid transition leads to a compiler error. My initial idea was to implement the state machine shown below. In addition to the existing states, I added another state: the `Error` state. It is used to perform the transition from a failed round/phase to a new round.

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic3RhdGVEaWFncmFtXG5cdFsqXSAtLT4gSWRsZVxuICBTdW0yIC0tPiBbKl1cbiAgXG4gIElkbGUgLS0-IFN1bVxuICBTdW0gLS0-IFVwZGF0ZVxuICBVcGRhdGUgLS0-IFN1bTJcbiAgU3VtMiAtLT4gSWRsZVxuXG4gIFN1bSAtLT4gRXJyb3JcbiAgVXBkYXRlIC0tPiBFcnJvclxuICBTdW0yIC0tPiBFcnJvclxuICBFcnJvciAtLT4gSWRsZVxuXG5cblxuXG5cblxuICAiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)

~~However, I noticed that the state machine pattern does not allow me to create branches and I don't know how to solve this.~~ Besides that, the pattern makes it easy to encapsulate shared functions/data (shared by all states) and state specific functions/data. Nevertheless, it creates a lot of visual noise and it is pretty easy to lose track of the structure.

**Parallel message processing**

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcblx0QlttZXNzYWdlX3J4XSAtLT4gQ3tsb29wfVxuXHRDIC0tPnx0b2tpbzpzcGF3bnwgRFtNZXNzYWdlVmFsaWRhdG9yMV1cblx0QyAtLT58dG9raW86c3Bhd258IEVbTWVzc2FnZVZhbGlkYXRvcjJdXG5cdEMgLS0-fHRva2lvOnNwYXdufCBGW01lc3NhZ2VWYWxpZGF0b3JOXVxuICBEIC0tPiB8c2lua190eHwgR3tNZXNzYWdlU2lua31cbiAgRSAtLT4gfHNpbmtfdHh8IEdcbiAgRiAtLT4gfHNpbmtfdHh8IEdcblx0XHRcdFx0XHQiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)


rust/src/coordinator_async.rs
```diff
@@ -0,0 +1,275 @@
...
+                coordinator_state: CoordinatorState {
+                    seed: RoundSeed::generate(),
+                    ..Default::default()
+                },
+                message_rx: rx,
+            },
+        ))
+    }
+
+    pub async fn next(self) -> State<Idle> {
```
little-dude, line 97: Could you explain where `State.next` is called and why it needs to be async? I don't see it.

Robert-Steiner: There will be a `run` that calls `next()`. Something like this:

```rust
async fn run() {
    let (_, state_start) = State::new().unwrap(); // Start
    let mut state_idle = state_start.next(); // Idle

    loop {
        let state_sum = state_idle.next().await; // Sum
        let state_update = state_sum.next().await; // Update
        let state_sum2 = state_update.next().await; // Sum2
        state_idle = state_sum2.next().await; // Idle
    }
}
```

You are right, the `next` method in the `Idle` state doesn't have to be async.

rust/src/coordinator_async/message_processing.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    async fn handle_message(
+        self,
+        message_validation_fut: impl Future<Output = Result<(), PetError>>,
+    ) {
+        // Extract all fields of the MessageHandler struct. This is necessary to bypass borrow
+        // issues in the tokio::select macro.
+        // It is important to extract _cancel_complete_tx as well, otherwise the channel
+        // will be dropped too early.
```
janpetschexain, line 150: this issue might be avoided by not destructuring `self`, but instead use `self.sink_tx` and `self.notify_cancel` in the `select!` macro. then `self` will be dropped at the end of the method. this would require `mut self` instead of `self` as an argument to `handle_message()`.

rust/src/coordinator_async/message_processing.rs
```diff
@@ -0,0 +1,278 @@
...
+            result = message_validation_fut => {let _ = sink_tx.send(result);}
+            _ = notify_cancel.recv() => {println!("drop message validation future")}
+        };
+
+        // _cancel_complete_tx is dropped
+    }
+}
+
+// Sum message validator
+impl MessageHandler {
```
janpetschexain, line 167: why not combine the three `impl MessageHandler { }` blocks into one block?

rust/src/coordinator_async/message_processing.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub update: f64,
+    pub seed: RoundSeed,
+}
+
+// A sink to collect the results of the MessageHandler tasks.
+pub struct MessageSink {
+    // The minimum number of successfully validated messages.
+    min_messages: usize,
+    // A counter that is incremented when a message has been successfully validated.
+    successful_messages: usize,
```
janpetschexain, line 36: i'm still undecided if we should actually short-circuit a phase. selection bias can have a negative impact on the model quality and this choice of aborting earlier than the `max_duration` can have various consequences:
- update participants with less training data will probably be faster and therefore have a higher representation in the aggregated model, even though they contribute less
- malicious update participants who just submit garbage immediately will have a higher representation in the aggregated model
- update participants from different parts of the world with slower internet will have a lower representation in the aggregated model, even though the model would benefit from more different and less similar data points
- etc

basically, the `sum`/`update` thresholds of the protocol give a `mean_messages` value we can expect, therefore we should only check that the `successful_messages` are greater than the minimum number of message required for a save protocol and otherwise just wait for the timeout. this was perviously called `min_sum`/`min_update` in the `Coordinator`, but i'm not sure if the `min_messages` of the `MessageSink` is supposed to have the same meaning?

rust/src/coordinator_async/sum.rs
```diff
@@ -0,0 +1,111 @@
...
+
+            Err(_) => StateMachine::Error(State {
+                _inner: Error {},
+                coordinator_state: self.coordinator_state,
+                message_rx: self.message_rx,
+            }),
+        }
+    }
+
+    async fn run(&mut self) -> Result<(), PetError> {
```
janpetschexain, line 45: all the cloning in this method seems quite wasteful, is this because of the aforementioned borrow checker issues for the `select!` macro?

PR #414, opened 2020-06-05
WIP: use tower

little-dude:
TODO:

1. ~~remove the transport and clients code. I think we actually don't need it since we can make the services clonable by wrapping them in a `tower::Buffer` layer. If so, that would be a huge simplication~~ DONE
2. integrate with https://github.com/xainag/xain-fl/pull/419. I'm a bit worried here because I'm not sure yet how much change is needed. Maybe little, but maybe a lot. In particular, in this PR I've simplified the state machine a lot and reduced it to three phases, while #419 actually adds new states.
3. update the client. There will definitely be conflict here, so I need to take a look at https://github.com/xainag/xain-fl/pull/420 first
4. the usual tests and doc


Robert-Steiner:
I really like the whole service and layer idea. The services `decrypt` to `route` are the services that are missing in my implementation. I also wanted to filter the messages by the phase so that only the requests that belong to this phase are processed. However, it was difficult for me to implement that idea in the state machine pattern. I tried to use generics to but it didn't worked out. The way you solved it with tower is really nice.
Do you know whether it will later be possible to process the request concurrently / in parallel?
I'm not sure if we can merge your ideas easily because your approach is more based on the request type and my is more based on the phase. But to be honest, I like the tower approach more, because the logic is more strictly separated which makes it lees prone to errors in my opinion.

little-dude:
> Do you know whether it will later be possible to process the request concurrently / in parallel?

Yeah the requests can be parallelized. However, when updating the dictionaries, we can only handle requests sequentially. The service that does that will kind of act like a lock. This can be done, I think by setting a concurrently limit of `1` on that service (https://docs.rs/tower/0.3.1/tower/builder/struct.ServiceBuilder.html#method.concurrency_limit), and adding a queue for pending requests (https://docs.rs/tower/0.3.1/tower/builder/struct.ServiceBuilder.html#method.buffer)

little-dude:
Dumping some notes to not forget:

- message parsing is a real footgun. I just spent an hour finding a bug there. The `Decrypt` layer should really return three vec: one for the signature, one for the header and one for the payload.
- I added a bunch of `<S as Service<Xxx>::Future: 'static + Send` bounds, to make rustc happy, but it may not be necessary everywhere. Should try to remove some of them
- We need one more service for the round parameters
- I'm currently using `tokio_tower::pipeline` whereas we should be using `tokio_tower::multiplex`
- Needs logging, and tests (do I even need to write that down?)

little-dude:
re labeled the PR as WIP since I'm planning to try to make significant simplifications by removing the transport and tower client logic.

PR #415, opened 2020-06-08
WIP fix to poll round params data

finiteprods:
**update**. This addresses a break in the build introduced in https://github.com/xainag/xain-fl/pull/397

- now compatible with the changes around round parameters introduced in https://github.com/xainag/xain-fl/pull/411
- [PB-783](https://xainag.atlassian.net/browse/PB-783)

--------------------
This compiles, but isn't quite as it should be. In the loop of `during_round`, we poll for `round_params_data`. Now the "old" `round_params` are inside, and also `Option`al. So really we should be polling until `round_params` is _also_ available. What we'd really like is something like
```rust
// doesn't compile!
let round_params = loop {
    if let Some(round_params_data) = self.handle.get_round_parameters().await {
        if let Some(round_params) = round_params_data.round_parameters {
            break round_params;
        }
    }
    // interval tick...
};
```
But the `Arc` around `round_params_data` makes it problematic to do...
```rust
error[E0507]: cannot move out of an `Arc`
   --> src/client.rs:108:45
    |
108 |                 if let Some(round_params) = round_params_data.round_parameters {
    |                             ------------    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider borrowing here: `&round_params_data.round_parameters`
    |                             |
    |                             data moved here
    |                             move occurs because `round_params` has type `coordinator::RoundParameters`, which does not implement the `Copy` trait
```

### DIFF
--- a/rust/src/client.rs
+++ b/rust/src/client.rs
@@ -48,6 +48,9 @@ pub struct Client {
     /// Interval to poll for service data
     interval: time::Interval,
 
+    /// Coordinator public key
+    _coordinator_pk: CoordinatorPublicKey,
+
     id: u32, // NOTE identifier for client for testing; may remove later
 }
 
@@ -64,6 +67,7 @@ impl Client {
             handle,
             participant,
             interval: time::interval(Duration::from_secs(period)),
+            _coordinator_pk: CoordinatorPublicKey::zeroed(),
             id: 0,
         })
     }
@@ -80,6 +84,7 @@ impl Client {
             handle,
             participant,
             interval: time::interval(Duration::from_secs(period)),
+            _coordinator_pk: CoordinatorPublicKey::zeroed(),
             id,
         })
     }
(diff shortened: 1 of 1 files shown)


janpetschexain:
one way would be like this:
```
let round_params = loop {
    if let Some(round_params_data) = self.handle.get_round_parameters().await {
        if let Some(ref round_params) = round_params_data.round_parameters {
            break round_params.clone();
        }
    }
    debug!(client_id = %self.id, "round params not ready, retrying.");
    self.interval.tick().await;
};
```
cloning is quite cheap for this object.

janpetschexain:
but it might be even better to move the code that acts on the returned `round_params` inside the inner `if let` clause, such that it can use the borrowed values without cloning:
```rust
pub async fn during_round(&mut self) -> Result<Task, ClientError> {
    loop {
        if let Some(round_params_data) = self.handle.get_round_parameters().await {
            if let Some(ref round_params) = round_params_data.round_parameters {
                let round_seed: &[u8] = round_params.seed.as_slice();
                debug!(client_id = %self.id, "computing sigs and checking task");
                self.participant.compute_signatures(round_seed);
                let (sum_frac, upd_frac) = (round_params.sum, round_params.update);
                match self.participant.check_task(sum_frac, upd_frac) {
                    Task::Sum => break self.summer(round_params.pk).await,
                    Task::Update => break self.updater(round_params.pk).await,
                    Task::None => break self.unselected().await,
                }
            }
        }
        debug!(client_id = %self.id, "round params not ready, retrying.");
        self.interval.tick().await;
    }
}
```

finiteprods:
nice suggestions @janpetschexain, thanks. I'll go with your first only because less excessive indenting seems worth the price of the clone! but i'd be happy to go with your other suggestion too.

janpetschexain:
> nice suggestions @janpetschexain, thanks. I'll go with your first only because less excessive indenting seems worth the price of the clone! but i'd be happy to go with your other suggestion too.

i'm tending towards the second solution, since we don't have a limit on how often this code is called per FL round, which especially means for the `Task::None` participants that the cloning will happen every few moments.

janpetschexain:
sry, i didn't mean to move it from draft to review. the button looked like the usual review button for me to add review comments. it also seems that i can't move it back to draft.

finiteprods:
> sry, i didn't mean to move it from draft to review. the button looked like the usual review button for me to add review comments. it also seems that i can't move it back to draft.

as you've probably seen already, gone with approach #2 and also (since this came up in review of https://github.com/xainag/xain-fl/pull/397 and wasn't addressed there), clients now track the current coordinator public key to distinguish between the current round and the next.

PR #416, opened 2020-06-08
Redis client

Robert-Steiner:
Summary: 

- added macros that implement `FromRedisValue` and `ToRedisArgs` for various crypto types, for the `CoordinatorState` and for the `MaskObject`
- added `redis::Client` (some methods might not be needed and will be removed after the integration into the state machine is done)

### DIFF
--- a/rust/xaynet-server/src/storage/redis.rs
+++ b/rust/xaynet-server/src/storage/redis.rs
@@ -459,6 +459,7 @@ mod tests {
     use super::*;
     use crate::state_machine::tests::utils::{mask_settings, model_settings, pet_settings};
     use num::{bigint::BigUint, traits::identities::Zero};
+    use serial_test::serial;
     use xaynet_core::{
         crypto::{EncryptKeyPair, SigningKeyPair},
         mask::{BoundType, DataType, GroupType, MaskConfig, MaskObject, ModelType},
@@ -490,6 +491,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn integration_set_and_get_coordinator_state() {
         // test the writing and reading of the coordinator state
         let client = init_client().await;
@@ -514,6 +516,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn integration_get_best_masks_one_mask() {
         // test the writing and reading of one mask
         let client = init_client().await;
@@ -535,6 +538,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn integration_get_best_masks_two_masks() {
         // test the writing and reading of two masks
         // the first mask is incremented twice
@@ -575,6 +579,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn integration_get_best_masks_no_mask() {
         // ensure that get_best_masks returns an empty vec if no mask exist
         let client = create_redis_client().await;
@@ -585,6 +590,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn integration_sum_dict() {
         // test multiple sum dict related methods
         let client = init_client().await;
@@ -653,6 +659,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn integration_flush_dicts_return() {
         let client = init_client().await;
 
--- a/rust/xaynet-server/src/storage/redis.rs
+++ b/rust/xaynet-server/src/storage/redis.rs
@@ -6,26 +6,26 @@
 //!     "coordinator_state": "...", // bincode encoded string
 //!     // Sum dict
 //!     "sum_dict": { // hash
-//!         "SumParticipantPublicKey_1": "SumParticipantEphemeralPublicKey_1",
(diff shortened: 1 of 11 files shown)


rsaffi:
Rebased against the updated `master`!

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -0,0 +1,644 @@
...
+                PublicEncryptKeyWrite::from(ephm_pk),
+            )
+            .await
+    }
+
+    /// Removes an entry in the [`SumDict`].
+    ///
+    /// Returns [`DeleteSumParticipant::Ok`] if field was deleted or
+    /// [`DeleteSumParticipant::DoesNotExist`] if field does not exist.
+    pub async fn remove_sum_dict_entry(
```
little-dude, line 197: Do we actually need to remove participants?

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -0,0 +1,644 @@
...
+    LocalSeedDict,
+    SeedDict,
+    SumDict,
+    SumParticipantEphemeralPublicKey,
+    SumParticipantPublicKey,
+    UpdateParticipantPublicKey,
+};
+
+#[macro_export]
+macro_rules! redis_connection {
```
little-dude, line 64: Couldn't this be a method of `Client` instead of a macro?

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -0,0 +1,644 @@
...
+        let seed_dict = result
+            .into_iter()
+            .map(|(pk, seed)| (pk.into(), seed.into()))
+            .collect();
+
+        Ok(seed_dict)
+    }
+
+    /// Retrieves the [`SeedDict`] or an empty [`SeedDict`] when the [`SumDict`] does not exist.
+    pub async fn get_seed_dict(mut self) -> RedisResult<SeedDict> {
```
little-dude, line 263: Do we ever need to retrieve the full seed dict? I thought we only needed parts of it for each sum participant.

little-dude (commented):
I still need some time to think about the "no rollback" implications for `update_seed_dict`. But the rest looks really good to me :+1: I just left a few comments mostly about the doc, along with a bunch of questions.

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -0,0 +1,644 @@
...
+//!         "UpdateParticipantPublicKey_1": "EncryptedMaskSeed",
+//!         "UpdateParticipantPublicKey_2": "EncryptedMaskSeed"
+//!     }
+//!     "SumParticipantPublicKey_2": {
+//!         "UpdateParticipantPublicKey_1": "EncryptedMaskSeed",
+//!         "UpdateParticipantPublicKey_2": "EncryptedMaskSeed"
+//!     }
+//!     // Mask dict
+//!     "mask_dict": [ // sorted set
+//!         ("mask_abc", 12341), // (mask: bincode encoded string, score/counter: number)
```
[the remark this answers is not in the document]
Robert-Steiner: It is the actual mask object. I can rename it to `mask_object_1` and `mask_object_2`. wdyt?

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: We remove the sum participant in the sum2 phase 
https://github.com/xaynetwork/xaynet/pull/519/files#diff-5b85450e9124bca33ff393eee368f61fR187
My idea was to add this command to the `incr_mask_count` method. However, we will then have the same issue that we have with the `update_seed_dict` method. (the mask count would increase even if the sum particpant is not in the sum dict)

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: I just tried it out and it is possible. I will remove the macro

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: Unfortunately yes, we broadcast the seed dict at the end of the update phase 
https://github.com/xaynetwork/xaynet/pull/519/files#diff-b15272399ff491c84d93d4ca104eaab1R77

Robert-Steiner:
Thanks for the review 🙂

> I still need some time to think about the "no rollback" implications for `update_seed_dict`. 

I agree, maybe we should set up a meeting to discuss it

rust/xaynet-server/src/storage/redis.rs
little-dude: Oh ok. With these changes though we shouldn't need to do so anymore. `xaynet_server::services::fetchers::SeedDictService` should actually talk to redis directly. We should just send an event saying that a new dict is available. Same goes for the sum dictionary. But better address that in a follow-up PR.

rust/xaynet-server/src/storage/redis.rs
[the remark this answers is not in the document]
little-dude: Sounds good :+1:

little-dude (approved):

PR #417, opened 2020-06-12
PB-774 client async runtime

janpetschexain:
**References**
- [PB-774](https://xainag.atlassian.net/browse/PB-774)

**Summary**
- wrap the `Client` and a tokio `Runtime` into a `FFIClient`
- create a custom runtime in `new_client()` and start the client in that runtime in `run_client()`
- remove the callback experiment and replace it with a purposeful callback in `run_client()`
- implement error handling across the FFI boundary
- update docs and tests
- add `new_model()` for a zero-initialized cached primitive model and restrict `update_model()` to take no foreign memory but only cached models
- simplify the type-suffixed functions by removing the suffix and replace it with type flags and void pointers instead

### DIFF
--- a/rust/src/sdk/api.rs
+++ b/rust/src/sdk/api.rs
@@ -70,9 +70,9 @@ use crate::{
 pub struct PrimitiveModel {
     /// A raw mutable pointer to an array of primitive values.
     pub ptr: *mut c_void,
-    /// The length of that respective array.
+    /// The length of that array.
     pub len: c_ulong,
-    /// The data type of that respective array.
+    /// The data type of the array's elements.
     pub dtype: c_uint,
 }
 
@@ -89,7 +89,7 @@ pub(crate) enum CachedModel {
 }
 
 #[derive(Debug)]
-/// A wrapper for a [`Client`] within a [`Runtime`].
+/// A wrapper for a [`Client`] within an asynchronous runtime.
 ///
 /// This is returned from [`new_client()`]. See the [workflow] on how to use it.
 ///
@@ -101,33 +101,27 @@ pub struct FFIClient {
 
 #[allow(unused_unsafe)]
 #[no_mangle]
-/// Creates a new [`Client`] within a [`Runtime`].
+/// Creates a new [`Client`] within an asynchronous runtime.
 ///
-/// Takes a `period` in seconds for which the [`Client`] will try to poll the coordinator for new
-/// broadcasted FL round data. Specifies the `min_threads` and `max_threads` available to the
-/// [`Runtime`].
+/// Takes a `period` in seconds after which the [`Client`] will try to poll the coordinator for new
+/// broadcasted FL round data again.
 ///
 /// # Errors
 /// Returns a null pointer in case of invalid arguments or if the initialization of the runtime
 /// or the client fails. The following must hold true for arguments to be valid:
 /// - `period` > 0
-/// - 0 < `min_thread` <= `max_thread` <= 32,768
 ///
 /// # Safety
 /// The method depends on the safety of the `callback` and on the consistent definition and layout
 /// of its `input` across the FFI-boundary.
-pub unsafe extern "C" fn new_client(
-    period: c_ulong,
-    min_threads: c_ulong,
-    max_threads: c_ulong,
-) -> *mut FFIClient {
-    if period == 0 || min_threads == 0 || max_threads < min_threads || max_threads > 32_768 {
+pub unsafe extern "C" fn new_client(period: c_ulong) -> *mut FFIClient {
+    if period == 0 {
         return ptr::null_mut() as *mut FFIClient;
     }
     let runtime = if let Ok(runtime) = Builder::new()
         .threaded_scheduler()
-        .core_threads(min_threads as usize)
-        .max_threads(max_threads as usize)
+        .core_threads(1)
+        .max_threads(4)
         .thread_name("xain-fl-client-runtime-worker")
         .enable_all()
         .build()
@@ -146,17 +140,17 @@ pub unsafe extern "C" fn new_client(
 
 #[allow(unused_unsafe)]
 #[no_mangle]
-/// Starts the [`Client`] and executes its tasks in an asynchronous [`Runtime`].
+/// Starts the [`Client`] and executes its tasks in an asynchronous runtime.
 ///
 /// Takes a `callback(state, code)` function pointer and a void pointer to the `state` of the
-/// callback. The callback will be triggered when the client must be stopped because of a panic or
+/// callback. The callback will be called when the client must be stopped because of a panic or
 /// error or when the client terminates successfully. See the [errors] section for the error `code`
 /// definitions.
 ///
 /// # Errors
-/// Ignores null pointer `client`s and triggers the callback immediately.
+/// Ignores null pointer `client`s and calls the callback immediately.
 ///
-/// Triggers the callback with one of the following error codes:
+/// Calls the callback with one of the following error codes:
 /// - `-1`: client didn't start due to null pointer
 /// - `0`: no error (only for clients with finite number of FL rounds)
 /// - `1`: client panicked due to unexpected/unhandled error
@@ -169,9 +163,8 @@ pub unsafe extern "C" fn new_client(
 /// The method dereferences from the raw pointer arguments. Therefore, the behavior of
 /// the method is undefined if the arguments don't point to valid objects.
 ///
-/// If the callback is triggered because of a panicking client, it is the users responsibility
-/// to not access possibly invalid state and to drop the client. If certain parts of the state
-/// can be guaranteed to be valid, they may be read before dropping.
+/// If the callback is called because of a panicking client, it is the users responsibility
+/// to not access possibly invalid client state and to drop the client.
 ///
 /// [errors]: #errors
 /// [`ParticipantInitErr`]: ../../client/enum.ClientError.html#variant.ParticipantInitErr
@@ -193,6 +186,16 @@ pub unsafe extern "C" fn run_client(
         // safe if the raw pointer `client` comes from a valid allocation of a `FFIClient`
         (&(*client).runtime, &mut (*client).client)
     };
+
+    // `UnwindSafe` basically says that there is no danger in accessing a value after a panic
+    // happened. this is generally true for immutable references, but not for mutable references.
+    // currently the docs have a note that it is the user's responsibility to not access possibly
+    // invalid values (we could clean up the client ourself but then the paradigm of create-use-
+    // destroy all happening on one side, in this case the non-Rust side of the API, is violated
+    // and might as well lead to severe bugs/segfaults). the main issue is that we must catch the
+    // panic, because letting it propagate across the FFI-boundary is undefined behavior and will
+    // most likely result in segfaults. what we can do is to improve error handling on our side to
+    // reduce the number of possible panics and return proper errors instead.
     let code = match panic::catch_unwind(unsafe {
         // even though `&mut Client` is `!UnwindSafe` we can assert this because the user will be
         // notified about a panic immediately to be able to safely act accordingly
@@ -216,9 +219,9 @@ pub unsafe extern "C" fn run_client(
 /// Stops and destroys a [`Client`] and frees its allocated memory.
 ///
 /// Tries to gracefully stop the client for `timeout` seconds by blocking the current thread before
-/// shutting it down forcefully (cf. the remarks about memory safety of the [runtime shutdown] in
-/// case of elapsed timeout). Usually, no timeout (i.e. 0 seconds) suffices, but stopping might take
-/// indefinitely if the client performs long blocking tasks.
+/// shutting it down forcefully (outstanding tasks are potentially leaked in case of an elapsed
+/// timeout). Usually, no timeout (i.e. 0 seconds) suffices, but stopping might take indefinitely
+/// if the client performs long blocking tasks.
 ///
 /// # Errors
 /// Ignores null pointer `client`s and returns immediately.
@@ -226,8 +229,6 @@ pub unsafe extern "C" fn run_client(
 /// # Safety
 /// The method dereferences from the raw pointer arguments. Therefore, the behavior of the method is
 /// undefined if the arguments don't point to valid objects.
-///
-/// [runtime shutdown]: https://docs.rs/tokio/0.2.21/tokio/runtime/struct.Runtime.html#method.shutdown_timeout
 pub unsafe extern "C" fn drop_client(client: *mut FFIClient, timeout: c_ulong) {
     if !client.is_null() {
         let client = unsafe {
@@ -646,7 +647,7 @@ mod tests {
 
     #[test]
     fn test_new_client() {
-        let client = unsafe { new_client(10, 1, 4) };
+        let client = unsafe { new_client(10) };
         assert!(!client.is_null());
         unsafe { drop_client(client, 0) };
     }
@@ -670,7 +671,7 @@ mod tests {
         }
 
         // check that the client panics when running it without a service
-        let client = unsafe { new_client(10, 1, 4) };
+        let client = unsafe { new_client(10) };
         let mut state = State {
             client_crashed: false,
             error_code: 0_i32 as c_int,
@@ -694,7 +695,7 @@ mod tests {
                 #[allow(unused_unsafe)]
                 #[test]
                 fn [<test_new_model_ $prim>]() {
-                    let client = unsafe { new_client(10, 1, 4) };
+                    let client = unsafe { new_client(10) };
 
                     // check that the new model is cached
                     let (model, _) = dummy_model(0., 10);
@@ -724,7 +725,7 @@ mod tests {
                 #[allow(unused_unsafe)]
                 #[test]
                 fn [<test_get_model_ $prim>]() {
-                    let client = unsafe { new_client(10, 1, 4) };
+                    let client = unsafe { new_client(10) };
 
                     // check that the primitive model is null if the global model is unavailable
                     assert!(unsafe { &*client }.client.global_model.is_none());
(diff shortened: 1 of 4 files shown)


rust/src/sdk/api.rs
```diff
@@ -87,85 +84,174 @@ macro_rules! PrimModel {
...
+/// [`ParticipantInitErr`]: ../../client/enum.ClientError.html#variant.ParticipantInitErr
+/// [`ParticipantErr`]: ../../client/enum.ClientError.html#variant.ParticipantErr
+/// [`DeserialiseErr`]: ../../client/enum.ClientError.html#variant.DeserialiseErr
+/// [`GeneralErr`]: ../../client/enum.ClientError.html#variant.GeneralErr
+pub unsafe extern "C" fn run_client(
+    client: *mut FFIClient,
+    callback: unsafe extern "C" fn(*mut c_void, c_int),
+    state: *mut c_void,
+) {
+    if client.is_null() {
```
little-dude, line 200: if `client` is `null` there clearly is a bug in the client code, and I'm not sure we should try to call the callback in that case.

rust/src/sdk/api.rs
```diff
@@ -87,85 +84,174 @@ macro_rules! PrimModel {
...
+        Ok(Err(ClientError::ParticipantErr(_))) => 3_i32,
+        Ok(Err(ClientError::DeserialiseErr(_))) => 4_i32,
+        Ok(Err(ClientError::GeneralErr)) => 5_i32,
+    } as c_int;
     unsafe {
-        // safe if the `callback` is sound and the same definition and layout is used for `Input`
-        // across the FFI-boundary by the caller
-        callback(state, input)
+        // safe if the `callback` is sound
+        callback(state, code)
```
little-dude, line 224: The doc states:

> The callback will be triggered when the client must be stopped because of a panic or error

But it is actually called even when there's no error.

rust/src/sdk/api.rs
janpetschexain: i'm not sure myself to be honest. the error code is part of the callback, so the user could match the code and act accordingly. otherwise we might just return the error code from `run_client()` and call the callback without the error code, but the error code might be useful for callbacks for regular errors. or we could just have a `!client.is_null()` in the beginning and ignore null pointers at all. what do you think?

rust/src/sdk/api.rs
janpetschexain: this callback is currently unreachable because `client.start()` will only ever stop if an error happens. we could either remove this specific call of the callback or update the docs?

rust/src/sdk/api.rs
```diff
@@ -87,85 +84,174 @@ macro_rules! PrimModel {
...
-        _model_cached: client.cached_model.is_some(),
-    } as *const _ as *const c_void;
+    let (runtime, client) = unsafe {
+        // safe if the raw pointer `client` comes from a valid allocation of a `FFIClient`
+        (&(*client).runtime, &mut (*client).client)
+    };
+    let code = match panic::catch_unwind(
+        // even though `&mut Client` is `!UnwindSafe` we can assert this because the user will be
+        // notified about a panic immediately to be able to safely act accordingly
+        panic::AssertUnwindSafe(|| runtime.handle().block_on(client.start())),
```
little-dude, line 213: I'm not sure about this. I don't fully grasp what `UnwindSafe` means and under which condition can we assert that `&mut Client` is `UnwindSafe`. But I'm a bit worried because:

- `Client.start` is very complex, so that makes the evaluation of unwind safety difficult
- even if we can prove that this is currently unwind safe, I'm worried we can make this unsafe without realizing it, by making changes to `Client` in the future.

So I have two questions:

- do we really need to `catch_unwind` here? Couldn't we just let it crash?
- should we add an `unsafe` block here?

rust/src/sdk/api.rs
```diff
@@ -242,33 +334,41 @@ macro_rules! get_model {
             #[doc = "to the latest global model.\n"]
```
little-dude, line 334: nit: I think we could make this a bit less noisy with a raw string literal in a single `#[doc]` attribute. Like:

```rust
#[doc = r#"some lines
another line
etc.
"#]
```

rust/src/sdk/api.rs
janpetschexain: `UnwindSafe` basically says that there is no danger in accessing a value after a panic happened. this is generally true for immutable references, but not for mutable references.

currently i've put the note that it is the user's responsibility to not access possibly invalid values (we could clean up the `client` ourself but then the paradigm of create-use-destroy all happening on one side, in this case the non-Rust side of the API, is violated and might as well lead to severe bugs/segfaults).

the main issue is that we must catch the panic, because letting it propagate across the FFI-boundary is undefined behavior and will most likely result in segfaults. what we can do is to improve error handling on our side to reduce the number of possible panics and return proper errors instead. (for example the first panic that is catched here is a panic due to closed channels if the client is started without a coordinator service. things like that should of course be proper errors, but in the end we can't rule out every possible panic and need to have a way to catch them.)

`unsafe` is not needed for `AssertUnwindSafe` according to the docs, but i agree that we should enclose this in an `unsafe` block nonetheless. there is some hidden docs about unwind safety:
```
// Implementations of the `UnwindSafe` trait:
//
// * By default everything is unwind safe
// * pointers T contains mutability of some form are not unwind safe
// * Unique, an owning pointer, lifts an implementation
// * Types like Mutex/RwLock which are explicitly poisoned are unwind safe
// * Our custom AssertUnwindSafe wrapper is indeed unwind safe
```

rust/src/sdk/api.rs
janpetschexain: good point, i didn't know that this is possible. it is indeed quite ugly to read due to the mixture of macros and docs.

rust/src/sdk/api.rs
janpetschexain: an interesting thing i just noticed is that `&mut T` is `!UnwindSafe` but `*mut T` is indeed `UnwindSafe`. i'm not sure why this is the case, but maybe we can leverage this since we are working with raw pointers most of the time anyways.

rust/src/sdk/api.rs
janpetschexain: unfortunately, i didn't find a way to make it work. raw string literals don't respect any of the markdown and linebreaks (this is of course intended, it is a raw string literal after all) and using just common string literals triggers some weird markdown formatting (code blocks etc). i guess we have to stick with the ugly one doc attribute per line.

rust/src/sdk/api.rs
janpetschexain: i updated the docs for now.

Robert-Steiner:
Since I have no experience with ffi I can't really help here. 
However, the documentation is great. 👍
 
One thing I wonder is how the memory of the models (local/global) is de-allocated. Does this happen on the rust or on the caller side?

janpetschexain:
> One thing I wonder is how the memory of the models (local/global) is de-allocated. Does this happen on the rust or on the caller side?

everything that is a part/field of the `Participant`/`Client`/`FFIClient` is allocated and dropped on the rust side. basically the call to `Box::into_raw()` leaks heap-allocated memory, whereafter we only have access to it via raw pointers. the `Box::from_raw()` counterpart allows the rust compiler to take ownership of this object again to drop it. everything that happens in between on the caller side can mutate the object, but it can't drop the object since the memory layout for structs/enums which are not `#[repr(C)]` is generally not known to compilers/garbage collectors outside of the rust compiler.

this leaves the user with the cases where the models are allocated outside of our function calls, for example when there is no global model available in the first round and the app therefore has to allocate and initialize a model outside of the rust code. currently we try to read this as a slice and allocate a local copy within the client. we might actually provide functions like `new_model_f32()` etc to have better control over memory or to make this a default option for the `get_model_f32()`. the latter would require a length information as an argument to the function. i'll ask blue what they prefer as well.

Robert-Steiner:
> > One thing I wonder is how the memory of the models (local/global) is de-allocated. Does this happen on the rust or on the caller side?
> 
> everything that is a part/field of the `Participant`/`Client`/`FFIClient` is allocated and dropped on the rust side. basically the call to `Box::into_raw()` leaks heap-allocated memory, whereafter we only have access to it via raw pointers. the `Box::from_raw()` counterpart allows the rust compiler to take ownership of this object again to drop it. everything that happens in between on the caller side can mutate the object, but it can't drop the object since the memory layout for structs/enums which are not `#[repr(C)]` is generally not known to compilers/garbage collectors outside of the rust compiler.
> 
> this leaves the user with the cases where the models are allocated outside of our function calls, for example when there is no global model available in the first round and the app therefore has to allocate and initialize a model outside of the rust code. currently we try to read this as a slice and allocate a local copy within the client. we might actually provide functions like `new_model_f32()` etc to have better control over memory or to make this a default option for the `get_model_f32()`. the latter would require a length information as an argument to the function. i'll ask blue what they prefer as well.

Ah I see, so the whole client is allocated on the heap, right?
But what happens if the caller holds a pointer to the global model and the client replaces this model. Is the pointer still valid?

janpetschexain:
> Ah I see, so the whole client is allocated on the heap, right?
> But what happens if the caller holds a pointer to the global model and the client replaces this model. Is the pointer still valid?

correct, the whole client is heap allocated (biggest part is the model which is a heap allocated vector anyways, so no large overhead).

regarding the replacement, there are multiple options:
1) the client drops the old model and allocates a new model, then the raw pointer is invalid and accessing the underlying memory will probably segfault.
2) the client overwrites the old model with a new model which has the same memory layout and location as the old one, then this wouldn't be possible in safe rust but the raw pointer still points to a valid memory location. of course the values at this memory are different which can be quite unexpected for the user and should be avoided.
3) the client mutates the old model into the new model, then this would only be possible in safe rust if the scopes of the mutable and immutable references are limited properly, but the raw pointer still points to a valid memory location. again this might be unexpected for the user if this wasn't called by the user.

currently the model memory is only changed when the user calls for it and the validity of the respective pointers is outlined in `# Workflow` step 4. we should keep surprises for the user as limited as possible, this resulted for example in the design choice to keep the pointer valid even after the current round has ended.

janpetschexain:
added a `new_model_N()` function which creates a zero-initialized cached primitive model. this can be called in case no global model is available from `get_model_N()`. also removed the `model` input from `update_model_N()`which means that the local model update can only be created from the cached model. this reduces possible bugs with models from foreign memory locations.

rust/src/sdk/api.rs
little-dude: I just asked about that, let's see: https://stackoverflow.com/questions/62422474/why-does-mut-t-implement-unwindsafe-but-mut-t-doesnt

little-dude: > UnwindSafe basically says that there is no danger in accessing a value after a panic happened. this is generally true for immutable references, but not for mutable references.

> currently i've put the note that it is the user's responsibility to not access possibly invalid values (we could clean up the client ourself but then the paradigm of create-use-destroy all happening on one side, in this case the non-Rust side of the API, is violated and might as well lead to severe bugs/segfaults).

> the main issue is that we must catch the panic, because letting it propagate across the FFI-boundary is undefined behavior and will most likely result in segfaults. what we can do is to improve error handling on our side to reduce the number of possible panics and return proper errors instead. (for example the first panic that is catched here is a panic due to closed channels if the client is started without a coordinator service. things like that should of course be proper errors, but in the end we can't rule out every possible panic and need to have a way to catch them.)

Thanks that clarifies everything. I think it would be nice to include these explanations in comments.

rust/src/sdk/api.rs
```diff
@@ -1,171 +1,243 @@
...
+///
+/// [errors]: #errors
+/// [`ParticipantInitErr`]: ../../client/enum.ClientError.html#variant.ParticipantInitErr
+/// [`ParticipantErr`]: ../../client/enum.ClientError.html#variant.ParticipantErr
+/// [`DeserialiseErr`]: ../../client/enum.ClientError.html#variant.DeserialiseErr
+/// [`GeneralErr`]: ../../client/enum.ClientError.html#variant.GeneralErr
+pub unsafe extern "C" fn run_client(
+    client: *mut FFIClient,
+    callback: unsafe extern "C" fn(*mut c_void, c_int),
+    state: *mut c_void,
```
little-dude, line 184: So in theory, users could store anything in there, including a `*mut FFIClient` or a `*PrimitiveModelXxx`, which in case of a panic would not be guaranteed to be valid. Is that correct?

rust/src/sdk/api.rs
```diff
@@ -174,235 +246,393 @@ pub unsafe extern "C" fn drop_client(client: *mut Client) {
...
-                if client.is_null()
-                    || model.ptr.is_null()
-                    || model.len == 0
-                    || model.len > (isize::MAX as usize / mem::size_of::<$prim>()) as c_ulong
-                {
-                    // TODO: add error handling
-                    panic!("invalid primitive model");
+        // deserialize and convert the global model to a primitive model and cache it
+        client.has_new_global_model_since_last_cache = false;
+        if let Ok(deserialized_model) = bincode::deserialize::<Model>(&global_model) {
```
little-dude, line 483: Unrelated to this PR of course, but this back and forth between model representations is very unfortunate... We need to address that properly at some point.

rust/src/sdk/api.rs
janpetschexain: i'll those as comments

rust/src/sdk/api.rs
```diff
@@ -1,171 +1,243 @@
...
+/// this memory.
+pub(crate) enum CachedModel {
     F32(Vec<f32>),
     F64(Vec<f64>),
     I32(Vec<i32>),
     I64(Vec<i64>),
 }
 
+#[derive(Debug)]
+/// A wrapper for a [`Client`] within a [`Runtime`].
```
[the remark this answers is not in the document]
janpetschexain: i'll replace the tokio links with an unlinked `asynchronous runtime` comment.

rust/src/sdk/api.rs
```diff
@@ -1,171 +1,243 @@
...
+}
+
 #[allow(unused_unsafe)]
 #[no_mangle]
-/// Creates a new [`Client`].
+/// Creates a new [`Client`] within a [`Runtime`].
 ///
 /// Takes a `period` in seconds for which the [`Client`] will try to poll the coordinator for new
-/// broadcasted FL round data.
+/// broadcasted FL round data. Specifies the `min_threads` and `max_threads` available to the
```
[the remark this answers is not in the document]
janpetschexain: i'm ok with that. in the test they are currently `min=1` and `max=4`, we can take those as defaults for now and have a performance evaluation once a real app runs this.

rust/src/sdk/api.rs
```diff
@@ -1,171 +1,243 @@
...
-    // TODO: actually start the client, requires tokio running
-
-    #[repr(C)]
-    struct Input {
-        _round_started: bool,
-        _participant_initialized: bool,
-        _model_cached: bool,
+#[allow(unused_unsafe)]
+#[no_mangle]
+/// Starts the [`Client`] and executes its tasks in an asynchronous [`Runtime`].
```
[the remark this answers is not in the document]
janpetschexain: i removed the tokio links. i would still mention the runtime since otherwise it might be unclear why the deconstructor acts like it does.

rust/src/sdk/api.rs
```diff
@@ -1,171 +1,243 @@
...
+/// - `2`: client stopped due to error [`ParticipantInitErr`]
+/// - `3`: client stopped due to error [`ParticipantErr`]
+/// - `4`: client stopped due to error [`DeserialiseErr`]
+/// - `5`: client stopped due to error [`GeneralErr`]
+///
+/// # Safety
+/// The method dereferences from the raw pointer arguments. Therefore, the behavior of
+/// the method is undefined if the arguments don't point to valid objects.
+///
+/// If the callback is triggered because of a panicking client, it is the users responsibility
```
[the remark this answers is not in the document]
janpetschexain: i went with `trigger` first to avoid the duplication of `call`, but the latter is more precise, i'll change that.

rust/src/sdk/api.rs
```diff
@@ -1,171 +1,243 @@
...
+/// to not access possibly invalid state and to drop the client. If certain parts of the state
```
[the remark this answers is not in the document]
janpetschexain: i think part of the ambiguity comes from the `UnwindSafe`ty: i removed the last sentences since it is probably quite impossible the check the `Client`'s invariants from the caller side after a panic (we might do that ourself in the future and then return an error code instead of a panic code). i rephrased the "state" to "client state" to better reflect the meaning.

we may also rename the callback's `state` to sth different to avoid ambiguity.

rust/src/sdk/api.rs
janpetschexain: that's correct. but since the `callback()` itself is defined somewhere on the caller side, everything happening in there is completely out of our control, no matter if the input arguments are valid or not.

the user might very well easily kill a client with an unsound callback, we can only make sure that a corrupted/killed client can't tear down for example the coordinator.

rust/src/sdk/api.rs
janpetschexain: i agree, it's a lot of allocation. nevertheless, at some point the user needs to have a model of primitive data type to actually train on it. but at least the serde part should be properly splitted from that.

rust/src/sdk/api.rs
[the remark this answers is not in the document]
little-dude: Sounds good to me.

rust/src/sdk/api.rs
little-dude: I'm wondering whether that could be exploited to create vulnerabilities. In doubt I'm tempted to ask: do we really need it? =/

janpetschexain: i don't have much experience with callback vulnerabilities, so not sure. i agree that it's probably better to be on the safe side: we can replace the callback by returning a simple error code for now and if we need callbacks in the future we can simply reintroduce them. i'll update this accordingly.

in the end the whole callback story is just the result of checking if it is feasible to have them in the API and the possibility that they might be needed from the app side.

PR #419, opened 2020-06-12
State machine

Robert-Steiner:
**Summary**
* Implement state machine pattern in coordinator

## State Machine

The state machine consists of the states `Idle`, `Sum`, `Update`, `Sum2` and `Error`.
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic3RhdGVEaWFncmFtICBcbiAgSWRsZSAtLT4gU3VtXG4gIFN1bSAtLT4gVXBkYXRlXG4gIFVwZGF0ZSAtLT4gU3VtMlxuICBTdW0yIC0tPiBJZGxlXG5cbiAgU3VtIC0tPiBFcnJvclxuICBVcGRhdGUgLS0-IEVycm9yXG4gIFN1bTIgLS0-IEVycm9yXG4gIEVycm9yIC0tPiBJZGxlXG5cblxuXG5cblxuXG4gICIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

## Tasks that are performed in the individual states

**Idle**
- clears the round dictionary
- updates the seed
 
**Sum**
- generates a new round key pair
- runs the sum phase (until the minimum number of sum massages is reached)
- freezes the sum dictionary

**Update**
- runs the update phase (until the minimum number of update massages is reached)

**Sum2**
- runs the sum2 phase (until the minimum number of mask is reached)
- performs the aggregation of masks and the unmasking of the final mask 

**Error**
- the transition into the `Error` state can only take place if all senders of the `request` channel have been dropped or the aggregation has failed 
- logs the error
- in case of a `ChannelError` the state machine `panic`s
---

**Example:**
You can find an example how to create and use the state machine in `src/rust/bin/state-machine.rs`

**Missing parts:**
- In the current implementation, the state machine does not emit any protocol events. This will be added in #414.
- I haven't added tests yet. I would like to add them in a follow up PR (after #414 is merged). 
- The implementation of phase timeouts, etc.
- [x] A clean shutdown of the state machine.


rust/src/state_machine/update.rs
```diff
@@ -0,0 +1,109 @@
...
+            }
+            Ok(())
+        } else {
+            Err(PetError::InvalidMessage)
+        }
+    }
+
+    /// Check whether enough update participants submitted their models and seeds to start the sum2
+    /// phase.
+    fn has_enough_seeds(&self) -> bool {
```
janpetschexain, line 101: this is probably due to my non-optimal naming choice: since we are waiting for enough seed dicts and updated local models, we should rename this method to sth like `has_enough_updates()` to also align it with the naming of `has_enough_sums()` from `State<Sum>`.

regarding the `has_enough_masks()` from `State<Sum2>` i'm undecided: we are actually only waiting for enough masks, but it would be more in line with `has_enough_sum2s()`. we might even call it `has_enough_sums()` as well (without the `2`) since this is waiting for the same sum participants as in `State<Sum>`.

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+pub enum StateError {
+    #[error("state failed: channel error: {0}")]
+    ChannelError(&'static str),
+    #[error("state failed: round error: {0}")]
+    RoundError(#[from] RoundFailed),
+}
+
+pub struct State<S> {
+    // Inner state
+    _inner: S,
```
janpetschexain, line 94: i wouldn't use the underscore naming here, since this indicates unused in rust instead of private like in some other languages.

Robert-Steiner:
Thanks for the feedback @janpetschexain 👍
I was busy fixing a bug in the state machine. (I forgot to call aggregator.aggregate in the update phase). I fixed it but now the aggregator is reinitialized twice (in Sum2 and idle state).
I'm not a big fan of my solution to be honest but I didn't want to change something on the aggregator in this pr.

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// Create a new state machine with the initial state `Idle`.
+    /// Fails if there is insufficient system entropy to generate secrets.
+    pub fn new() -> Result<(mpsc::UnboundedSender<Request>, Self), InitError> {
+        // crucial: init must be called before anything else in this module
+        sodiumoxide::init().or(Err(InitError))?;
+
+        let (request_tx, request_rx) = mpsc::unbounded_channel::<Request>();
+        let coordinator_state = CoordinatorState {
+            seed: RoundSeed::generate(),
+            ..Default::default()
```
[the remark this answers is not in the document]
Robert-Steiner: Nice, I didn't know that
I will change it

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+        debug!("received new message");
+        self.request_rx.recv().await.ok_or(StateError::ChannelError(
+            "all message senders have been dropped!",
+        ))
+    }
+
+    // Handle invalid requests.
+    fn handle_invalid_message(response_tx: oneshot::Sender<Result<(), PetError>>) {
+        debug!("invalid message");
+        let _ = response_tx.send(Err(PetError::InvalidMessage));
```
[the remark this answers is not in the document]
Robert-Steiner: It's the other way around. When I remove it, cargo throws a warning

```
warning: unused `std::result::Result` that must be used
   --> src/state_machine/mod.rs:120:9
    |
120 |         response_tx.send(Err(PetError::InvalidMessage));
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: 1 warning emitted
```

rust/src/state_machine/idle.rs
```diff
@@ -0,0 +1,62 @@
...
+        info!("state transition");
+        StateMachine::Idle(Self {
+            _inner: Idle,
+            coordinator_state,
+            request_rx,
+        })
+    }
+
+    pub async fn next(mut self) -> StateMachine {
+        self.clear_round_dicts();
```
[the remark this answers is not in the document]
Robert-Steiner: You are right, it would make more sense to clean it after the sum2 phase. 
The reason I did it this way was because I didn't want to add `clear_round_dicts` again in error state.

rust/src/state_machine/mod.rs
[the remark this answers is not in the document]
little-dude: Yes you do need `let _` to make it clear that you don't care about a `Result`, because it has the `#[must_use]` attribute.

rust/src/state_machine/idle.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+use sodiumoxide::crypto::hash::sha256;
+use tokio::sync::mpsc;
+
+#[derive(Debug)]
+pub struct Idle;
+
+impl State<Idle> {
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new(
```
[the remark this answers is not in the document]
Robert-Steiner: We can also return `Self` here but then I have to add the enum variant in the `next` method:
 
```diff
    pub async fn next(mut self) -> StateMachine {
-        State::<Sum>::new(self.coordinator_state, self.request_rx)
+        StateMachine::Sum(State::<Sum>::new(self.coordinator_state, self.request_rx))
    }
```

Another option could be to derive `From` from the `StateMachine` and call 
```
State::<Sum>::new(self.coordinator_state, self.request_rx).into()
```
in the `next` method.

But I also happy to use a different name.
wdyt @little-dude, @finiteprods ?

rust/src/state_machine/mod.rs
Robert-Steiner: I agree I will remove it

rust/src/state_machine/mod.rs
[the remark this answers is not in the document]
janpetschexain: ah i see, i didn't notice that the return type is a `Result<>`. then we should probably handle the result appropriately when error handling will be improved.

rust/src/state_machine/error.rs
```diff
@@ -0,0 +1,32 @@
+use super::{idle::Idle, CoordinatorState, Request, State, StateError, StateMachine};
+use tokio::sync::mpsc;
+
+#[derive(Debug)]
+pub struct Error {
```
little-dude, line 5: Why not use `StateError` directly here?

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// Functions that are available to all states
+impl<S> State<S> {
+    /// Receives the next [`Request`].
+    /// Returns [`StateError::ChannelError`] when all senders are dropped.
+    async fn next_request(&mut self) -> Result<Request, StateError> {
+        debug!("received new message");
+        self.request_rx.recv().await.ok_or(StateError::ChannelError(
+            "all message senders have been dropped!",
```
little-dude, line 113: I think this is more a shutdown condition than an error condition. Maybe we could have an additional `Shutdown` state instead, if that doesn't require too much changes to the PR?

rust/src/state_machine/mod.rs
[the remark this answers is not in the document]
little-dude: I think in this case it's fine to not handle the error. The only reason this would error out is if the receiver was dropped, which just means that whoever made the request was not interested in the response. This is probably worth a comment though.

rust/src/state_machine/error.rs
```diff
@@ -0,0 +1,32 @@
...
+            _inner: Error { error },
+            coordinator_state,
+            request_rx,
+        })
+    }
+
+    pub async fn next(self) -> StateMachine {
+        error!("state transition failed! error: {:?}", self._inner.error);
+        if let StateError::ChannelError(e) = self._inner.error {
+            panic!(e)
```
little-dude, line 27: What is the motivation for panicking? An alternative would be to return `self` directly.

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+
+#[derive(Error, Debug)]
+pub enum StateError {
+    #[error("state failed: channel error: {0}")]
+    ChannelError(&'static str),
+    #[error("state failed: round error: {0}")]
+    RoundError(#[from] RoundFailed),
+}
+
+pub struct State<S> {
```
little-dude, line 97: Maybe this could be named `StateMachine` and `CoordinatorState` could be name `State`?

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    // round parameters
+    sum: f64,
+    update: f64,
+    seed: RoundSeed,
+    min_sum: usize,
+    min_update: usize,
+
+    // round dictionaries
+    /// Dictionary built during the sum phase.
+    sum_dict: SumDict,
```
little-dude, line 40: The dictionaries are kind of specific to each phase, so I'm wondering if we should move them to the phase's state:

```rust
pub struct Sum {
    sum_dict: SumDict,
}

pub struct Update {
    frozen_sum_dict: SumDict,
    seed_dict: SeedDict,
}

pub struct Sum2Phase {
    frozen_sum_dict: SumDict,
    frozen_seed_dict: SeedDict,
    mask_dict: MaskDict,
}
```

If we have an aggregation phase in the future, after the sum2 phase, we could also move the `aggregation` attribute out of `CoordinatorState`.

rust/src/state_machine/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    // Handle invalid requests.
+    fn handle_invalid_message(response_tx: oneshot::Sender<Result<(), PetError>>) {
+        debug!("invalid message");
+        let _ = response_tx.send(Err(PetError::InvalidMessage));
+    }
+}
+
+pub enum StateMachine {
```
little-dude, line 124: Suggestion, maybe not for this PR:

- replacing the `StateMachine::Error(ChannelError)` by a shutdown state
- adding a `StateMachine::Aggregation` state.

rust/src/state_machine/mod.rs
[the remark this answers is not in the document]
Robert-Steiner: I have added a comment

rust/src/state_machine/mod.rs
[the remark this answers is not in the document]
Robert-Steiner: I noticed that I can't use `Self` here because `Self` is in this case `StateMachine`.

rust/src/state_machine/error.rs
Robert-Steiner: Good point😅. I replaced it.

rust/src/state_machine/mod.rs
janpetschexain: moving the dicts into state specific data is a good idea in general, but we have to be careful here, for example the `sum_dict` is modified in the `Sum` and `Sum2` state, but the `Update` state of course comes in between, so we would need a way to transfer it.

rust/src/state_machine/update.rs
Robert-Steiner: I renamed 
`has_enough_seeds` to `has_enough_updates()` and `has_enough_masks` to `has_enough_sums`

rust/src/state_machine/mod.rs
janpetschexain: having a specific `Aggregation` state should reflect the lazy aggregation strategy, meaning that the state machine would cycle back and forth from the `Sum2` and `Aggregation` state until the transitions criteria to `Idle` are satisfied.

rust/src/state_machine/mod.rs
janpetschexain: i agree that we should differentiate between recoverable error states and terminal shutdown states.

rust/src/state_machine/mod.rs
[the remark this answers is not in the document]
janpetschexain: ah that's true. `CoordinatorState::default()` should work nonetheless. it's a minor issue anyways, not worth to spend to much effort on it.

Robert-Steiner:
Alright, I tried to implement all your suggestions. 

What I have changed in the last commits:

- renamed `State` to `PhaseState`
- moved the `SumDict`, `SeedDict`, `MaskDict` and `Aggregator` into the respective `PhaseState`s 
	- This allowed me to remove `clear_round_dicts` and the double reinitializtion of the `Aggregator` in the `Sum2` and `Init` state  
- changed the return value of `PhaseState::<S>::new()` to `Self`
	- The main reason for this change is to simplify the testing of `PhaseState`. Before that change, I had to use `match` to access the inner value:

```
let state_machine = PhaseState<Idle>::new();
match state_machine {
	StateMachine::Idle(idle) => idle.method
}
```

- To keep the `next` method of `PhaseState` simple, I have added `#[derive(From)]` to `StateMachine`, so that I can write 

```diff
+ PhaseState::<Sum>::new(self.coordinator_state, self.request_rx).into()
// instead of
- StateMachine::Sum(PhaseState::<Sum>::new(self.coordinator_state, self.request_rx))
```

- added an `Unmask` state to separate the unmasking code from the `Sum2` state (maybe this separation is too much?)
- added a `Shutdown` state to shutdown the state machine in the event of unrecoverable errors or intended program stops.
- removed the `panic` for `ChannelError`
- in the `Shutdown` state a clean shutdown of the request channel is performed to consume all unprocessed messages. At the moment this does not make much sense, since in the case of a `ChannelError` the request channel is empty, but in the future we may have such a case where the state machine is in the shutdown state and there are still unprocessed messages in the channel.

I'm looking forward to hearing your feedback.

rust/src/state_machine/idle.rs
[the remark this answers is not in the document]
little-dude: I like both Jan's suggestion of using `from_xxx()` methods, and the idea of of returning `Self` and have `next()` return the enum, with the slight preference for the latter. The `From` impl is also a good idea.

rust/src/state_machine/error.rs
```diff
@@ -0,0 +1,37 @@
...
+            coordinator_state,
+            request_rx,
+        }
+    }
+
+    pub async fn next(self) -> Option<StateMachine> {
+        error!("state transition failed! error: {:?}", self.inner);
+        let next_state = match self.inner {
+            StateError::ChannelError(_) => {
+                PhaseState::<Shutdown>::new(self.coordinator_state, self.request_rx).into()
```
little-dude, line 30: nit: a `warn!("shutting down")` would be nice here

little-dude:
We should either squash or clean the history up a bit before merging.

Robert-Steiner:
Closed because it is part of #421.

PR #421, opened 2020-06-17
state machine + tower integration

little-dude:
refactor state machine and service

This commit introduces two major changes:

1. Use an asynchronous state machine
2. Split the service into multiple independant services implemented
   with `tower::Service`

# State Machine

The new state machine has been extracted from
https://github.com/xainag/xain-fl/pull/419.

It has the following phases:

- `Idle`: regenerate new round parameters and new keys, and broadcast
  them
- `Sum`: wait for enough sum participant to join and broadcast the
  resulting sum dictionary
- `Update`: aggregate the masked models from the update participants
  and broadcast the resulting seed dictionary
- `Sum2`: retrieve the masks computed by the sum participants
- `Unmask`: unmask the aggregated masked model and broadcast it
- `Error`: the state machine transition to this state whenever an
  un-expected error occurs

The "broadcasting" of events is realized via a small wrapper around
tokio watcher channels: `EventListener` and `EventBroadcaster`. The
state machine holds broadcasters, and anyone with a listener can
retrieve the latest events.

# Services

There are two main types of services:

- the services for fetching data broadcasted by the state machine:
  - `MaskLengthService`: for fetching the length of the model
  - `ModelService`: for fetching the last available global model
  - `RoundParametersService`: for fetching the current round parameters
  - `ScalarService`: for fetching the scalar used for aggregation
  - `SeedDictService`: for fetching the seed dictionary
  - `SumDictService`: for fetching the sum dictionary
- the services for handling PET messages from the participant:
  - `MessageParserService`: decrypt and parses incoming message
  - `PreProcessorService`: performs sanity checks on the messages
    (verify the task signatures, etc.)
  - `StateMachineService`: pass the messages down to the state machine
    for actual processing

The first type of services are gathered behind the `Fetcher` trait. A
`Fetcher` is a service that provides all the subservices listed
above. The second type of services are gathered behind the
`PetMessageHandler` trait.

There are three main motivations for splitting message parsing,
pre-processing, and actual processing:

- keep the state machine as simple as possible by keeping only the
  core PET logic
- make it possible to parallelize some stages. The message processing
  is ultimately sequential, because we need to update global
  dictionaries and models. But there is not reason not to parallelize
  the message parsing and pre-processing. Making them separate
  services makes this easier
- modularity: as the logic for each of these stages grow more complex,
  modularity will make the code more maintenable. For instance we'd
  like to support multipart messages. This logic could perhaps be just
  contained in the `MessageParserService`


rust/src/protocol/mod.rs
```diff
@@ -1,166 +1,8 @@
...
-
-        Ok((
-            request_tx,
-            PhaseState::<Idle>::new(coordinator_state, request_rx).into(),
-        ))
-    }
-}
+pub mod state_machine;
 
 #[cfg(test)]
```
little-dude, line 7: I guess I should move the tests to `src/protocol/state_machine.rs` now, correct @Robert-Steiner ?

Robert-Steiner: Yes, that would be nice.

Robert-Steiner:
The module layout looks good to 👍

rust/src/protocol/mod.rs
little-dude: done.

rust/src/services/messages/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+}
+
+#[rustfmt::skip]
+impl<MP, PP, SM> Service<Traced<MessageParserRequest>> for PetMessageService<MP, PP, SM>
+where
+    MP:
+        Service<Traced<MessageParserRequest>, Response = MessageParserResponse> + Clone + 'static + Send,
+        <MP as Service<Traced<MessageParserRequest>>>::Future: 'static + Send,
+        <MP as Service<Traced<MessageParserRequest>>>::Error: Into<Box<dyn ::std::error::Error + 'static + Sync + Send>>,
```
janpetschexain, line 268: i agree that turning off the `rustfmt` for those kind of heavy trait bounds is reasonable. but the indention is off one level which makes it quite hard to read. same for the two other impls below.

```suggestion
    MP: Service<Traced<MessageParserRequest>, Response = MessageParserResponse> + Clone + 'static + Send,
    <MP as Service<Traced<MessageParserRequest>>>::Future: 'static + Send,
    <MP as Service<Traced<MessageParserRequest>>>::Error: Into<Box<dyn ::std::error::Error + 'static + Sync + Send>>,
```

rust/src/services/fetchers/mask_length.rs
```diff
@@ -0,0 +1,35 @@
...
+pub struct MaskLengthService(EventListener<MaskLengthUpdate>);
+
+impl MaskLengthService {
+    pub fn new(events: &EventSubscriber) -> Self {
+        Self(events.mask_length_listener())
+    }
+}
+
+impl Service<MaskLengthRequest> for MaskLengthService {
+    type Response = Option<usize>;
```
janpetschexain, line 21: the aliased type may be used here directly. similarly for some of the other fetchers (some of them already use it).

```suggestion
    type Response = MaskLengthResponse;
```

rust/src/state_machine/events.rs
```diff
@@ -0,0 +1,271 @@
...
+
+impl EventPublisher {
+    /// Initialize a new event publisher with the given initial events.
+    pub fn init(
+        keys: KeyPair,
+        params: RoundParameters,
+        phase: PhaseEvent,
+    ) -> (Self, EventSubscriber) {
+        let round = params.id;
+        let (keys_tx, keys_rx) = watch::channel::<Event<KeyPair>>(Event { round, event: keys });
```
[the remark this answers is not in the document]
janpetschexain: i now noticed that even though the Fetcher (and indirectly the PetService) gets the EventSubscriber , it doesn’t expose any method to actually request the secret key from the service. i missed that yesterday, sorry for that!

it's probably helpful to write a comment reflecting this in the code.

rust/src/state_machine/coordinator.rs
```diff
@@ -0,0 +1,112 @@
...
+    crypto::{ByteObject, KeyPair},
+    mask::{MaskConfig, MaskObject},
+    state_machine::events::{EventPublisher, EventSubscriber},
+};
+
+pub type RoundId = u64;
+
+#[derive(Debug, Clone)]
+pub struct RoundParameters {
+    pub id: RoundId,
```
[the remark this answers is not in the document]
little-dude: ### Re: `RoundId

Agree. I'll remove it

### Re: not broadcasting all the round parameters in the Idle phase

With the latest changes, we now broadcast the round parameters in the Idle phase. There's is not risk of participants getting a head start, since all the parameters are  published at once.

rust/src/state_machine/events.rs
```diff
@@ -0,0 +1,271 @@
...
+#[derive(Debug, Clone)]
+pub struct Event<E> {
+    /// Metadata that associate this event to the round in which it is
+    /// emitted.
+    pub round: RoundId,
+    /// The event itself
+    pub event: E,
+}
+
+// FIXME: should we simply use `Option`s here?
```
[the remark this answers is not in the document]
little-dude: Ok Let's keep that for now then. Thanks for the input.

rust/src/state_machine/phases/sum.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        // the scalar must be published later for the participants
+        let scalar = 1_f64
+            / (self.coordinator_state.expected_participants as f64
+                * self.coordinator_state.round_params.update);
+        info!("broadcasting scalar: {}", scalar);
+        self.coordinator_state.events.broadcast_scalar(
+            self.coordinator_state.round_params.id,
+            ScalarUpdate::New(scalar),
+        );
```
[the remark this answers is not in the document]
little-dude: Sounds good.

rust/src/services/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        // allow processing 10 request concurrently, and allow up
+        // to 10 requests to be pending. Once 10 requests are
+        // being processed and 10 are queued, the service will
+        // report that it's not ready.
+        //
+        // FIXME: what's a good concurrency limit? Should we limit
+        // the number of concurrent messages being processed to
+        // the number of threads in the rayon thread-pool? Or is
+        // rayon smart enough to avoid too many context switches?
+        .buffer(100)
```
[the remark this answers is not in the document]
little-dude: `buffer(10)` means that the requests go through a queue with 10 slots before reaching our service. A request is dequeued and handed to the service when the service is ready to process a new request.

`concurrency_limit(10)` limits the number of requests that our service can process concurrently to 10. This means that if the _inner_ service is already handling 10 requests (ie there are 10 Futures returned by `Service::call` currently running), the service will wait for a future to finish before accepting a new request.

The two are a bit redundant, but slightly different, and the order in which they wrap the service matter. See: https://docs.rs/tower/0.3.1/tower/builder/struct.ServiceBuilder.html#order

rust/src/services/messages/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+        Into<Box<dyn ::std::error::Error + 'static + Sync + Send>>,
+{
+    async fn call_parser(&self, enc_message: Traced<Vec<u8>>) -> MessageParserResponse {
+        let span = enc_message.span().clone();
+        let mut svc = self.clone();
+        async move {
+            poll_fn(|cx| <Self as Service<Traced<MessageParserRequest>>>::poll_ready(&mut svc, cx))
+                .await
+                .map_err(Into::into)
+                .unwrap();
```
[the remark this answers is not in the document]
little-dude: Yeah. It's probably worth a comment. Ideally I shouldn't unwrap at all but I'd like to have a story for reworking the error handling anyway, where I'd like to go through every unwrap in the codebase.

rust/src/services/messages/mod.rs
little-dude: Ah ok. I thought grouping the bounds by type parameters made that more readable actually. I'll undo it if that's confusing.

rust/src/services/mod.rs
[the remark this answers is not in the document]
janpetschexain: thanks for the explanation and link. this means that the documentation is off and should read 100 instead of 10 because of the chosen service parameters.

rust/src/services/messages/mod.rs
janpetschexain: to me the layout reads that `MP` is a type with 3 trait bounds, but actually there are three trait bounds for three different types (although they are tightly coupled). not too much of a big deal though.

rust/src/client.rs
```diff
@@ -220,7 +225,7 @@ impl Client {
 
         debug!(client_id = %self.id, "sum message sent, polling for seed dict.");
         loop {
-            if let Some(seeds) = self.proxy.get_seeds(self.participant.pk).await? {
+            if let Some(seeds) = self.proxy.get_seeds(self.participant.pk.clone()).await? {
```
janpetschexain, line 228: the `pk` is `Copy`, no need to call `clone()` here.

rust/src/request.rs
```diff
@@ -1,30 +1,45 @@
...
 use reqwest::{Client, Error, IntoUrl, Response, StatusCode};
 
-#[derive(Debug)]
 /// Proxy for communicating with the service.
 pub enum Proxy {
-    InMem(Handle),
+    InMem(
+        Box<dyn Fetcher + Send + Sync>,
+        Box<dyn PetMessageHandler + Send + Sync>,
+    ),
```
janpetschexain, line 21: from looking at `rest.rs` where the traits are used as simple trait bounds, it seems that we may get rid of dynamic dispatch here as well. this is not urgent though, since the in-memory client is used for testing only.

rust/src/request.rs
```diff
@@ -83,79 +111,108 @@ impl Proxy {
         }
     }
 
+    pub async fn get_global_seeds(&self) -> Result<Option<SeedDict>, ClientError> {
```
janpetschexain, line 114: the naming `global_seeds` is quite unfortunate. the (local) seeds are used to generate the local masks, but there is no equivalent for the global mask (since the global mask is aggregated from the local masks and no seed is known for the global mask).

this is somewhat connected to the task PB-778. we may name it like `get_all_seeds` or sth similar, but in the end we should get the overall naming right in the near future (as a note to all of us).

little-dude: I'll go with `get_all_seeds` for now, thanks!

rust/src/request.rs
little-dude: Agree. This is a "quick & dirty" solution

rust/src/services/messages/mod.rs
little-dude: I removed the `rustfmt::skip` attribute, so it should look more familiar now

rust/src/request.rs
little-dude: Actually it's not even being used so I just removed it

rust/src/services/messages/mod.rs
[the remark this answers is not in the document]
little-dude: added comments

rust/src/state_machine/phases/sum.rs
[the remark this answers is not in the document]
little-dude: done

rust/src/services/mod.rs
```diff
@@ -0,0 +1,179 @@
+mod fetchers;
+mod messages;
+
+pub use self::{
+    fetchers::{FetchError, Fetcher},
+    messages::{PetMessageError, PetMessageHandler},
+};
```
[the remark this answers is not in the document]
little-dude: This is a re-export, and the `use` statement below is not `pub`. That's why it's separated from the rest.

rust/src/services/mod.rs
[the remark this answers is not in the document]
little-dude: fixed.

rust/src/state_machine/phases/update.rs
```diff
@@ -0,0 +1,183 @@
...
+        // See `Self::handle_invalid_message`
+        let _ = response_tx.send(self.update_seed_dict_and_aggregate_mask(
+            &participant_pk,
+            &local_seed_dict,
+            masked_model,
+        ));
+    }
+
+    /// Updates the local seed dict and aggregates the masked model.
+    fn update_seed_dict_and_aggregate_mask(
```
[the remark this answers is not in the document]
little-dude: Good point. I changed the order of operations as suggested and added a comment.

rust/src/state_machine/events.rs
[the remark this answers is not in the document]
little-dude: I added a doc comment for now, what do you think?

rust/src/services/messages/message_parser.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let handler = Handler {
+            keys: keys_ev.event,
+            phase: phase_ev.event,
+        };
+
+        let (tx, rx) = oneshot::channel::<Self::Response>();
+
+        trace!("spawning pre-processor handler on thread-pool");
+        self.thread_pool.spawn(move || {
```
Robert-Steiner, line 114: Using `rayon` here is really cool. I also had in mind that we could use `tokio::spawn_blocking` but then I saw that they say in the tokio documentation:

> If your code is CPU-bound and you wish to limit the number of threads used to run it, you should run it on another thread pool such as rayon. You can use an oneshot channel to send the result back to Tokio when the rayon task finishes.

So `rayon` is the way to go. However, I don't really know how `tokio` and `rayon` split up the cpu cores among each other. So the issue I see here is that `tokio` and `rayon` run on the same core and interfere each other.

rust/src/services/messages/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+fn with_tracing<S, R>(service: S) -> TracingService<S, R>
+where
+    S: Service<Traced<R>>,
+{
+    ServiceBuilder::new()
+        .layer(request_span::layer(req_span as for<'r> fn(&'r _) -> _))
+        .service(service)
+}
+
+type TracingService<S, R> = request_span::Service<S, Traced<R>, fn(&Traced<R>) -> tracing::Span>;
```
Robert-Steiner, line 65: I don't get what that means to be honest😅
Is it an alias for a function?

rust/src/state_machine/coordinator.rs
```diff
@@ -0,0 +1,107 @@
...
+        Self::from_slice_unchecked(randombytes(Self::LENGTH).as_slice())
+    }
+}
+
+/// A dictionary created during the sum2 phase of the protocol. It counts the model masks
+/// represented by their hashes.
+pub type MaskDict = HashMap<MaskObject, usize>;
+
+#[derive(Debug, Clone, Copy)]
+pub enum PhaseEvent {
```
Robert-Steiner, line 99: `PhaseEvent` could also be moved to `events.rs`

rust/src/services/messages/mod.rs
little-dude: It's a normal type alias. The not so usual part of this type is `fn(&Traced<R>) -> tracing::Span`. It's the type of a function that takes a `&Traced<R>` and returns a `Span`. This is not to be confused with the `Fn`, `FnOnce` and `FnMut` _traits_ (not _types_!), which are implemented by functions.

little-dude: See: https://doc.rust-lang.org/reference/types/function-item.html

rust/src/services/messages/message_parser.rs
little-dude: Good question. I don't know =/

rust/src/services/messages/mod.rs
janpetschexain: it just looks a bit unusual here, but what rust calls a *function item type* is equivalent to a simple *function pointer* in other languages.

little-dude: Yeah rust also has function pointers. The difference is somewhat subtle: https://stackoverflow.com/a/54629954/1836144

rust/src/services/mod.rs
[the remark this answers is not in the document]
janpetschexain: ahh true, i missed the `pub`.

rust/src/services/messages/message_parser.rs
Robert-Steiner: I can do some research :)
Already found something https://docs.rs/tokio/0.2.21/tokio/#cpu-bound-tasks-and-blocking-code

rust/src/bin/main.rs
```diff
@@ -10,27 +16,33 @@ struct Opt {
...
-    let settings = Settings::new(opt.config_path).unwrap_or_else(|err| {
-        eprintln!("{}", err);
-        process::exit(1);
-    });
-
     let _fmt_subscriber = FmtSubscriber::builder()
         .with_env_filter(EnvFilter::from_default_env())
         .with_ansi(true)
         .init();
+    sodiumoxide::init().unwrap();
```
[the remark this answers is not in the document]
little-dude: added a comment about that. I could also just remove it but it doesn't hurt and I'd rather be on the safe side here.

rust/src/state_machine/coordinator.rs
little-dude: Good point. I moved it

rust/src/services/messages/pre_processor/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+        coordinator::RoundParameters,
+        events::{Event, EventListener, EventSubscriber, PhaseEvent},
+    },
+    utils::trace::{Traceable, Traced},
+};
+
+/// Route the request to the service that is ready to process it,
+/// depending on the current coordinator phase.
+pub struct PreProcessorService {
+    params_listener: EventListener<RoundParameters>,
```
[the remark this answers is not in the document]
little-dude: I have a PR for updating the doc, I probably added it there.

rust/src/services/fetchers/round_parameters.rs
```diff
@@ -0,0 +1,35 @@
...
+use futures::future::{self, Ready};
+use tower::Service;
+
+use crate::state_machine::{
+    coordinator::RoundParameters,
+    events::{EventListener, EventSubscriber},
+};
+
+pub struct RoundParamsRequest;
+pub type RoundParamsResponse = RoundParameters;
```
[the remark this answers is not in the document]
little-dude: Exactly

rust/src/services/messages/pre_processor/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+            PhaseEvent::Update => self.update.poll_ready(cx),
+            PhaseEvent::Sum2 => self.sum2.poll_ready(cx),
+            _ => Poll::Ready(Ok(())),
+        }
+    }
+
+    fn call(&mut self, req: Traced<PreProcessorRequest>) -> Self::Future {
+        let MessageOwned { header, payload } = req.into_inner().0;
+        // `call()` is only called after `poll_ready()` returned
+        // `Poll::Ready` so at this point, phase is `Some`
```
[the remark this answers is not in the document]
little-dude: Outdated comment indeed.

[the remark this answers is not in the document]
little-dude: Removed it!

rust/src/services/fetchers/mod.rs
```diff
@@ -0,0 +1,293 @@
...
+
+use std::task::{Context, Poll};
+
+use futures::future::poll_fn;
+use tower::Service;
+
+use crate::state_machine::coordinator::RoundParameters;
+
+#[async_trait]
+pub trait Fetcher {
```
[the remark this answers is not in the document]
little-dude: Yeah I'm not 100% sure we actually need that trait. For now it's convenient but maybe I'll just get rid of it at some point.

PR #422, opened 2020-06-19
PB-806 Update Dockerfile and docker-compose

rsaffi:


docker/Dockerfile
```diff
@@ -6,20 +9,22 @@ COPY docker/install_rust.sh .
 RUN bash ./install_rust.sh
 
 # Install dev tools
-RUN apt-get update && apt-get install -y valgrind
```
rsaffi, line 9: Removed the `apt update` and that is already executed inside the `install_rust.sh` script.

docker/Dockerfile
```diff
@@ -1,25 +1,32 @@
+# https://github.com/rust-lang/docker-rust-nightly/blob/master/buster/Dockerfile
+FROM buildpack-deps:stable-curl AS builder
+
 # Install Rust nightly
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH
+RUN apt update
```
rsaffi, line 8: Making apt update more explicit from the `Dockerfile` rather than inside custom script `install_rust.sh`.

docker/install_rust.sh
```diff
@@ -2,17 +2,16 @@
...
-apt-get update
-apt-get install -y --no-install-recommends \
+apt install -y --no-install-recommends \
         ca-certificates \
         gcc \
         libc6-dev \
         wget \
         libssl-dev \
         pkg-config \
+        build-essential
```
rsaffi, line 12: Added this (for now, at least), to bypass this:
```
(...)
   Compiling tokio v0.2.20
   Compiling futures-util v0.3.4
   Compiling thiserror v1.0.16
   Compiling tracing v0.1.13
   Compiling futures-executor v0.3.4
   Compiling futures v0.3.4
error: failed to run custom build command for `libsodium-sys v0.2.5`

Caused by:
  process didn't exit successfully: `/target/debug/build/libsodium-sys-2c273b8b56cf939d/build-script-build` (exit code: 101)
```

docker/.dev.env
```diff
@@ -0,0 +1,2 @@
+MINIO_ACCESS_KEY=minio
+MINIO_SECRET_KEY=minio123
```
rsaffi, line 2: I decided to go with an env file that has credentials on it **just** because this is for DEV and scope is limited to a developer's dev environment. Having the credentials directly on the `docker-compose.yml` meant having to input them twice: once for `minio` container and again for `setup-initial-bucket` container. This way, we only have them defined once and then referenced by both containers.

rsaffi:
This PR does not contemplate the changes that will occur to the config files for the coordinator, as that is being done on a separate branch. Whichever branch is merged last into `pet` will need to account for that.

docker/install_rust.sh
little-dude: seems legit.

docker/Dockerfile
```diff
@@ -1,25 +1,32 @@
...
 # Now copy the actual source code
-COPY rust/src src
+COPY rust/src /src
 # Fix timestamp. cargo incremental build as issues caused by the above
 # trick to compile the deps separately
-RUN touch src/lib.rs
+RUN touch /src/lib.rs
 RUN cargo build --all-features
+
+FROM alpine:latest
```
little-dude, line 31: Is this gonna work? alpine is based on musl, while the builder is based on glibc.

docker/Dockerfile.release
```diff
@@ -1,11 +1,22 @@
...
+COPY rust/Cargo.lock /
+COPY rust/Cargo.toml /
+# Now copy the actual source code
+COPY rust/src /src
+# Fix timestamp. cargo incremental build as issues caused by the above
+# trick to compile the deps separately
+RUN touch /src/lib.rs
 RUN cargo build --release --features=telemetry,influx_metrics
+
+FROM alpine:latest
```
little-dude, line 21: Same as above, I don't think this will work

docker/Dockerfile
rsaffi: I went with the lightest ones for the sake of keeping the environment as light as possible.
The binary that the rust coordinator outputs is dynamically linked or statically linked? If it's static then it _shouldn't_ be a problem, right?

I can also change the `alpine` for another debian-based image that keeps everything light, no problem.

docker/docker-compose.yml
```diff
@@ -1,31 +1,94 @@
...
+      - .dev.env
     command: server /data
+    volumes:
+      - minio-data:/data
+    networks:
+      - xain-fl
     ports:
       - "9000:9000"
 
+  setup-initial-buckets:
```
Robert-Steiner, line 50: I'm not sure if we still need this part. In the current S3 PR, the coordinator is responsible to create the buckets if they don't exist. IMO we can remove this for now and add it back if we need to.

docker/docker-compose.yml
```diff
@@ -1,31 +1,94 @@
...
     hostname: redis
+    container_name: redis
+    entrypoint: /usr/local/bin/redis-server --appendonly yes --appendfsync everysec # using combination of RDB and AOF for persistence: https://redis.io/topics/persistence
     volumes:
       - redis-data:/data
     networks:
       - xain-fl
-    ports:
-      - "6379:6379"
+    expose:
```
Robert-Steiner, line 85: I would like to keep `ports` here because sometimes I need to access Redis via the redis-cli.

docker/Dockerfile
little-dude: Rust binaries dynamically link against the system's libc library.
There is a way to create binaries statically linked with musl, so these binaries can run on any Linux, but afaik that's not the default.

little-dude: See: https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.html

little-dude: I'm not against producing musl binaries for docker though, so feel free to go that direction if you have a preference for alpine based containers.

docker/docker-compose-valgrind.yml
```diff
@@ -1,34 +1,36 @@
-version: "3.7"
+versioe: "3.8"
```
Robert-Steiner, line 1: @little-dude Do we still want to keep the `valgrind` stuff?

little-dude: Probably not. But it doesn't hurt so I didn't comment about it.

docker/docker-compose.yml
rsaffi: I had changed this when we were running the dev `docker-compose` stack on DigitalOcean for the first testing sessions and we ended up with an inadvertently exposed Redis.
Can revert back to `ports`, as this docker-compose shouldn't be used in production anyway.

docker/docker-compose-valgrind.yml
rsaffi: If not, please do let me know, otherwise I just get rid of it.

little-dude: Let's get rid of it then.

docker/Dockerfile
rsaffi: Ok, I will weight pros and cons of either going with non-alpine image vs. alpine with statically linked binary.

docker/docker-compose-valgrind.yml
rsaffi: Ok, will get rid of it, then!

rsaffi:
PR was updated with all the suggested changes!

PR #423, opened 2020-06-22
Make coordinator configurable

Robert-Steiner:
**Summary**

- added a module for loading and validating coordinator settings
- updated existing coordinator settings and removed obsolete aggregator configuration files  

**Ways to configure the coordinator**

Settings can be defined via a configuration file or via environment variables, whereby environment variables always overwrite the settings defined in the configuration file.

For the configuration files the following file formats are supported: JSON, YAML, TOML and HJSON.

Environment variables have the format `XAIN_{SETTINGS_NAME}` with `__` used as separator between settings names. 

For example the environment variable of 

```toml
[api]
bind_address = "0.0.0.0:1234"
```

is:

`XAIN_API__BIND_ADDRESS=0.0.0.0:1234`

**Validation of coordinator settings**

For the validation of the setting values I used the crate `validator`. It contains macros that can be used to add validation to struct fields. It comes with some built-in validators which makes it easy to use. 

However, the error message is by no means easy to read.

```
    ValidationErrors(
        {
            "api": Struct(
                ValidationErrors(
                    {
                        "bind_address": Field(
                            [
                                ValidationError {
                                    code: "url",
                                    message: None,
                                    params: {
                                        "value": String(
                                            ":1233",
                                        ),
                                    },
                                },
                            ],
                        ),
                    },
                ),
            ),
        },
    ),
```

If we want to use it, we should consider writing a formatter that prints this error in a readable format.


rust/src/settings.rs
```diff
@@ -0,0 +1,184 @@
...
+
+#[derive(Debug, Validate, Deserialize)]
+/// PET protocol settings
+pub struct PetSettings {
+    #[validate(range(min = 1))]
+    pub min_sum: usize,
+    #[validate(range(min = 3))]
+    pub min_update: usize,
+
+    #[validate(range(min = 0, max = 1.0))]
```
janpetschexain, line 35: do you know, for example from testing, if the bounds on `range` are inclusive or exclusive or partly, meaning `min <= value <= max` or `min < value < max` or `min <= value < max`? the documentation doesn't say anything about this. if it were exclusive, we could already rule out corner cases.

especially for the integer values of the `PetSettings` this is quite important.

rust/src/settings.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Fails when the loading of the configuration file or its validation failed.
+    pub fn new(path: &str) -> Result<Self, SettingsError> {
+        let settings: Settings = Self::load(path)?;
+        settings.validate()?;
+        Ok(settings)
+    }
+
+    /// Loads the coordinator settings via a configuration file without validation.
+    /// Fails when the loading of configuration file failed.
+    pub fn new_dirty(path: &str) -> Result<Self, ConfigError> {
```
janpetschexain, line 174: in other parts of the code we use the `_unchecked()` suffix, but that usually implies a return value and not a return result/option. maybe `new_unvalidated()` is a more telling name?

rust/src/settings.rs
janpetschexain: i had a look into the `validator` source code (i should have done this right away^^): the bounds are inclusive, i.e. it is always checked that `min <= value <= max`. also the data type doesn't matter, everything is treated as `f64`.

btw the same holds true for the bounds of the `length` check. this makes sense in the validation case, but it is quite counter intuitive from the usual behavior of `Range<_>`.

rust/src/settings.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[validate]
+    pub api: ApiSettings,
+    #[validate]
+    pub pet: PetSettings,
+    pub mask: MaskSettings,
+}
+
+#[derive(Debug, Validate, Deserialize, Clone, Copy)]
+/// PET protocol settings
+pub struct PetSettings {
```
janpetschexain, line 30: what do you think of adding an additional struct-level validation for `PetSettings`? it would allow checking for the corner cases mentioned last week in the slack channel.

it should be enough to add this attribute to the `PetSettings`:
```
#[validate(schema(function = "validate_fractions"))]
```
where the function `"validate_fractions"` will be
```rust
fn validate_fractions(s: PetSettings) -> Result<(), ValidationError> {
    if 0. < s.sum
        && s.sum < 1.
        && 0. < s.update
        && s.update < 1.
        && 0 < s.sum + s.update - s.sum*s.update
        && s.sum + s.update - s.sum*s.update < 1
    {
        Ok(())
    } else {
        Err(ValidationError::new("starvation"))
    }
}
```
(i'm not sure on the input arguments, the documentation is quite lacking.)

rust/src/settings.rs
Robert-Steiner: I changed it to `new_unchecked`. It sounds better 😁

Robert-Steiner:
Changes since the last commits

- impl `From MaskSettings` for `MaskConfig`
- impl `Default` for `MaskSettings` and `PetSettings`
- added a `-c` flag to the coordinator binary to specify the path of the configuration file
- integrated the `Settings` struct into the `Coordinator` struct
- updated test-drives and docker-compose files

rust/src/settings.rs
Robert-Steiner: Good point👍

I will add this

Robert-Steiner: Seems to work!
If we add this validation we can actually remove the validation `#[validate(range(min = 0, max = 1.0))]` on the `sum` and `update` fields right?

janpetschexain: sure, the validation on the fields is now subsumed by the validation of the struct. we should still document the constraints somewhere.

rust/src/settings.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+
+#[derive(Debug, Validate, Deserialize, Clone, Copy)]
+#[doc(hidden)]
+pub struct Settings {
+    #[validate]
+    pub api: ApiSettings,
+    #[validate]
+    pub pet: PetSettings,
+    pub mask: MaskSettings,
```
finiteprods, line 25: i'm not sure why this one is not `#[validate]`d?

rust/src/settings.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Fails when the loading of the configuration file or its validation failed.
+    pub fn new(path: PathBuf) -> Result<Self, SettingsError> {
+        let settings: Settings = Self::load(path)?;
+        settings.validate()?;
+        Ok(settings)
+    }
+
+    /// Loads the coordinator settings via a configuration file without validation.
+    /// Fails when the loading of configuration file failed.
+    pub fn new_unchecked(path: PathBuf) -> Result<Self, ConfigError> {
```
finiteprods, line 209: do we need an unchecked `new`? just wondering

rust/src/bin/test-drive-net.rs
```diff
@@ -32,7 +25,7 @@ async fn main() -> Result<(), ClientError> {
 
     let mut tasks = vec![];
     for id in 0..10 {
-        let mut client = Client::new_with_addr(1, id, "http://127.0.0.1:3030")?;
+        let mut client = Client::new_with_addr(1, id, "http://127.0.0.1:8081")?;
```
finiteprods, line 28: we already spoke, just again for reference: would be good to "config" this as well, in a follow-up.

rust/src/settings.rs
Robert-Steiner: For `MaskSettings` we don't need to add  `#[validate]` because of the property of `config` for enums. `config` only allows valid enum variances. So `config` already performs the validation. 
If you try to use an invalid value (for example `group_type="None"`), the coordinator will fail with the error: `configuration loading failed: enum GroupType does not have variant constructor None`.

rust/src/settings.rs
Robert-Steiner: Actually that's true. I will remove it

rust/src/bin/test-drive-net.rs
Robert-Steiner: Thanks a lot for the review👍 
I created a ticket (PB-837)

PR #425, opened 2020-06-23
PB-822 Updating coordinator config file on Kubernetes

rsaffi:


rsaffi:
:warning: Keep in mind that I branched from `master` and this PR is for merging back into `master`.

Come to think of it, given there's a task number associated with this change (PB-822), maybe I'll update the PR just with the update to the configuration file and keep all the other changes to an update/cleanup of the `k8s` folder.

WDYT @dkravetz?

k8s/coordinator/development/config-release-coordinator.toml
```diff
@@ -1,21 +1,14 @@
...
-[federated_learning]
-rounds = 100
-participants_ratio = 0.8
-min_clients = 100
-heartbeat_timeout = 5
-
-[metric_store]
-database_url = "http://influxdb-development:8086"
-database_name = "metrics"
+[pet]
```
dkravetz, line 4: Maybe the file name can be updated to reflect the new values? perhaps something like "config.toml" (it is under `coordinator/` after all)

rsaffi: Good point!

rsaffi:
Changes applied as suggested!

PR #426, opened 2020-06-23
PB-794 fix mask length

janpetschexain:
**References**
- [PB-794](https://xainag.atlassian.net/browse/PB-794)

**Summary**
- add a model/mask length GET request to the rest api
- handle the request by the in-memory handle and the network handle
- adjust the coordinator, participant and client to use the model/mask length information in the sum2 phase

### DIFF
--- a/rust/src/rest.rs
+++ b/rust/src/rest.rs
@@ -60,21 +60,12 @@ async fn handle_message(body: Bytes, handle: Handle) -> Result<impl warp::Reply,
 
 async fn handle_sums(handle: Handle) -> Result<impl warp::Reply, Infallible> {
     let sums = handle.get_sum_dict().await;
-    Ok(build_response(sums))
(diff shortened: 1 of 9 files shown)


rust/src/rest.rs
```diff
@@ -29,19 +29,26 @@ pub async fn serve(addr: impl Into<SocketAddr> + 'static, handle: Handle) {
         .and(warp::get())
         .and(part_pk())
         .and(with_hdl(handle.clone()))
-        .and_then(handle_seeds)
-        .recover(handle_reject);
+        .and_then(handle_seeds);
+        // .recover(handle_reject);
```
finiteprods, line 33: feel free to remove this comment

janpetschexain: thanks for catching this! seems that i forgot about it.

PR #429, opened 2020-06-24
PB-840 switch to stable rust

janpetschexain:
**References**
- [PB-840](https://xainag.atlassian.net/browse/PB-840)

**Summary**
- rewrite the two occurrences of unstable feature usage
- set the rust toolchain, github ci and dockerfiles to rust stable

As soon as this is merged, please make sure that you are using stable rust as well, otherwise it might cause conflicts when pushing merge requests to the CI. Therefore please check that:
- the `RUSTUP_TOOLCHAIN` env variable is either not set or set to stable
- `rustup override list` doesn't show any nightly overrides for the `xain_fl` directory (can for example be wiped via `rustup override unset` and `rustup override unset --nonexistent`)
- either the default toolchain is set to stable or the toolchain for the `xain_fl` directory is set to stable (can for example be set via `rustup default stable` and checked via `rustup toolchain list`)
- the active toolchain for the `xain_fl` directory is stable (can be checked via `rustup show`)

All `rustfmt` features that we are currently using require nightly. since code formatting doesn't affect stability of the code at all, i'm strongly arguing in favor of using nightly formatting. this doesn't affect the rest of the toolchain, which can of course stay stable. When formatting the code locally please check that:
- `rustfmt` is called as nightly from the CLI like `cargo +nightly fmt`
- `rustfmt` is called with the additional argument `+nightly` from an IDE like VSCode (`"rust-analyzer.rustfmt.extraArgs": ["+nightly"],`)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,7 +34,7 @@ jobs:
 
       - name: cargo fmt
         working-directory: ./rust
-        run: cargo fmt --all -- --check
+        run: cargo +nightly fmt --all -- --check
 
   check:
     name: rust-build
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,11 +9,11 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: Install nightly
+      - name: Install stable
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
           components: rustfmt
 
       - name: Cache cargo registry
@@ -46,11 +46,11 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: Install nightly
+      - name: Install stable
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
 
       - name: Cache cargo registry
         uses: actions/cache@v1
@@ -84,11 +84,11 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: Install nightly
+      - name: Install stable
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
           components: clippy
 
       - name: Cache cargo registry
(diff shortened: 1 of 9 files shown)


.github/workflows/rust.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
       - name: cargo fmt
         working-directory: ./rust
-        run: cargo fmt --all -- --check
+        run: cargo +nightly fmt --all -- --check
```
Robert-Steiner, line 37: Is this required? Because we install the `nightly` toolchain in L12

rust/rustfmt.toml
```diff
@@ -1,3 +1,4 @@
+# requires unstable rustfmt until the options are stabilized
```
Robert-Steiner, line 1: Have you tried the option `unstable_features=true`? [unstable_features](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#unstable_features) Maybe it allows us to run fmt via `cargo fmt` instead of `cargo +nightly fmt`.

.github/workflows/rust.yml
janpetschexain: for some reason this is required. i tried it without `+nightly` and the formatting was done like on stable even though the nightly toolchain is installed for this task. i have no idea why, maybe it is some caching issue since the other tasks are installing the stable toolchain.

rust/rustfmt.toml
janpetschexain: i've tried that, but it doesn't work. i found an open issue on the `rustfmt` github which says that this option itself is not stable and therefore doesn't have an effect for stable toolchains.

.github/workflows/rust.yml
Robert-Steiner: Ah true. By default cargo uses stable. The reason why it worked before is that it was overwritten by then `rust-toolchain` file.

rust/rustfmt.toml
Robert-Steiner: Alright 🙈

janpetschexain:
> The `Dockerfile.release` build failed but not because of your pr. I will discuss this with @rsaffi

ok, thanks for checking!

PR #430, opened 2020-06-24
Fix dockerfile

Robert-Steiner:
Summary:
- fixes the error: "coordinator_1  | /app/coordinator: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory"
- removed old rpc port
- removed old prometheus files


docker/docker-compose-release.yml
```diff
@@ -7,11 +7,10 @@ services:
       dockerfile: docker/Dockerfile.release
     command: /app/coordinator -c /app/config.toml
     volumes:
-      - ${PWD}/../configs/docker-release.toml:/app/config.toml
```
Robert-Steiner, line 10: I'm happy to revert that change. We just have to make sure that we document in which folder the user must be in order to start docker-compose. Normally I start docker-compose in the root of the repo via the command: `docker-compose -f docker/docker-compose.yml up`

docker/docker-compose.yml
```diff
@@ -11,11 +11,12 @@ services:
       - influxdb
     command: /app/coordinator -c /app/config.toml
     volumes:
-      - ${PWD}/../configs/docker-dev.toml:/app/config.toml
+      - ${PWD}/configs/docker-dev.toml:/app/config.toml
     networks:
       - xain-fl
+    environment:
```
Robert-Steiner, line 17: This is something that we need to update on the kubernetes setup as well, otherwise only errors are logged.

rsaffi: Will do so!

rsaffi: Done!

PR #431, opened 2020-06-25
PB-875 document crypto module

janpetschexain:
**References**
- [PB-875](https://xainag.atlassian.net/browse/PB-875)

**Summary**
- add module level documentation
- update struct and method documentation
- unravel import indirections
- consistify naming and usage of the newly introduced `KeyPair`

note that this requires `cargo +nightly doc` to generate proper internal links, which is not an issue since this is supported by the hosted rust docs as well.

### DIFF
--- a/rust/src/crypto/encrypt.rs
+++ b/rust/src/crypto/encrypt.rs
@@ -13,7 +13,7 @@ use super::ByteObject;
 /// Number of additional bytes in a ciphertext compared to the corresponding plaintext.
 pub const SEALBYTES: usize = sealedbox::SEALBYTES;
 
-/// Generates a new random key pair for encryption.
+/// Generates a new random `C25519` key pair for encryption.
 pub fn generate_encrypt_key_pair() -> (PublicEncryptKey, SecretEncryptKey) {
     let (pk, sk) = box_::gen_keypair();
     (PublicEncryptKey(pk), SecretEncryptKey(sk))
@@ -47,7 +47,7 @@ impl KeyPair {
     PartialOrd,
     Debug,
 )]
-/// A public key for asymmetric authenticated encryption.
+/// A `C25519` public key for asymmetric authenticated encryption.
 pub struct PublicEncryptKey(box_::PublicKey);
 
 impl ByteObject for PublicEncryptKey {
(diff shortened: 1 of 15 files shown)


rust/src/crypto/encrypt.rs
```diff
@@ -1,32 +1,38 @@
...
-    let (pk, sk) = box_::gen_keypair();
-    (PublicEncryptKey(pk), SecretEncryptKey(sk))
-}
+/// Number of additional bytes in a ciphertext compared to the corresponding plaintext.
+pub const SEALBYTES: usize = sealedbox::SEALBYTES;
 
 #[derive(Debug, Clone)]
-pub struct KeyPair {
+/// A `C25519` key pair for asymmetric authenticated encryption.
+pub struct EncryptKeyPair {
```
Robert-Steiner, line 18: I like the renaming.

rust/src/crypto/encrypt.rs
```diff
@@ -59,37 +66,40 @@ impl ByteObject for PublicEncryptKey {
...
 
-    /// Encrypt a message `m` with this public key. The resulting
-    /// ciphertext length is [`SEALBYTES`] + `m.len()`.
+    /// Encrypts a message `m` with this public key. The resulting ciphertext length is
+    /// [`SEALBYTES`]` + m.len()`.
     ///
-    /// The function creates a new key pair for each message, and
-    /// attaches the public key to the ciphertext. The secret key is
-    /// overwritten and is not accessible after this function returns.
+    /// The function creates a new ephemeral key pair for each message, and attaches the ephemeral
```
Robert-Steiner, line 75: Does `m` contains more than one message? If not I would change it to:
`The function creates a new ephemeral key pair for a message`

rust/src/crypto/encrypt.rs
```diff
@@ -1,32 +1,38 @@
-use super::ByteObject;
+//! Wrappers around some of the [sodiumoxide] encrytion primitives.
```
Robert-Steiner, line 1:
```suggestion
//! Wrappers around some of the [sodiumoxide] encryption primitives.
```

rust/src/crypto/encrypt.rs
janpetschexain: i'll change it. it's one message per returned `Vec<u8>`.

rust/src/crypto/encrypt.rs
janpetschexain: changed.

rust/src/crypto/encrypt.rs
finiteprods: yeah i agree the "each message" part is a bit confusing. you could even say "**the** message", or just `m`

rust/src/crypto/sign.rs
```diff
@@ -124,11 +146,11 @@ impl ByteObject for Signature {
 }
 
 impl Signature {
-    /// Length in bytes of a [`Signature`]
+    /// Length in bytes of this signature.
     pub const LENGTH: usize = sign::SIGNATUREBYTES;
 
-    /// Compute the floating point representation of the hashed
-    /// signature and ensure that it is below the given threshold:
+    /// Computes the floating point representation of the hashed signature and ensure that it is
```
finiteprods, line 152:
```suggestion
    /// Computes the floating point representation of the hashed signature and ensures that it is
```

janpetschexain: i'll fix this typo in one of the other doc rebases i'm currently on so it won't get lost.

PR #434, opened 2020-06-25
Add logging settings

Robert-Steiner:
Summary:
- adds the possibility to define `logging`in the config file

### DIFF
--- a/configs/config.toml
+++ b/configs/config.toml
@@ -1,3 +1,6 @@
+[log]
+filter = "xain_fl=debug,info"
+
 [api]
 bind_address = "127.0.0.1:8081"
 
--- a/rust/src/settings.rs
+++ b/rust/src/settings.rs
@@ -30,6 +30,8 @@ pub struct Settings {
 
 impl Settings {
     /// Loads and validates the coordinator settings via a configuration file.
+    ///
+    /// # Errors
     /// Fails when the loading of the configuration file or its validation failed.
     pub fn new(path: PathBuf) -> Result<Self, SettingsError> {
         let settings: Settings = Self::load(path)?;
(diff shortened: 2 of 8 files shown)


rust/src/settings.rs
```diff
@@ -15,14 +17,32 @@ pub enum SettingsError {
...
     pub api: ApiSettings,
     #[validate]
     pub pet: PetSettings,
     pub mask: MaskSettings,
+    pub log: LoggingSettings,
+}
+
+impl Settings {
+    /// Loads and validates the coordinator settings via a configuration file.
+    /// Fails when the loading of the configuration file or its validation failed.
```
janpetschexain, line 33:
```suggestion
    /// Loads and validates the coordinator settings via a configuration file.
    ///
    /// # Errors
    /// Fails when the loading of the configuration file or its validation failed.
```

configs/config.toml
```diff
@@ -1,3 +1,6 @@
+[log]
+filter = "xain_fl=debug,info"
```
janpetschexain, line 2: i thought that higher priority levels would automatically include lower priority levels when logging, like `error > warn > info > debug > trace` for the [level filters](https://docs.rs/tracing-subscriber/0.2.6/tracing_subscriber/filter/struct.LevelFilter.html). is that not the case?

Robert-Steiner: That is right. However, a filter allows us to define multiple logging directives. 
So `xain_fl=debug` means: turn on `debug` logging for the `xain_fl` module and 
`info` means turn on `info` logging for all modules.

janpetschexain: ahh true, i missed that only the first level after the filter actually applies to the filter.

PR #435, opened 2020-06-26
PB-870 document mask module

janpetschexain:
**References**
- [PB-870](https://xainag.atlassian.net/browse/PB-870)

**Summary**
- add module level documentation
- update struct and method documentation

note that this requires `cargo +nightly doc` to generate proper internal links, which is not an issue since this is supported by the hosted rust docs as well.

### DIFF
--- a/rust/src/mask/config/mod.rs
+++ b/rust/src/mask/config/mod.rs
@@ -84,6 +84,9 @@ impl TryFrom<u8> for DataType {
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
 /// The bounds of the numerical values.
+///
+/// For a value `v` to be absolutely bounded by another value `b`, it has to hold that
+/// `-b <= v <= b` or equivalently `|v| <= b`.
 pub enum BoundType {
     /// Numerical values absolutely bounded by 1.
     B0 = 0,
@@ -115,7 +118,7 @@ impl TryFrom<u8> for BoundType {
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
-/// The number of models to be aggregated at most.
+/// The maximum number of models to be aggregated at most.
 pub enum ModelType {
     /// At most 1_000 models to be aggregated.
     M3 = 3,
@@ -129,7 +132,7 @@ pub enum ModelType {
 
 impl ModelType {
     /// Gets the maximum number of models that can be aggregated for this model type.
-    pub fn nb_models_max(&self) -> usize {
+    pub fn max_nb_models(&self) -> usize {
         10_usize.pow(*self as u8 as u32)
     }
 }
@@ -151,20 +154,20 @@ impl TryFrom<u8> for ModelType {
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 /// A masking configuration.
 ///
-/// Consists of the following parts:
-/// - the order of the finite group
-/// - the data type of the numbers to be masked
-/// - the bounds of the numbers to be masked
-/// - the number of models to be aggregated at most
+/// This configuration is applied for masking, aggregation and unmasking of models.
 pub struct MaskConfig {
+    /// The order of the finite group.
     pub group_type: GroupType,
+    /// The original primitive data type of the numerical values to be masked.
     pub data_type: DataType,
+    /// The bounds of the numerical values.
     pub bound_type: BoundType,
+    /// The maximum number of models to be aggregated at most.
     pub model_type: ModelType,
 }
 
 impl MaskConfig {
-    /// Returns the number of bytes needed for an element of a mask or masked model.
+    /// Returns the number of bytes needed for an element of a mask object.
     pub(crate) fn bytes_per_number(&self) -> usize {
         let max_number = self.order() - BigUint::from(1_u8);
         (max_number.bits() + 7) / 8
--- a/rust/src/mask/config/mod.rs
+++ b/rust/src/mask/config/mod.rs
@@ -1,4 +1,4 @@
-//! Mask configuration parameters.
+//! Masking configuration parameters.
 //!
 //! See the [mask module] documentation since this is a private module anyways.
 //!
@@ -16,7 +16,7 @@ use num::{
 use thiserror::Error;
 
 #[derive(Debug, Error)]
-/// Errors related to invalid mask configurations.
+/// Errors related to invalid masking configurations.
 pub enum InvalidMaskConfigError {
     #[error("invalid group type")]
     GroupType,
@@ -55,7 +55,7 @@ impl TryFrom<u8> for GroupType {
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
-/// The original primitive data type of the numerical values.
+/// The original primitive data type of the numerical values to be masked.
 pub enum DataType {
     /// Numbers of type f32.
     F32 = 0,
@@ -128,7 +128,7 @@ pub enum ModelType {
 }
 
 impl ModelType {
-    /// Gets the number of models to be aggregated at most for this model type.
+    /// Gets the maximum number of models that can be aggregated for this model type.
     pub fn nb_models_max(&self) -> usize {
         10_usize.pow(*self as u8 as u32)
     }
@@ -149,7 +149,7 @@ impl TryFrom<u8> for ModelType {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-/// A mask configuration.
+/// A masking configuration.
 ///
 /// Consists of the following parts:
 /// - the order of the finite group
@@ -164,14 +164,13 @@ pub struct MaskConfig {
 }
 
 impl MaskConfig {
-    /// Returns the number of bytes needed for an element of a mask or masked model wrt this mask
-    /// configuration.
-    pub fn bytes_per_number(&self) -> usize {
+    /// Returns the number of bytes needed for an element of a mask or masked model.
+    pub(crate) fn bytes_per_number(&self) -> usize {
         let max_number = self.order() - BigUint::from(1_u8);
         (max_number.bits() + 7) / 8
     }
 
-    /// Gets the additional shift value for masking/unmasking wrt this mask configuration.
+    /// Gets the additional shift value for masking/unmasking.
     pub fn add_shift(&self) -> Ratio<BigInt> {
         use BoundType::{Bmax, B0, B2, B4, B6};
         use DataType::{F32, F64, I32, I64};
@@ -191,7 +190,7 @@ impl MaskConfig {
         }
     }
 
-    /// Gets the exponential shift value for masking/unmasking wrt this mask configuration.
+    /// Gets the exponential shift value for masking/unmasking.
     pub fn exp_shift(&self) -> BigInt {
         use BoundType::{Bmax, B0, B2, B4, B6};
         use DataType::{F32, F64, I32, I64};
@@ -209,7 +208,7 @@ impl MaskConfig {
         }
     }
 
-    /// Gets the finite group order value for masking/unmasking wrt this mask configuration.
+    /// Gets the finite group order value for masking/unmasking.
     pub fn order(&self) -> BigUint {
         use BoundType::{Bmax, B0, B2, B4, B6};
         use DataType::{F32, F64, I32, I64};
--- a/rust/src/mask/config/mod.rs
+++ b/rust/src/mask/config/mod.rs
@@ -1,16 +1,23 @@
+//! Mask configuration parameters.
+//!
+//! See the [mask module] documentation since this is a private module anyways.
+//!
+//! [mask module]: ../index.html
+
 pub(crate) mod serialization;
 
 use std::convert::TryFrom;
-use thiserror::Error;
 
 use num::{
     bigint::{BigInt, BigUint},
     rational::Ratio,
     traits::{pow::Pow, Num},
 };
+use thiserror::Error;
 
 #[derive(Debug, Error)]
-pub enum InvalidMaskConfig {
+/// Errors related to invalid mask configurations.
+pub enum InvalidMaskConfigError {
     #[error("invalid group type")]
     GroupType,
     #[error("invalid data type")]
@@ -34,24 +41,21 @@ pub enum GroupType {
 }
 
 impl TryFrom<u8> for GroupType {
-    type Error = InvalidMaskConfig;
+    type Error = InvalidMaskConfigError;
 
-    /// Get the group type. Fails if the encoding is unknown.
     fn try_from(byte: u8) -> Result<Self, Self::Error> {
-        use GroupType::{Integer, Power2, Prime};
-
-        Ok(match byte {
-            0 => Integer,
-            1 => Prime,
-            2 => Power2,
-            _ => return Err(InvalidMaskConfig::GroupType),
-        })
+        match byte {
+            0 => Ok(GroupType::Integer),
+            1 => Ok(GroupType::Prime),
+            2 => Ok(GroupType::Power2),
+            _ => Err(InvalidMaskConfigError::GroupType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
-/// The data type of the numbers to be masked.
+/// The original primitive data type of the numerical values.
 pub enum DataType {
     /// Numbers of type f32.
     F32 = 0,
@@ -64,53 +68,48 @@ pub enum DataType {
 }
 
 impl TryFrom<u8> for DataType {
-    type Error = InvalidMaskConfig;
+    type Error = InvalidMaskConfigError;
 
-    /// Get the data type. Fails if the encoding is unknown.
     fn try_from(byte: u8) -> Result<Self, Self::Error> {
-        use DataType::{F32, F64, I32, I64};
-
-        Ok(match byte {
-            0 => F32,
-            1 => F64,
-            2 => I32,
-            3 => I64,
-            _ => return Err(InvalidMaskConfig::DataType),
-        })
+        match byte {
+            0 => Ok(DataType::F32),
+            1 => Ok(DataType::F64),
+            2 => Ok(DataType::I32),
+            3 => Ok(DataType::I64),
+            _ => Err(InvalidMaskConfigError::DataType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
-/// The bounds of the numbers to be masked.
+/// The bounds of the numerical values.
 pub enum BoundType {
-    /// Numbers absolutely bounded by 1.
+    /// Numerical values absolutely bounded by 1.
     B0 = 0,
-    /// Numbers absolutely bounded by 100.
+    /// Numerical values absolutely bounded by 100.
     B2 = 2,
-    /// Numbers absolutely bounded by 10_000.
+    /// Numerical values absolutely bounded by 10_000.
     B4 = 4,
-    /// Numbers absolutely bounded by 1_000_000.
+    /// Numerical values absolutely bounded by 1_000_000.
     B6 = 6,
-    /// Numbers absolutely bounded by their data types' maximum absolute value.
+    /// Numerical values absolutely bounded by their original primitve data types' maximum absolute
+    /// value.
     Bmax = 255,
 }
 
 impl TryFrom<u8> for BoundType {
-    type Error = InvalidMaskConfig;
+    type Error = InvalidMaskConfigError;
 
-    /// Get the bound type. Fails if the encoding is unknown.
     fn try_from(byte: u8) -> Result<Self, Self::Error> {
-        use BoundType::{Bmax, B0, B2, B4, B6};
-
-        Ok(match byte {
-            0 => B0,
-            2 => B2,
-            4 => B4,
-            6 => B6,
-            255 => Bmax,
-            _ => return Err(InvalidMaskConfig::ModelType),
-        })
+        match byte {
+            0 => Ok(BoundType::B0),
+            2 => Ok(BoundType::B2),
+            4 => Ok(BoundType::B4),
+            6 => Ok(BoundType::B6),
+            255 => Ok(BoundType::Bmax),
+            _ => Err(InvalidMaskConfigError::ModelType),
+        }
     }
 }
 
@@ -129,51 +128,54 @@ pub enum ModelType {
 }
 
 impl ModelType {
+    /// Gets the number of models to be aggregated at most for this model type.
     pub fn nb_models_max(&self) -> usize {
-        10usize.pow(*self as u8 as u32)
+        10_usize.pow(*self as u8 as u32)
     }
 }
 
 impl TryFrom<u8> for ModelType {
-    type Error = InvalidMaskConfig;
+    type Error = InvalidMaskConfigError;
 
-    /// Get the model type. Fails if the encoding is unknown.
     fn try_from(byte: u8) -> Result<Self, Self::Error> {
-        use ModelType::{M12, M3, M6, M9};
-        Ok(match byte {
-            3 => M3,
-            6 => M6,
-            9 => M9,
-            12 => M12,
-            _ => return Err(InvalidMaskConfig::ModelType),
-        })
+        match byte {
+            3 => Ok(ModelType::M3),
+            6 => Ok(ModelType::M6),
+            9 => Ok(ModelType::M9),
+            12 => Ok(ModelType::M12),
+            _ => Err(InvalidMaskConfigError::ModelType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-/// A mask configuration name. Consists of identifiers for its parts:
+/// A mask configuration.
+///
+/// Consists of the following parts:
 /// - the order of the finite group
 /// - the data type of the numbers to be masked
 /// - the bounds of the numbers to be masked
 /// - the number of models to be aggregated at most
 pub struct MaskConfig {
-    pub group_type: GroupType,
-    pub data_type: DataType,
-    pub bound_type: BoundType,
-    pub model_type: ModelType,
+    pub(crate) group_type: GroupType,
+    pub(crate) data_type: DataType,
+    pub(crate) bound_type: BoundType,
+    pub(crate) model_type: ModelType,
 }
 
 impl MaskConfig {
-    /// Return the number of bytes needed to represent a mask element
-    /// with this configuration
-    pub fn bytes_per_digit(&self) -> usize {
+    /// Returns the number of bytes needed for an element of a mask or masked model wrt this mask
+    /// configuration.
+    pub fn bytes_per_number(&self) -> usize {
         let max_number = self.order() - BigUint::from(1_u8);
         (max_number.bits() + 7) / 8
     }
 
+    /// Gets the additional shift value for masking/unmasking wrt this mask configuration.
     pub fn add_shift(&self) -> Ratio<BigInt> {
         use BoundType::{Bmax, B0, B2, B4, B6};
         use DataType::{F32, F64, I32, I64};
+
         match self.bound_type {
             B0 => Ratio::from_integer(BigInt::from(1)),
             B2 => Ratio::from_integer(BigInt::from(100)),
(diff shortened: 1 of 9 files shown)


rust/src/mask/config/mod.rs
```diff
@@ -34,24 +41,21 @@ pub enum GroupType {
...
+            2 => Ok(GroupType::Power2),
+            _ => Err(InvalidMaskConfigError::GroupType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
-/// The data type of the numbers to be masked.
+/// The original primitive data type of the numerical values.
```
little-dude, line 58: I feel like "numerical values" is a bit too broad. It think "numerical values to mask" would make what this is about clearer.

rust/src/mask/config/mod.rs
```diff
@@ -129,29 +128,30 @@ pub enum ModelType {
 }
 
 impl ModelType {
+    /// Gets the number of models to be aggregated at most for this model type.
```
little-dude, line 131: I'd suggest

> Get the maximum number of models that can be aggregated for this model type

Sidenote: I've used the imperative form in the parts I've documented. I don't mind using the third person but we should agree on something to be consistent.

rust/src/mask/config/mod.rs
```diff
@@ -129,29 +128,30 @@ pub enum ModelType {
...
+            9 => Ok(ModelType::M9),
+            12 => Ok(ModelType::M12),
+            _ => Err(InvalidMaskConfigError::ModelType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-/// A mask configuration name. Consists of identifiers for its parts:
+/// A mask configuration.
```
little-dude, line 152: Not sure, but shouldn't we speak about "masking configuration" rather than "mask configuration"?

rust/src/mask/config/mod.rs
```diff
@@ -164,16 +164,18 @@ pub struct MaskConfig {
 }
 
 impl MaskConfig {
-    /// Return the number of bytes needed to represent a mask element
-    /// with this configuration
-    pub fn bytes_per_digit(&self) -> usize {
+    /// Returns the number of bytes needed for an element of a mask or masked model wrt this mask
```
little-dude, line 167: Several remarks:

1. In the code we use `MaskObject` to something that is either a mask or a masked model. Maybe we can do the same in the documentation?

2. I find that "to represent" from the sentence makes it a little less clear personally, but maybe it's just me

3. Unrelated to this PR but I wonder whether this should be `pub` actually: it really is just an implementation detail.

little-dude: ah and :+1: for the `s/digit/number`

rust/src/mask/config/mod.rs
```diff
@@ -164,16 +164,18 @@ pub struct MaskConfig {
...
+    /// configuration.
+    pub fn bytes_per_number(&self) -> usize {
         let max_number = self.order() - BigUint::from(1_u8);
         (max_number.bits() + 7) / 8
     }
 
+    /// Gets the additional shift value for masking/unmasking wrt this mask configuration.
```
little-dude, line 174: nit: "wrt this" seems a bit heavy to me. This method belongs to `MaskConfig` so maybe we don't even need to explicitly say it:

> Gets the additional shift value for masking/unmasking

If we want to keep it, maybe we can just use "with"

> Gets the additional shift value for masking/unmasking with this configuration

rust/src/mask/masking.rs
```diff
@@ -65,21 +73,43 @@ impl Into<MaskObject> for Aggregation {
...
     pub fn len(&self) -> usize {
         self.object.data.len()
     }
 
+    /// Gets the mask configuration of the aggregator.
     pub fn config(&self) -> MaskConfig {
         self.object.config
     }
 
+    /// Validates if unmasking of the aggregated, masked model with the given `mask` may be
```
little-dude, line 94:
```suggestion
    /// Validates if unmasking of the aggregated masked object with the given `mask` may be
```

little-dude, line 94:
```suggestion
    /// Validates if unmasking of the aggregated masked model with the given `mask` may be
```

rust/src/mask/masking.rs
```diff
@@ -65,21 +73,43 @@ impl Into<MaskObject> for Aggregation {
...
+    /// # Errors
+    /// Fails in one of the following cases:
+    /// - The aggregator has not yet aggregated any models.
+    /// - The number of aggregated masked models is larger than the chosen mask configuration
+    ///   allows.
+    /// - The mask configuration of the aggregator and of the `mask` don't coincide.
+    /// - The length of the aggregated masked model and the `mask` don't coincide.
+    /// - The `mask` itself is invalid.
+    ///
+    /// Even though it does not produce any meaningful values, it may be validated that "unmasking"
```
little-dude, line 109: I don't really understand this last sentence

rust/src/mask/masking.rs
```diff
@@ -128,6 +172,25 @@ impl Aggregation {
             .collect()
     }
 
+    /// Validates if aggregation of the aggregated mask or masked model with the given `object`
```
little-dude, line 175: As said in another comment, maybe we could talk about "mask object" here.

rust/src/mask/config/mod.rs
janpetschexain: changed.

rust/src/mask/config/mod.rs
janpetschexain: changed the wording.

janpetschexain: i agree that we should have consistent styles, i have no strong preference for imperative or non-imperative style, i guess we should consistify this across the code base once the contributions to the documentations are all merged.

rust/src/mask/config/mod.rs
janpetschexain: not sure as well^^ i guess "masking configuration" is closer to the process and less focused on the actual mask. i'll change it accordingly. the english language can be strange.

rust/src/mask/config/mod.rs
janpetschexain: some questions on the remarks:
1. i currently use the phrasing "mask or masked model"  to express this dual use of `MaskObject`. do you mean the same or that it should be replaced with "mask object"?
2. not sure what you mean here, since the less clear formulation is part of the old version, whereas it doesn't appear in the new version.

i agree on making 3. private.

rust/src/mask/config/mod.rs
janpetschexain: removed the redundancy.

rust/src/mask/masking.rs
janpetschexain: changed.

rust/src/mask/masking.rs
janpetschexain: the reason lies in the dual use of `MaskObject`. usually we have a masked model and validate that it can be unmasked safely with a mask (both of which are `MaskObject`s of course). but nothing stops a user to apply this validation to a mask itself, so to say "validate if this mask can be unmasked with that mask" or even "validate if this mask can be unmasked with that masked model". this will not produce any meaningful result, but is technically possible due to the chosen types.

rust/src/mask/config/mod.rs
```diff
@@ -64,53 +68,48 @@ pub enum DataType {
...
+    /// Numerical values absolutely bounded by 100.
     B2 = 2,
-    /// Numbers absolutely bounded by 10_000.
+    /// Numerical values absolutely bounded by 10_000.
     B4 = 4,
-    /// Numbers absolutely bounded by 1_000_000.
+    /// Numerical values absolutely bounded by 1_000_000.
     B6 = 6,
-    /// Numbers absolutely bounded by their data types' maximum absolute value.
+    /// Numerical values absolutely bounded by their original primitve data types' maximum absolute
```
Robert-Steiner, line 96:
```suggestion
    /// Numerical values absolutely bounded by their original primitive data types maximum absolute
```

rust/src/mask/masking.rs
```diff
@@ -101,6 +131,20 @@ impl Aggregation {
         Ok(())
     }
 
+    /// Unmasks the aggregated masked model with the given `mask`.
+    ///
+    /// It should be checked that [`validate_unmasking()`] returns `true` before calling this, since
```
Robert-Steiner, line 136: Not related to this pr:
Was there a reason to not include the validation in the unmask method?

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! // unmask the aggregated masked model with the aggregated mask
+//! if let Ok(_) = model_aggregator.validate_unmasking(&global_mask) {
+//!     let global_model = model_aggregator.unmask(global_mask);
+//!     assert_eq!(
+//!         global_model,
+//!         Model::from_primitives_bounded(vec![0.5_f32; number_weights].into_iter()),
+//!     );
+//! };
+//! ```
+
```
Robert-Steiner, line 183: The module documentation is great 👍

rust/src/mask/config/mod.rs
janpetschexain: changed.

rust/src/mask/masking.rs
janpetschexain: it was combined in the first implementation, but was separated in the refactoring. the reason is twofold, first this provides some logical and visual separation (one can of course argue if this is needed at all since the methods are not that complex). and second this allows for atomic changes where all the conditions have to be checked first before any action is taken (this is not the case for unmasking directly, but is for example needed for aggregation, and this way we have consistency in the method style).

rust/src/mask/config/mod.rs
janpetschexain: since i'm not sure how to proceed on this issue, i'll leave it as "mask or masked model" for now and it can be changed later on if needed.

janpetschexain: just to leave another idea for later: we might even make the `MaskObject` a trait and then have distinct structs for `Mask` and `MaskedModel` to avoid type confusion.

rust/src/mask/masking.rs
janpetschexain: i updated [the other comment](https://github.com/xainag/xain-fl/pull/435#discussion_r448936330) accordingly.

rust/src/mask/masking.rs
Robert-Steiner: Ok that makes sense. Thanks for the explanation!

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//!
+//! // derive the masks of the local masked models
+//! let local_mask_1 = local_mask_seed_1.derive_mask(number_weights, config);
+//! let local_mask_2 = local_mask_seed_2.derive_mask(number_weights, config);
+//! ```
+//!
+//! ## Aggregation
+//! Masked models can be aggregated via an [`Aggregation`]. Masks can be aggregated via an
+//! [`Aggregation`] as well, since an aggregated masked model can only be unmasked by the
+//! aggregated masks belonging to the masked models. It should always be validated that the
```
finiteprods, line 116:
```suggestion
//! aggregation of masks for each model.
```

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! let (local_mask_seed_1, masked_local_model_1) = Masker::new(config).mask(scalar, local_model_1);
+//! let (local_mask_seed_2, masked_local_model_2) = Masker::new(config).mask(scalar, local_model_2);
+//!
+//! // derive the masks of the local masked models
+//! let local_mask_1 = local_mask_seed_1.derive_mask(number_weights, config);
+//! let local_mask_2 = local_mask_seed_2.derive_mask(number_weights, config);
+//! ```
+//!
+//! ## Aggregation
+//! Masked models can be aggregated via an [`Aggregation`]. Masks can be aggregated via an
```
finiteprods, line 114:
```suggestion
//! Masked models can be aggregated via an [`Aggregation`]. Masks themselves can be aggregated via an
```

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! [`Aggregation`] as well, since an aggregated masked model can only be unmasked by the
```
finiteprods, line 115:
```suggestion
//! [`Aggregation`] as well. An aggregated masked model can only be unmasked by the
```

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! aggregated masks belonging to the masked models. It should always be validated that the
+//! aggregation may be safely performed wrt the chosen masking configuration before actually
```
finiteprods, line 117:
```suggestion
//! Aggregation should always be validated beforehand so that it may be safely performed wrt the chosen masking configuration
```

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! aggregating to avoid the possible loss of information.
```
finiteprods, line 118:
```suggestion
//! without possible loss of information.
```

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! if let Ok(_) = model_aggregator.validate_aggregation(&masked_local_model_1) {
+//!     model_aggregator.aggregate(masked_local_model_1);
+//! };
+//! if let Ok(_) = model_aggregator.validate_aggregation(&masked_local_model_2) {
+//!     model_aggregator.aggregate(masked_local_model_2);
+//! };
+//! ```
+//!
+//! ## Unmasking
+//! A masked model can be unmasked by the corresponding mask via an [`Aggregation`] in the end. It
```
finiteprods, line 152:
```suggestion
//! A masked model can be unmasked by the corresponding mask via an [`Aggregation`]. It
```
or alternatively, elaborate on "in the end"

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! // mask the local models
+//! let (local_mask_seed_1, masked_local_model_1) = Masker::new(config).mask(scalar, local_model_1);
+//! let (local_mask_seed_2, masked_local_model_2) = Masker::new(config).mask(scalar, local_model_2);
+//!
+//! // derive the masks of the local masked models
+//! let local_mask_1 = local_mask_seed_1.derive_mask(number_weights, config);
+//! let local_mask_2 = local_mask_seed_2.derive_mask(number_weights, config);
+//! ```
+//!
+//! ## Aggregation
```
finiteprods, line 113: sorry i should have used a single block for the below rather than chopping it up....

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! };
+//! if let Ok(_) = model_aggregator.validate_aggregation(&masked_local_model_2) {
+//!     model_aggregator.aggregate(masked_local_model_2);
+//! };
+//! ```
+//!
+//! ## Unmasking
+//! A masked model can be unmasked by the corresponding mask via an [`Aggregation`] in the end. It
+//! should always be validated that the unmasking may be safely performed wrt the chosen mask
+//! configuration before actually unmasking to avoid the possible loss of information.
```
finiteprods, line 154: "validation should always be validated beforehand so that it may be safely performed ...." similar to above

rust/src/mask/model.rs
```diff
@@ -58,32 +68,33 @@ enum PrimitiveType {
...
 #[derive(Error, Debug)]
 #[error("Could not convert primitive type {0:?} to model weight")]
+/// Errors related to model conversion from primitives.
 pub struct PrimitiveCastError<P: Debug>(P);
 
-/// Convert this type into a an iterator of type `P`. This trait is
-/// used to convert a [`Model`], which has its own internal
-/// representation of the weights into primitive types (`f64`, `f32`,
-/// `i32` `i64`).
+/// An interface to convert a collection of numerical values into a an iterator of primitive values.
```
finiteprods, line 82:
```suggestion
/// An interface to convert a collection of numerical values into an iterator of primitive values.
```

rust/src/mask/model.rs
```diff
@@ -58,32 +68,33 @@ enum PrimitiveType {
...
+///
+/// This trait is used to convert a [`Model`], which has its own internal representation of the
+/// weights, into primitive types ([`f32`], [`f64`], [`i32`], [`i64`]). The opposed trait is
```
finiteprods, line 85:
```suggestion
/// weights, into primitive types ([`f32`], [`f64`], [`i32`], [`i64`]). The opposite trait is
```
or "dual"

rust/src/mask/model.rs
```diff
@@ -92,15 +103,23 @@ pub trait IntoPrimitives<P: 'static>: Sized {
     }
 }
 
-/// Convert a stream of numerical primitive types (`i32`, `i64`,
-/// `f32`, `f64`) into this type.
+/// An interface to convert a collection of primitive values into a an iterator of numerical values.
+///
+/// This trait is used to convert primitive types ([`f32`], [`f64`], [`i32`], [`i64`]) into a
+/// [`Model`], which has its own internal representation of the weights. The opposed trait is
```
finiteprods, line 109: "opposite" / "dual"

rust/src/mask/object/serialization.rs
```diff
@@ -26,29 +36,34 @@ impl<T: AsRef<[u8]>> MaskObjectBuffer<T> {
...
 
+    /// Creates a new buffer from `bytes`.
     pub fn new_unchecked(bytes: T) -> Self {
         Self { inner: bytes }
     }
 
+    /// Checks if this buffer conforms to the required buffer length for mask objects.
+    ///
+    /// # Errors
+    /// Fails if the buffer is to small.
```
finiteprods, line 47:
```suggestion
    /// Fails if the buffer is too small.
```

rust/src/mask/config/mod.rs
finiteprods: personally i prefer "mask object" to the wordier "mask or masked model". but i'd be happy for this to be deferred.

rust/src/mask/masking.rs
janpetschexain: i'll update the docs to clarify this.

rust/src/mask/config/mod.rs
janpetschexain: i'll change it to "mask object". if we implement changes to the `MaskObject` later on, we can change the docs anyways.

janpetschexain:
> The one thing I noticed is that everything is using the third person. I have been using the imperative form in my own documentation effort so we should agree on something with @Robert-Steiner and @finiteprods on that matter.

i added a [ticket](https://xainag.atlassian.net/browse/PB-887) for this to the backlog.

rust/src/mask/config/mod.rs
```diff
@@ -34,24 +41,21 @@ pub enum GroupType {
 }
 
 impl TryFrom<u8> for GroupType {
-    type Error = InvalidMaskConfig;
+    type Error = InvalidMaskConfigError;
 
-    /// Get the group type. Fails if the encoding is unknown.
```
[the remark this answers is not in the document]
janpetschexain: the short answer is, that the trait definition already provides a docstring which is automatically applied to the impl.

the long answer is manyfold:
- in most places we don't give a custom docstring to a trait impl and this specific docstring doesn't provide more information than what is already clear from the trait definition docstring and the type on which the trait is implemented. therefore i removed it for consistency.
- in some cases it makes sense to have a custom docstring on a trait impl to overwrite the trait defined docstring, for example if the trait defined docstring is too generic. but this has an ugly edge case which i recently encountered: if the trait definition has a default impl, then there is no way to overwrite the default docstring without reimplementing the default method, like in the following example:

```rust
trait T {
    /// docstring for method without default impl
    fn method(&self);

    //// docstring for method with default impl
    fn method_with_default(&self) {
        println!("this has a default impl")
    }
}

struct S;

impl T for S {
    fn method(&self) {
        println!("this has no default impl")
    }

// the following is not possible (one would have to implement this again even though it has a default impl in the trait)

//    /// custom docstring for method with default impl
//    fn method_with_default(&self);
}
```

rust/src/mask/config/mod.rs
```diff
@@ -64,53 +68,48 @@ pub enum DataType {
...
+            3 => Ok(DataType::I64),
+            _ => Err(InvalidMaskConfigError::DataType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
-/// The bounds of the numbers to be masked.
+/// The bounds of the numerical values.
```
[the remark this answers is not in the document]
janpetschexain: i'll add a comment to the docs.

rust/src/mask/config/mod.rs
```diff
@@ -129,29 +128,30 @@ pub enum ModelType {
 }
 
 impl ModelType {
+    /// Gets the maximum number of models that can be aggregated for this model type.
     pub fn nb_models_max(&self) -> usize {
```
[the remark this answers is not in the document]
janpetschexain: i agree that `nb` is an uncommon abbreviation for `number`. `num` is more common, but both are technically wrong and the correct `no` should be used as an abbreviation. on the other hand `no` might lead to confusion.

i'll go with the same words for now, but update the word order as suggested. naming might then be handled in another ticket.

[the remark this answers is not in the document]
janpetschexain: closed this duplicate.

rust/src/mask/config/mod.rs
```diff
@@ -129,29 +128,30 @@ pub enum ModelType {
...
+            _ => Err(InvalidMaskConfigError::ModelType),
+        }
     }
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-/// A mask configuration name. Consists of identifiers for its parts:
+/// A masking configuration.
+///
+/// Consists of the following parts:
```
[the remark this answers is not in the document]
janpetschexain: the config applies to more than aggregation, which is masking, aggregation and unmasking. i'll update the docstring accordingly.

rust/src/mask/config/mod.rs
```diff
@@ -204,6 +208,7 @@ impl MaskConfig {
         }
     }
 
+    /// Gets the finite group order value for masking/unmasking.
```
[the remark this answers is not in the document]
janpetschexain: i'm a bit hesitant on this matter. there is always the question on wether 1) the docstring explains only how this method is used for what, or 2) the docstring also explains the details on how and why this works as it does. i left out the mathematical details on the latter and included a more detailed documentation on the former in the module docs.

also we want to change this catalogue, which is quite limiting in the choices it provides, to a computable config which then covers all cases in the future. this will make a docstring on the current details obsolete.

rust/src/mask/config/serialization.rs
```diff
@@ -25,10 +37,15 @@ impl<T: AsRef<[u8]>> MaskConfigBuffer<T> {
...
 
+    /// Creates a new buffer from `bytes`.
     pub fn new_unchecked(bytes: T) -> Self {
         Self { inner: bytes }
     }
 
+    /// Checks if this buffer conforms to the required buffer length for masking configurations.
+    ///
+    /// # Errors
+    /// Fails if the buffer is to small.
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/masking.rs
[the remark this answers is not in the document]
janpetschexain: good catch, i'll change the respective occurences.

rust/src/mask/masking.rs
```diff
@@ -165,39 +238,44 @@ impl Aggregation {
...
 
+    /// Creates a new masker with the given masking `config`uration and `seed`.
     pub fn with_seed(config: MaskConfig, seed: MaskSeed) -> Self {
         Self { config, seed }
     }
 }
 
 impl Masker {
-    /// Mask the model wrt the mask configuration. Enforces bounds on the scalar and weights.
+    /// Masks the model wrt the masking configuration. Enforces bounds on the scalar and weights.
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/masking.rs
```diff
@@ -165,39 +238,44 @@ impl Aggregation {
...
-    /// - shift the weights into the finite group
-    /// - mask the weights with random elements from the finite group
+    /// - Clamp the scalar and the weights according to the masking configuration.
+    /// - Shift the weights into the non-negative reals.
+    /// - Shift the weights into the non-negative integers.
+    /// - Shift the weights into the finite group.
+    /// - Mask the weights with random elements from the finite group.
+    ///
+    /// The random elements are derived from a seeded PRNG. Unmasking as performed in [`unmask()`]
+    /// proceeds in reverse order. For more details see this [confluence page].
```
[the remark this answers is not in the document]
janpetschexain: i thought about this as well. there is no harm but also no benefit, i'll remove the link for now.

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
+//! Masking, aggregation and unmasking of models.
+//!
+//! # Models
+//! A [`Model`] is a collection of weights/parameters which are represented as finite numerical
+//! values (i.e. rational numbers) of arbitrary precision. Particularly, a model in itself is not
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! bound to a primitive data type, but it can be created from those and converted back into those.
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! embedded. The smaller the gap between the maximum possible embedded weights and the group order
+//! is, the less theoretically possible information flow about the masks may be observed. Specific
+//! group orders provide potentially higher performance on the other hand, which always makes this
+//! a tradeoff between security and performance. The group type variants are:
+//! - Integer: no gap but potentially slowest performance.
+//! - Prime: usually small gap with higher performance.
+//! - Power2: usually higher gap with potentially highest performance.
+//!
+//! ## Data type
+//! The [`DataType`] describes the original primitives data type of the model weights. This
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! influences in combination with the bound type the preserved decimal places of the model weights
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/mod.rs
```diff
@@ -1,13 +1,204 @@
...
+//! # Masking, aggregation and unmasking
+//! Local models should be masked (i.e. encrypted) before they are communicated somewhere else to
+//! protect the possibly sensitive information learned from local data. The masking should allow
+//! for masked models to be aggregated while they are still masked (i.e. homomorphic encryption).
+//! Then the aggregated masked model can safely be unmasked without jeopardizing the secrecy of
+//! personal information if the model is generalized enough.
+//!
+//! ## Masking
+//! A [`Model`] can be masked with a [`Masker`], which requires a [`MaskConfig`]. During the
+//! masking the model weights are scaled, then embedded as elements of the chosen finite group and
```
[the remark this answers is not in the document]
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/mod.rs
janpetschexain: changed.

rust/src/mask/model.rs
janpetschexain: changed.

rust/src/mask/model.rs
janpetschexain: changed.

rust/src/mask/model.rs
janpetschexain: changed.

rust/src/mask/object/serialization.rs
janpetschexain: changed.

rust/src/mask/config/mod.rs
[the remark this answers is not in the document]
finiteprods: appreciate the detailed explanation of this edge case :+1:

rust/src/mask/config/mod.rs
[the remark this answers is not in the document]
finiteprods: that's fine, it's only suggestive. I also rule out `no` for the same reason.

rust/src/mask/config/mod.rs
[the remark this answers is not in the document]
finiteprods: i agree the module-level docs do a great job of describing usage at a high-level. perhaps it's just that the use of "magic numbers" always makes me a bit uncomfortable. But i take your point that it can be tackled later when we have it more generic.

PR #437, opened 2020-06-30
modularise and document client

finiteprods:
- new `client` module containing (what was previously) `client.rs`, `request.rs` and `participant.rs`.
- minor leftover fixes from [networking](https://github.com/xainag/xain-fl/pull/420) including to `sdk/api.rs`.
- module documentation for `client`, request / proxy also documented.
- module documentation for `participant` as well.


rust/src/client/mod.rs
```diff
@@ -1,3 +1,23 @@
+//! Provides client-side functionality to connect to a XayNet service.
+//!
+//! This functionality includes:
+//!
+//! * Abiding by (the underlying [`Participant`]'s side of) the PET protocol.
+//! * Handling the network communication with the XayNet service, including
+//!   polling of service data.
+//!
+//! A [`Client`] has an intentionally simple API - the idea is that it is
+//! initialised with some settings, and then [`start()`]ed. Currently for
```
janpetschexain, line 10: ``[`start()`]`` results in a dead link (the rustdoc auto linker has some issue yet in resolving method links). this should easily be fixed by adding the following line to the end of this documentation block:
```rust
//! [`start()`]: #method.start
```

rust/src/client/mod.rs
```diff
@@ -1,3 +1,23 @@
...
+//! simplicity, clients that have started running will do so indefinitely. It is
+//! therefore the user's responsibility to terminate clients that are no longer
+//! needed. Alternatively, it may be more convenient to run just a single round
+//! (or a known fixed number of rounds). In this case, use [`during_round()`].
```
janpetschexain, line 14: ``[`during_round()`]`` is a dead link as well:
```rust
//! [`during_round()`]: #method.during_round
```

rust/src/client/mod.rs
```diff
@@ -1,3 +1,23 @@
...
+//! For examples of usage, see the `test-drive` scripts.
+//!
+//! **Note. At present, the [`Client`] implementation is somewhat tightly
```
janpetschexain, line 17: not sure if `**Note` was intended to be bold like `*Note*`?

rust/src/client/mod.rs
```diff
@@ -6,15 +26,19 @@ use tokio::time;
...
     CoordinatorPublicKey,
     InitError,
     PetError,
 };
 
+mod request;
+pub use request::Proxy;
+
+pub mod participant;
+pub use participant::{Participant, Task};
```
janpetschexain, line 40: to avoid duplication by public re-exports, the module should become private to the crate:
```suggestion
mod participant;
pub use participant::{Participant, Task};
```
this also avoids having a `Modules` and `Re-exports` section in the docs containing `participant.rs` and focuses the docs on this module docs.

rust/src/client/mod.rs
```diff
@@ -108,8 +133,9 @@ impl Client {
     /// `id`: an ID to assign to the [`Client`].
     /// `handle`: handle for communicating with the (local) service.
```
janpetschexain, line 134: i'd suggest to add line breaks or an enumeration, because markdown writes this all in one line. similarly for some other methods below.

rust/src/client/mod.rs
```diff
@@ -6,15 +26,19 @@ use tokio::time;
...
 
+mod request;
+pub use request::Proxy;
+
+pub mod participant;
+pub use participant::{Participant, Task};
+
 #[derive(Debug, Error)]
 /// Client-side errors
 pub enum ClientError {
```
janpetschexain, line 44: do you know if there is a way to include the contexts of `thiserror`, which are annotated to the variants anyways, into the docstrings? otherwise it's probably nice to have an extra docstring for the variants mirroring the contexts. like:
```rust
pub enum ClientError {
    #[error("failed to initialise participant: {0}")]
    /// Failed to initialize participants.
    ParticipantInitErr(InitError),

    // etc
}
```

rust/src/client/request.rs
```diff
@@ -1,7 +1,35 @@
+//! Provides functionality to enable clients to communicate with a XayNet
```
janpetschexain, line 1: the `request` module is private, therefore this documentation is invisible in the generated docs. i'd suggest to make it a subsection of the `client/mod.rs` docs.

rust/src/client/request.rs
```diff
@@ -1,7 +1,35 @@
...
+//!
+//! * Wraps either a [`Handle`] (for local comms) or a [`ClientReq`] (for remote
+//!   comms over HTTP).
+//! * In the latter case, deals with logging and wrapping of network errors.
+//! * Deals with deserialization
+//!
+//! [`ClientReq`] is responsible for building the HTTP request and extracting
+//! the response body. As an example:
+//!
+//! `async fn get_sums(&self) -> Result<Option<bytes::Bytes>, reqwest::Error>`
```
janpetschexain, line 15: i've seen that you removed the code block because of running doc tests. if you want to keep the block, you can annotate it with `no_rust` which stops rustdoc from running this code. like
```
//! ```no_rust
//! // code
//! ```
```

rust/src/mask/model.rs
```diff
@@ -78,6 +78,8 @@ pub trait IntoPrimitives<P: 'static>: Sized {
     /// to `P`.
     fn into_primitives(self) -> Box<dyn Iterator<Item = Result<P, ModelCastError>>>;
 
+    fn to_primitives(&self) -> Box<dyn Iterator<Item = Result<P, ModelCastError>>>;
```
janpetschexain, line 81: this method should get some documentation since it is in a public trait. to align it with the changes in #435 i would suggest:
```rust
    /// Creates an iterator from numerical values that yields converted primitive values.
    ///
    /// # Errors
    /// Yields an error for each numerical value that can't be converted into a primitive value.
```

PR #441, opened 2020-07-02
PB-877 document crate root

janpetschexain:
**References**
- [PB-877](https://xainag.atlassian.net/browse/PB-877)

**Summary**
- add the intro text from the getting started page

links to the getting started readme and the homepage need to be added later once they are updated.


janpetschexain:
what do you guys think of moving the types which are currently at the crate root to some of the modules? we now have a more deeply structured crate than before when we introduced the types.

PR #443, opened 2020-07-03
add participant settings

Robert-Steiner:
This PR is not yet finished, but I have already opened it so that we can start a discussion about how we integrate the client setting in the client/participant and in the sdk.


rsaffi:
@Robert-Steiner I just rebased against the updated `master`!

PR #445, opened 2020-07-06
PB-846 add url input to ffi client

janpetschexain:
**References**
- [PB-846](https://xainag.atlassian.net/browse/PB-846)

**Summary**
- add a url input to the ffi client constructor
- move the flag which indicates a new round to the ffi client (`has_new_coord_pk_since_last_check`) to the end of the client round loop for atomic behavior

### DIFF
--- a/rust/src/sdk/api.rs
+++ b/rust/src/sdk/api.rs
@@ -37,9 +37,10 @@
 //! [#69173]: https://github.com/rust-lang/rust/issues/69173
 
 use std::{
+    ffi::CStr,
     iter::{IntoIterator, Iterator},
     mem,
-    os::raw::{c_int, c_uint, c_ulonglong, c_void},
+    os::raw::{c_char, c_int, c_uint, c_ulonglong, c_void},
     panic,
     ptr,
 };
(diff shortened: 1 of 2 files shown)


rust/src/sdk/api.rs
```diff
@@ -102,21 +103,33 @@ pub struct FFIClient {
...
-/// of its `input` across the FFI-boundary.
-pub unsafe extern "C" fn new_client(period: c_ulonglong) -> *mut FFIClient {
-    if period == 0 {
+/// The method dereferences from the raw pointer arguments. Therefore, the behavior of the method is
+/// undefined if the arguments don't point to valid objects.
+pub unsafe extern "C" fn new_client(address: *const c_char, period: c_ulonglong) -> *mut FFIClient {
+    if address.is_null() || period == 0 {
         return ptr::null_mut() as *mut FFIClient;
     }
+    let address = if let Ok(address) = unsafe {
```
Robert-Steiner, line 123: Just for my understanding.
`address` points to the address of c_char. So it does not copy the value of `address` but since `address` is `*const` it is safe because the value can't be dropped in the c-code. Is that right?

janpetschexain: that's right, memory management for the original string object is a task for the original owner of the string. to summarize what this is doing:
- input a char pointer to a string
- do some utf-8 and null validation
- create a rust reference to the original string object

`CStr` is closely related to `&str`, whereas the owned version are the `CString` which is closely related to `String`. it is important that the `CStr` reference (much like the `&str`)  is only valid as long as the original string object exists. this approach should be fine for a short read at the time of calling. if we need to store the address for later use, we should make a proper copy via `CString`.

PR #448, opened 2020-07-07
document HTTP server

finiteprods:
Documents the server-side of the networking code.
This was left out of #437 since it is not part of the `client` module.

### DIFF
--- a/rust/src/rest.rs
+++ b/rust/src/rest.rs
@@ -10,6 +10,12 @@ use warp::{
     Filter,
 };
 
+/// Starts a HTTP server at the given address, listening to GET requests for
+/// data and POST requests containing PET messages.
+///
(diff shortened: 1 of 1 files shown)


rust/src/rest.rs
```diff
@@ -1,3 +1,5 @@
+//! A RESTful API for the PET protocol interactions.
```
little-dude, line 1: This is a detail, but I'm not sure our API qualifies as RESTful. We don't have Create, Update, or Delete operations, we use a custom binary format for the messages, some GET requests carry a binary payload, etc. I think we should just talk about "HTTP API".

finiteprods: true; all in all, it's pretty unRESTful :) I'll open a PR to fix.

PR #456, opened 2020-07-09
PB-927 update num dependency and add target checks

janpetschexain:
**References**
- [PB-927](https://xainag.atlassian.net/browse/PB-927)

**Summary**
- update the `num` dependency to `0.3.0`
- fix a breaking change in byte counting for the mask serialization
- add some target checks for the mask serialization

### DIFF
--- a/rust/src/mask/config/mod.rs
+++ b/rust/src/mask/config/mod.rs
@@ -186,7 +186,7 @@ impl MaskConfig {
             // the largest bpn from the masking configuration catalogue is currently 173, hence
             // this is almost impossible on 32 bits targets and smaller targets are currently not
             // of interest for us
-            unimplemented!("the employed masking config is not supported on the target")
+            panic!("the employed masking config is not supported on the target")
         } else {
             bpn as usize
         }
--- a/rust/src/mask/config/mod.rs
+++ b/rust/src/mask/config/mod.rs
@@ -20,8 +20,6 @@ use thiserror::Error;
 const MAX_BPN: u64 = u16::MAX as u64;
 #[cfg(target_pointer_width = "32")]
 const MAX_BPN: u64 = u32::MAX as u64;
-#[cfg(target_pointer_width = "64")]
-const MAX_BPN: u64 = u64::MAX as u64;
 
 #[derive(Debug, Error)]
 /// Errors related to invalid masking configurations.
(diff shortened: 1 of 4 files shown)


rust/src/mask/config/mod.rs
```diff
@@ -168,9 +168,24 @@ pub struct MaskConfig {
...
 impl MaskConfig {
     /// Returns the number of bytes needed for an element of a mask object.
+    ///
+    /// # Panics
+    /// Panics if the bytes per number can't be represented as usize.
     pub(crate) fn bytes_per_number(&self) -> usize {
         let max_number = self.order() - BigUint::from(1_u8);
-        (max_number.bits() + 7) / 8
+        let bpn = (max_number.bits() + 7) / 8;
+        match mem::size_of::<usize>() {
```
little-dude, line 177: Looks like there's a cfg flag for this: https://stackoverflow.com/a/41896499/1836144
It would make the intent a bit clearer and have the benefit of happening at compile time.

janpetschexain: that's useful, i'll update it.

janpetschexain: but this just helps us with the `size_of::<usize>()` which is only part of the condition, since it is dependent on the value of `bpn` as well which in turn is not known at compile time. not sure if we can split the `match guard` then, wdyt?

little-dude: Something like this maybe:

```rust
#[cfg(target_pointer_width = "64")]
const MAX_BPN: u64 = u64::MAX as u64;

#[cfg(target_pointer_width = "32")]
const MAX_BPN: u32 = u32::MAX as u32;

#[cfg(target_pointer_width = "16")]
const MAX_BPN: u16 = u16::MAX as u16;
```

Then the condition simply becomes:
```rust
if bpn > MAX_BPN {
    panic!("not good");
}
```

janpetschexain: i've updated this in a similar fashion as suggested.

rust/src/mask/config/mod.rs
```diff
@@ -168,9 +176,20 @@ pub struct MaskConfig {
...
+    /// Panics if the bytes per number can't be represented as usize.
     pub(crate) fn bytes_per_number(&self) -> usize {
         let max_number = self.order() - BigUint::from(1_u8);
-        (max_number.bits() + 7) / 8
+        let bpn = (max_number.bits() + 7) / 8;
+        if bpn > MAX_BPN {
+            // the largest bpn from the masking configuration catalogue is currently 173, hence
+            // this is almost impossible on 32 bits targets and smaller targets are currently not
+            // of interest for us
+            unimplemented!("the employed masking config is not supported on the target")
```
little-dude, line 189: I think `panic!` is more appropriate here, since it _cannot_ ever be implemented.

little-dude:
Clippy fails with 

```
    --> src/mask/config/mod.rs:185:12
    |
185 |         if bpn > MAX_BPN {
    |            ^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::absurd_extreme_comparisons)]` on by default
    = help: because `MAX_BPN` is the maximum value for this type, this comparison is always false
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons
```

in for 64 bits targets. We could either silence clippy or add a `cfg!` to not do the comparison for 64 bits target.

rust/src/mask/config/mod.rs
janpetschexain: changed. i would guess it is possible to implement this with a larger fixed size type than `usize` and a lot of ugly memory juggling, but i get your point.

PR #461, opened 2020-07-13
PB-926 GitHub Actions for Docker Hub (PR, merges to master and tagged releases)

rsaffi:
Edited the logic for the `Docker Hub (PR)` workflow: only build if trigger word `#build` is found either on the PR's description (when it's first created) or if any comment is posted with such keyword, otherwise don't build an image. Comments that trigger a build automatically get a :rocket: reaction from `github-actions[bot]`.

The workflow for this is a bit cluttered, that's because a comment posted on the PR isn't associated with any branches (not even the branch where the PR is based), `master` is then assumed, so I needed to do that workaround querying GitHub's API to be able to get the PR's branch and checkout that branch to build, tag and publish the image to Docker Hub. 

---

Merges to `master` always trigger a build and the image is pushed under tags `latest` and `development`.

---

Tags on `master` that follow semantic versioning will trigger a **release build** (with optimizations) and will be tagged and published under a tag of same name.

### DIFF
--- a/.github/workflows/dockerhub-master.yml
+++ b/.github/workflows/dockerhub-master.yml
@@ -19,6 +19,6 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
           dockerfile: docker/Dockerfile
           cache_froms: buildpack-deps:stable-curl
-          repository: xain/xain-fl
+          repository: xaynetwork/xaynet
           tag_with_ref: true
           push: true
--- a/.github/workflows/dockerhub-master.yml
+++ b/.github/workflows/dockerhub-master.yml
@@ -1,0 +1,24 @@
+name: DockerHub (Master)
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-tag-push-master:
+    name: build-tag-push-master
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      
+      - name: build-tag-push
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          dockerfile: docker/Dockerfile
+          cache_froms: buildpack-deps:stable-curl
+          repository: xain/xain-fl
+          tag_with_ref: true
+          push: true
--- a/.github/workflows/dockerhub-master.yml
+++ b/.github/workflows/dockerhub-master.yml
@@ -21,4 +21,5 @@ jobs:
           cache_froms: buildpack-deps:stable-curl
           repository: xaynetwork/xaynet
           tag_with_ref: true
+          tags: development
           push: true
--- a/.github/workflows/dockerhub-pr.yml
+++ b/.github/workflows/dockerhub-pr.yml
@@ -19,6 +19,6 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
           dockerfile: docker/Dockerfile
           cache_froms: buildpack-deps:stable-curl
-          repository: xain/xain-fl
+          repository: xaynetwork/xaynet
           tag_with_ref: true
           push: true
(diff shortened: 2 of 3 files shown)


.github/workflows/dockerhub-pr.yml
```diff
@@ -0,0 +1,24 @@
+name: DockerHub (PR)
```
little-dude, line 1: Do we really need to push to dockerhub for each PR? Seems a bit overkill to me.

.github/workflows/dockerhub-release.yml
```diff
@@ -0,0 +1,24 @@
+name: DockerHub (Release)
+
+on:
+  push:
+    tags:
+      - v*
```
little-dude, line 6: So this workflow runs everytime we push a new `v*` tag? That's nice but that means we cannot really afford messing things up. On several occastion I pushed a tag, realized I forgot something, deleted the tag a pushed it again later for instance.

.github/workflows/dockerhub-pr.yml
rsaffi: I also thought it'd be overkill, but I also realized that by building an image out of a PR, any reviewer can very quickly test out that build by simply firing up a local instance that refers to that image.

I myself am fine with either having this or not, so let's also wait to hear from @Robert-Steiner as he's also a reviewer.

.github/workflows/dockerhub-release.yml
rsaffi: But in such occasion, this would be triggered twice and the latest run would be the latest image deployed to Docker Hub, so even on that occasion this would still be fine.

little-dude: Ah yes. I forgot you could overwrite images on dockerhub.

.github/workflows/dockerhub-pr.yml
Robert-Steiner: Yes, I think that could be a little too much for now. But I like the idea of fetching an image so that I can test it. Maybe we can use the [Pull Request Comment Trigger| Github action](https://github.com/marketplace/actions/pull-request-comment-trigger) to run this job on demand. wdyt?

rsaffi: I think that's a neat idea! I can apply that change.

rsaffi:
By the way @little-dude @Robert-Steiner : org and repo are now renamed, but I'll still do Docker Hub. I'll then need to update this PR, so at some point I'll need to request a new review from you guys.

.github/workflows/dockerhub-pr.yml
rsaffi: @Robert-Steiner I've been for the past 2h banging my head against the wall. The problem is: although I also like very much the idea of triggering the build upon a keyword on a comment posted to a PR, achieving that is proving to be tricky because the `actions/checkout` will insist on using `master` branch when the build is triggered by the comment. I will need to do a ridiculous workaround querying GitHub's API just so I can get the branch in which the PR is based.
I'm putting some more thought into it and thinking of alternatives.

Robert-Steiner: Yeah we can find a solution later 🙂 I'm already happy that we have an image for each merge

rsaffi: I'm insisting on it a bit, let's see...

rsaffi: I'll be damned, it works! :rocket:

.github/workflows/dockerhub-master.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: build-tag-push
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          dockerfile: docker/Dockerfile
+          cache_froms: buildpack-deps:stable-curl
+          repository: xaynetwork/xaynet
+          tag_with_ref: true
+          tags: development
```
rsaffi, line 24: This is not only tagging that image built on this merge-to-master with `latest` but also with `development`, since that's also what we use on the Development Kubernetes Cluster for the coordinator.

PR #465, opened 2020-07-16
refactor the state machine full round test

little-dude:
this makes the test more consistent with the unit tests we have for
the phases, and allows us to clean up the utils module of the various
`gen_xxx()` helpers.

### DIFF
--- a/rust/src/state_machine/tests/impls.rs
+++ b/rust/src/state_machine/tests/impls.rs
@@ -1,11 +1,14 @@
 use crate::{
     mask::MaskObject,
-    message::{MessageOwned, Payload, SumOwned, UpdateOwned},
+    message::{MessageOwned, Payload, Sum2Owned, SumOwned, UpdateOwned},
     state_machine::{
+        events::{DictionaryUpdate, MaskLengthUpdate},
         phases::{self, PhaseState},
         requests::{
             Request,
             RequestSender,
+            Sum2Request,
+            Sum2Response,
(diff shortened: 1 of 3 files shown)


rust/src/state_machine/tests/impls.rs
```diff
@@ -201,4 +221,37 @@ impl MessageOwned {
...
+        if let Payload::Sum2(Sum2Owned { mask, .. }) = &self.payload {
+            mask.clone()
+        } else {
+            panic!("not a sum2 message");
+        }
+    }
+}
+
+impl<D> DictionaryUpdate<D> {
+    pub fn unwrap(self) -> std::sync::Arc<D> {
```
little-dude, line 240: This kind of impls makes me thing we should just use `Option`s instead of defining our own enums.

little-dude:
note that this modifies the test slightly: we now have to sum participants instead of one.

rust/src/state_machine/tests/impls.rs
Robert-Steiner: Hm yeah on the one hand I like the idea. It would remove some code but on the other hand I think using the word `Invalidate` here make the code easier to understand.

janpetschexain: it's not only valuable for code understanding, but it also provides means to differentiate between "not yet available" and "already invalid", which couldn't be done with a `None`.

PR #468, opened 2020-07-17
Add model size configuration

Robert-Steiner:
**Summary**
- makes the model size value configurable
- adds `ModelSettings` to `Settings`
- fixes missing fields in the config

### DIFF
--- a/rust/src/mask/masking.rs
+++ b/rust/src/mask/masking.rs
@@ -55,12 +55,14 @@ pub enum AggregationError {
 pub struct Aggregation {
     nb_models: usize,
     object: MaskObject,
+    object_size: usize,
 }
 
 impl From<MaskObject> for Aggregation {
     fn from(object: MaskObject) -> Self {
         Self {
             nb_models: 1,
+            object_size: object.data.len(),
             object,
         }
     }
@@ -75,16 +77,17 @@ impl Into<MaskObject> for Aggregation {
 #[allow(clippy::len_without_is_empty)]
 impl Aggregation {
     /// Creates a new, empty aggregator for masks or masked models.
-    pub fn new(config: MaskConfig) -> Self {
+    pub fn new(config: MaskConfig, object_size: usize) -> Self {
         Self {
             nb_models: 0,
             object: MaskObject::new(config, vec![]),
+            object_size,
         }
     }
 
     /// Gets the length of the aggregated mask object.
     pub fn len(&self) -> usize {
-        self.object.data.len()
+        self.object_size
     }
 
     /// Gets the masking configuration of the aggregator.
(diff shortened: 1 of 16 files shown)


rust/src/mask/masking.rs
```diff
@@ -208,7 +211,7 @@ impl Aggregation {
 
         // If we have at least one object, make sure the object we're
         // trying to aggregate has the same length.
-        if self.nb_models > 0 && (self.object.data.len() != object.data.len()) {
+        if self.nb_models > 0 && (self.object_size != object.data.len()) {
```
finiteprods, line 214: perhaps the `self.nb_models > 0` check can be removed? to prevent even the first object in the aggregate from being the wrong length...

rust/src/mask/masking.rs
```diff
@@ -75,16 +77,17 @@ impl Into<MaskObject> for Aggregation {
 #[allow(clippy::len_without_is_empty)]
 impl Aggregation {
     /// Creates a new, empty aggregator for masks or masked models.
-    pub fn new(config: MaskConfig) -> Self {
+    pub fn new(config: MaskConfig, object_size: usize) -> Self {
         Self {
             nb_models: 0,
             object: MaskObject::new(config, vec![]),
+            object_size,
```
finiteprods, line 84: just a thought: could an alternative be to allocate the vec already here, e.g. `Vec::with_capacity(object_size)`? then we wouldn't need the `object_size` field anymore, I think (at the cost of an earlier allocation).

rust/src/mask/masking.rs
Robert-Steiner: Good point👍 I will remove it

rust/src/mask/masking.rs
Robert-Steiner: What we can do is:
```
object: MaskObject::new(config, Vec::with_capacity(object_size)]),
object_size,
```
but we can't remove `object_size`because capacity and length are two different things. 
Capacity is the amount of space allocated for any future elements and length is the number of actual elements within the vector. [docs.rs](https://doc.rust-lang.org/std/vec/struct.Vec.html#capacity-and-reallocation)

finiteprods: you're right, but after we set
```rust
object: MaskObject::new(config, Vec::with_capacity(object_size)),
```
later on, we can check against `self.object.data.capacity()`. we don't ever have more than `object_size` elements in `object`, so its capacity should also stay fixed. perhaps there are some downsides I haven't thought of though...

little-dude: Yeah since we never shrink or grow the vec, it _should_ be fine. But in theory, the capacity can change behind the scene so I'd rather not rely on that. That could lead to very hard to debug.

PR #469, opened 2020-07-20
purge pending requests at the end of each phase

little-dude:
Instead of adding the same code at the end of each phase, we leveraged
the existing `Phase` trait. The `Phase::next` method is now split in
two parts:

```rust
trait Phase<R> {
    async fn run(&mut self) -> Result<(), StateError>;
    fn next(self) -> Option<StateMachine<R>>;
}
```

The `run` method runs the phase, and the `next` method just consumes
the phase to produce the next one. By separating these two bits of
logic, we can inject code right after the phase ran to completion
and before the transtion is happens:

```rust
impl<R, S> PhaseState<R, S>
    where Self: Phase<R>
{
    pub async fn run_phase(mut self) -> Option<StateMachine<R>> {
           self.run().await.unwrap();
           do_custom_stuff_common_to_all_the_phases();
           self.next()
    }
}
```

In our case, `do_custom_stuff_common_to_all_the_phases()` consists in
purging the request channel. Again, this is implemented for all the
`PhaseState` via the new `Purge` trait, which is very similar to the
`Handler` trait:

```rust
pub trait Purge<R> {
    fn handle_outdated_request(&mut self, req: R);
}
```

Finally, because we needed to be able to know the current phase, we
added a bunch of `is_xxx()` methods to the `Phase` trait, that return
`true` if the state machine is in the specific phase.

### DIFF
--- a/rust/src/state_machine/phases/mod.rs
+++ b/rust/src/state_machine/phases/mod.rs
@@ -31,49 +31,29 @@ use crate::{
 use futures::StreamExt;
 use tokio::sync::oneshot;
 
+/// Name of the current phase
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum PhaseName {
+    Idle,
+    Sum,
+    Update,
+    Sum2,
+    Unmask,
+    Error,
+    Shutdown,
+}
+
 /// A trait that must be implemented by a state in order to move to a next state.
 #[async_trait]
 pub trait Phase<R> {
-    /// Moves from this state to the next state.
-    fn next(self) -> Option<StateMachine<R>>;
+    /// Name of the current phase
+    const NAME: PhaseName;
 
     /// Run this phase to completion
     async fn run(&mut self) -> Result<(), StateError>;
 
-    /// Return `true` if this is the error phase
-    fn is_error(&self) -> bool {
-        false
-    }
-
-    /// Return `true` if this is the idle phase
-    fn is_idle(&self) -> bool {
-        false
-    }
-
-    /// Return `true` if this is the shutdown phase
-    fn is_shutdown(&self) -> bool {
-        false
-    }
-
-    /// Return `true` if this is the sum phase
-    fn is_sum(&self) -> bool {
-        false
-    }
-
-    /// Return `true` if this is the sum2 phase
-    fn is_sum2(&self) -> bool {
-        false
-    }
-
-    /// Return `true` if this is the update phase
-    fn is_update(&self) -> bool {
-        false
-    }
-
-    /// Return `true` if this is the unmask phase
-    fn is_unmask(&self) -> bool {
-        false
-    }
+    /// Moves from this state to the next state.
+    fn next(self) -> Option<StateMachine<R>>;
 }
 
 /// A trait that must be implemented by a state to handle a request.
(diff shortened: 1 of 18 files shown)


rust/src/state_machine/phases/mod.rs
```diff
@@ -96,6 +159,43 @@ where
...
+impl<R, S> PhaseState<R, S>
+where
+    Self: Phase<R> + Purge<R>,
+{
+    /// Run the current phase to completion, then transition to the
+    /// next phase and return it.
+    pub async fn run_phase(mut self) -> Option<StateMachine<R>> {
+        if let Err(err) = self.run().await {
+            return Some(self.into_error_state(err));
+        } else {
```
Robert-Steiner, line 171: Was it intended to leave the "else" here?

rust/src/state_machine/phases/mod.rs
```diff
@@ -35,7 +35,45 @@ use tokio::sync::oneshot;
 #[async_trait]
 pub trait Phase<R> {
     /// Moves from this state to the next state.
-    async fn next(mut self) -> Option<StateMachine<R>>;
+    fn next(self) -> Option<StateMachine<R>>;
+
+    /// Run this phase to completion
+    async fn run(&mut self) -> Result<(), StateError>;
+
+    /// Return `true` if this is the error phase
```
Robert-Steiner, line 43: I think we can add a `PhaseStates` enum an add a `phase` field to the CoordinatorState. It's somewhat redundant, of course, but we may need this information (in which phase the coodrinator was in) later if we want to restore the coordinator state.

rust/src/state_machine/phases/mod.rs
little-dude: Somehow I thought rustc would reject the code without the `else` but it seems accepted. I'll remove the `else`.

little-dude:
> We are at the end of the Sum phase. The sum dict was published (to the fetcher service) and is now available to the update clients. Now, the state machine tries to clear the channel, which takes lets say about 30 seconds. At the same time, an update participant sends an update message, which however, is rejected by the PetMessages service because the state machine hasn't published the PhaseEvent::Update yet.

True in theory, but I don't think handling the pending requests at the end of the phase takes long in practice (a few ms per request maybe?), so I'm not too worried about it for the time being.

rust/src/state_machine/phases/mod.rs
little-dude: You'd prefer that to the current `is_xxx` methods? As you said it'd be redundant so I'm not really convinced.

Robert-Steiner: difficult to decide. I think it would be a bit cleaner, but on the other hand, we can already provide information about the state of the state machine without introducing an enum. 
I think using the `is_xxx` methods is fine

little-dude: Let's see if @finiteprods and @janpetschexain have an opinion on that matter.

janpetschexain: i personally would take an enum, but the `is_x` state functions are ok as well.

little-dude: You guys were right: it looks much better with 859c5b0

Robert-Steiner: Looks nice👍

PR #471, opened 2020-07-20
Mobile client

Robert-Steiner:
Not done yet but I can already be reviewed. 

What is done:
- [Implement the participant state machine](https://github.com/xaynetwork/xaynet/pull/471/files#diff-d21dc7d1214507400ec1f1b4ddc3b603)
  - The participant state machine keeps the state of the participant
- [Implement the client state machine](https://github.com/xaynetwork/xaynet/pull/471/files#diff-2aebda4299965a3ba558201a243a3e51)
  - The client state machine keeps the participant state machine, proxy, local/global model and has the logic to perform requests.
- [Make the participant configurable](https://github.com/xaynetwork/xaynet/pull/471/files#diff-d21dc7d1214507400ec1f1b4ddc3b603)
  - The secret key, the certificate and the maskconfig should be configurable.
- [Make the client configurable](https://github.com/xaynetwork/xaynet/pull/471/files#diff-2aebda4299965a3ba558201a243a3e51)
  - The url should be configurable.
- [Implement the mobile client](https://github.com/xaynetwork/xaynet/pull/471/files#diff-19e28fc87a19597a5a70c6206a1e52c4)
  - The mobile client contains the client state, the tokio runtime and the local/global model.
- [a runnable example](https://github.com/xaynetwork/xaynet/pull/471/files#diff-3c38ca016ceb77a8ff7c2eb31b40a4af)

I actually wanted to create a PR for each module `Participant`, `ClientState` and `MobileClient`. However, when I implemented the `ClientState` struct, I had to change a lot on the participant struct and it got a bit messy. For this reason, I packed everything in one PR. If you want I can also create a new PR for each module. 

My suggestion would be that we keep the old client/participant for now and just add the mobile client to the repository and first test if the mobile approach works


rust/src/client/mobile_client/participant/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+pub mod sum;
+pub mod sum2;
+pub mod undefined;
+pub mod update;
+
+pub use self::{sum::Sum, sum2::Sum2, undefined::Undefined, update::Update};
+
+pub struct ParticipantState {
+    // credentials
+    pub pk: ParticipantPublicKey,
```
little-dude, line 25: Maybe we could use the `EncryptKeyPair` for this now.

rust/src/client/mobile_client/participant/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+            pk: secret_key.public_key(),
+            sk: secret_key,
+            mask_config,
+            certificate,
+        }
+    }
+}
+
+#[derive(From)]
+pub enum Type {
```
little-dude, line 57: Maybe `Role`?

rust/src/client/mobile_client/client.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    async fn next(mut self) -> ClientStateMachine {
+        info!("selected to sum");
+
+        match self.run().await {
+            Ok(_) => self.move_into_sum2().into(),
+            Err(ClientError::RoundOutdated) => self.reset().into(),
+            Err(err) => {
+                error!("{:?}", err);
+                self.into()
```
little-dude, line 198: I think we need a timer for the error case. Because if I understand the implementation correctly, if some resource is not available yet, the client we'll keep querying it and fail with `TooEarly` in a sort of busy loop. On the longer run we might even want to have a more sophisticated retry mechanism with exponential backoff or something.

rust/src/bin/test-mobile.rs
```diff
@@ -0,0 +1,72 @@
+extern crate tracing;
```
little-dude, line 1: I think we should start putting those in a `rust/examples` directory. The `src/bin` directory should contain code we don't intend to ship.

rust/src/client/mobile_client/client.rs
Robert-Steiner: I agree that is a good idea.
Another idea I had is to add caching to make the client more efficient. At the moment we always request all values even if we already got a value in the previous call of `next()`.

rust/src/client/mobile_client/participant/mod.rs
Robert-Steiner: `Role` sounds good

rust/src/client/mobile_client/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    pub fn set_local_model(&self, local_model: Model) {
+        *self.local_model.borrow_mut() = Some(local_model);
+    }
+
+    pub fn get_global_model(&self) -> Option<Model> {
+        self.global_model.borrow().clone()
+    }
```
janpetschexain, line 50: we should either have a `# Panics` section in those method docs or switch to `try_borrow_mut()`/`try_borrow()` and add some error handling.

rust/src/client/mobile_client/client.rs
```diff
@@ -0,0 +1,364 @@
...
+    state_machine::coordinator::RoundParameters,
+    InitError,
+    PetError,
+};
+use derive_more::From;
+use std::{cell::RefCell, rc::Rc};
+
+#[derive(From)]
+pub enum ClientStateMachine {
+    Undefined(ClientState<Undefined>),
```
janpetschexain, line 26: `Undefined` seems a bit odd to me, it is actually well defined by those very lines of code^^. what do you think of `Pending` or `Dangling` or sth similar?

rust/src/client/mobile_client/client.rs
```diff
@@ -0,0 +1,364 @@
...
+            Type::Unselected(unsel_par) => {
+                info!("unselected");
+                ClientState::<Undefined>::new(proxy, unsel_par.reset(), local_model, global_model)
+                    .into()
+            }
+            Type::Summer(sum_par) => {
+                ClientState::<Sum>::new(proxy, round_params, sum_par, local_model, global_model)
+                    .into()
+            }
+            Type::Updater(upt_pat) => {
```
janpetschexain, line 139:
```suggestion
            Type::Updater(upt_par) => {
```
or maybe just `Type::...(participant)` for all three bindings.

rust/src/client/mobile_client/participant/mod.rs
janpetschexain: `SigningKeyPair`

rust/src/bin/test-mobile.rs
```diff
@@ -0,0 +1,72 @@
...
+fn main() -> Result<(), ()> {
+    let _fmt_subscriber = FmtSubscriber::builder()
+        .with_env_filter(EnvFilter::from_default_env())
+        .with_ansi(true)
+        .init();
+
+    // crucial: init must be called before anything else in this module
+    sodiumoxide::init().unwrap();
+
+    let SigningKeyPair { public: _, secret } = SigningKeyPair::generate();
```
janpetschexain, line 50:
```suggestion
    let SigningKeyPair { secret, .. } = SigningKeyPair::generate();
```

rust/src/client/mobile_client/participant/mod.rs
Robert-Steiner: Ah true, I will change it

Robert-Steiner: Shall we do the same with        

```
ephm_pk: SumParticipantEphemeralPublicKey,
ephm_sk: SumParticipantEphemeralSecretKey,
```

in `Participant<Sum>` and `Participant<Sum2>` as well?

Robert-Steiner:
@finiteprods @little-dude @janpetschexain
ready for review🙂

changes since the last review

- implemented the function `new_secret_key` to make it easier for a user to generate a new secret key
- implemented the FFI (currently, it is living in a separate repository)
- made the rustls feature of reqwest configurable on the xaynet crate (we had issues compiling reqwest on android due to openssl therefore we switched to rustls)
- moved the mobile client example into the `example` folder

Note: 
The new participant/client struct and the new mobile client are still in an early stage. For that reason, the documentation of the mobile client is hidden. However, the new participant/client struct is a starting point for a unified client state machine that works on a pc, web(wasm) and mobile.

rust/src/client/mobile_client/client.rs
```diff
@@ -0,0 +1,364 @@
...
+            ClientStateMachine::Sum2(state) => state.next().await,
+        }
+    }
+}
+
+pub struct ClientState<Type> {
+    proxy: Proxy,
+    round_params: RoundParameters,
+    participant: Participant<Type>,
+    local_model: Rc<RefCell<Option<Model>>>,
```
little-dude, line 65: Why do we need these to be wrapped into `Rc`? In `run()` and `fetch_global_model()`, where these two fields are used, it seems that we could do without it. Removing the `Rc<RefCell>` would be nice if we want to share this code with the "regular" multi-threaded client, because `ClientState` would be `Send`.

rust/src/client/mobile_client/client.rs
```diff
@@ -0,0 +1,364 @@
...
+        } = self;
+
+        let participant_type = participant.determine_role(
+            round_params.seed.as_slice(),
+            round_params.sum,
+            round_params.update,
+        );
+
+        match participant_type {
+            Role::Unselected(unsel_par) => {
```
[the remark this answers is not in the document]
Robert-Steiner: fixed

rust/src/client/mobile_client/client.rs
Robert-Steiner: That is a good point! Do you have something in mind how we could solve this?
Two solutions that I can think of right now are:

1. Replace `local`/`global` model with a separate component in the `ClientState` that is responsible for `get_local_model` and `set_global_model`. We might even define a trait here and have different implementations for each environment (web, mobile, ...). 
For example:
for the mobile environment we might need to store it on the file system but on a desktop environment it might be ok to keep it in memory. 
 
2. We just add the local/global model to the client state.

little-dude: I think on the long run #1 is the best solution. But this PR is already pretty massive so I think #2 is a good enough first implementation. The way I see it, we can "just" remove the `Rc<RefCell<>>` wrapping and adjust the code wherever rustc complains. But I haven't tried it so I may be missing something.

rust/src/client/mobile_client/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+mod client;
+use self::client::ClientStateMachine;
+
+pub mod participant;
+use self::participant::ParticipantSettings;
+
+pub struct MobileClient {
+    runtime: tokio::runtime::Runtime,
+    client_state: Option<ClientStateMachine>,
```
janpetschexain, line 15: are we ever in a situation where the `client_state` is `None`? the `new()` and `next()` methods always guarantee a `Some(_)` and none of the other methods are `take()`ing this. the `Option<_>`s for the models can still be checked because they are optional within the state as well. hence this could simplify the code a bit.

rust/src/client/mobile_client/client.rs
Robert-Steiner: Alright I removed the `Rc<RefCell<>>` and moved the `local` and `global` model into the `ClientState` struct. Maybe not the most elegant solution but I think it is good for now but I'm open to change it to something different.

rust/src/client/mobile_client/mod.rs
Robert-Steiner: Yeah I tried to remove the optional, but I got some ownership issues in the `next` method because of the move of the client into the tokio runtime.

little-dude: For reference:

```rust
    pub fn next(&mut self) {
        if let Some(current_state) = self.client_state.take() {
            let new_state = self
                .runtime
                .block_on(async move { current_state.next().await });
            self.client_state = Some(new_state)
        }
    }
```

I don't see an alternative to `Option` here indeed. The only way around that would be for `next` signature to be changed to `fn next(self) -> Self` but that has ramifications in other places.

janpetschexain: i see, we would need to clone the state for the `async move`, which makes `take()` the better option. thanks for the explanation 👍

little-dude: Note that `block_on` doesn't require the future to be `Send + 'static`. The reason for using `move` here is that `ClientStateMachine::next` takes `self`. So we actually have two solutions if we want to remove that extra `Option`

- make `MobileClient::next` take `self` and return `Self` (as I suggested above)
- make `ClientStateMachine::next` take `&mut` and use `async` instead of `async move`

Regardless of the `Option` discussion, it would be nice if the `next` methods had the same signature btw. I'm just not sure how hard that would be.

Robert-Steiner: Do you mean with `next` methods the methods of each client variant (mobile, web, ..)?

little-dude: Well we have:

- `ClientStateMachine::next`: `async fn next(self) -> Self`
- `ClientState<T>::next`: `async fn next(self) -> ClientStateMachine`
- `MobileClient::next`: `fn next(&mut self)`
- `Participant<Sum>::next`: `fn next(self) -> Participant<Sum2>`

I'm saying it would be nice if these were a little more consistent, since they have the same name. In particular, making `MobileClient::next` an `fn next(self) -> Self` would be more consistent with `ClientStateMachine::next` and `ClientState<T>::next`, _and_ probably allow us to get rid of the `Option` @janpetschexain mentionned.

As a sidenote: `Participant<Sum>::next` could maybe be renamed into something like `into_sum2` or maybe we could `impl Into<Participant<Sum2>> for Participant<Sum>`? We're getting too many `next` methods.

Robert-Steiner:
I removed the `Option`, implemented `impl Into<Participant<Sum2>> for Participant<Sum>` and renamed the method `next` of the mobile client to `perform_task`

@little-dude 
I agree, it's okay if I spuash the PR?

little-dude:
Nice, I think this is an improvement :+1: 

> @little-dude I agree, it's okay if I squash the PR?

Sure or just splitting it in a bunch of self-contained commits works as well :+1:

PR #474, opened 2020-07-21
automatically send the phase event before the phase starts

little-dude:
this started as a small improvement: instead of repeating the same
snippet for each phase, we wanted to do it once in
`PhaseState<R,S>::run_phase()`:

```rust
pub async fn run_phase(mut self) -> Option<StateMachine<R>> {
    let phase = <Self as Phase<R>>::NAME;
    self.coordinator_state.events.broadcast_phase(phase); // <-- add this
    self.run().await.unwrap();
    self.purge_outdated_requests().unwrap();
    self.next()
}
```

It turns out that this doesn't work for the idle phase, because the
round ID is updated in the `run()` implementation, _after_ the phase
event is broadcasted. Therefore, the event that is emitted has the
wrong round id.

There are multiple solutions:

  1. re-emit an event in the idle phase. That may work but that means
     there's a small interval of time during which the listeners still
     have an inconsistent Phase event.

  2. add some logic before broadcasting the phase, such that if this
     is the idle phase, update the round seed before anything
     else. But this logic really belong to the phase itself, so that
     feels very hacky

  3. generate the round seed in `PhaseState<R, Idle>::new`. But
     generating a round seed here is again a bit hacky. This should
     happen when the phase _runs_, in the `Handle::run` impl.

Therefore we went for a 4th solution, and introduced a round id that
is only used internally and has nothing to do with the round
parameters. Updating it in `PhaseState<R, Idle>::new` is fine, since
it has nothing to do with the _phase_ logic itself. The round
parameters are still fully taken care of in `Handle::run`.

Interestingly, this has opened the door to some simplifications in the
`EventPublisher` logic, and we were able to make its API simpler by
removing the round id from the methods arguments altogether.

### DIFF
--- a/rust/src/services/tests/fetchers.rs
+++ b/rust/src/services/tests/fetchers.rs
@@ -42,13 +42,12 @@ async fn test_mask_length_svc() {
     let resp = task.call(MaskLengthRequest).await;
     assert_eq!(resp, Ok(None));
 
-    let round_id = subscriber.params_listener().get_latest().round_id;
-    publisher.broadcast_mask_length(round_id.clone(), MaskLengthUpdate::New(42));
+    publisher.broadcast_mask_length(MaskLengthUpdate::New(42));
     assert_ready!(task.poll_ready()).unwrap();
     let resp = task.call(MaskLengthRequest).await;
     assert_eq!(resp, Ok(Some(42)));
 
-    publisher.broadcast_mask_length(round_id, MaskLengthUpdate::Invalidate);
+    publisher.broadcast_mask_length(MaskLengthUpdate::Invalidate);
     assert_ready!(task.poll_ready()).unwrap();
     let resp = task.call(MaskLengthRequest).await;
     assert_eq!(resp, Ok(None));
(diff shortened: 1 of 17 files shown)


rust/src/services/tests/fetchers.rs
```diff
@@ -110,13 +108,12 @@ async fn test_scalar_svc() {
     let resp = task.call(ScalarRequest).await;
     assert_eq!(resp, Ok(None));
 
-    let round_id = subscriber.params_listener().get_latest().round_id;
-    publisher.broadcast_scalar(round_id.clone(), ScalarUpdate::New(42.42));
+    publisher.broadcast_scalar(ScalarUpdate::New(42.42));
```
janpetschexain, line 111: this is violating the masking configuration, the scalar is truncated by the masking method to fit the config, but the coordinator shouldn't broadcast invalid values.
```suggestion
    publisher.broadcast_scalar(ScalarUpdate::New(0.42));
```

rust/src/state_machine/coordinator.rs
```diff
@@ -32,6 +32,8 @@ pub struct RoundParameters {
 pub struct CoordinatorState {
     /// The credentials of the coordinator.
     pub keys: EncryptKeyPair,
+    /// Internal ID used to identify a round
```
janpetschexain, line 35:
```suggestion
    /// Internal ID used to identify a round.
```

rsaffi:
#build

rust/src/state_machine/phases/idle.rs
```diff
@@ -98,7 +77,12 @@ where
 
 impl<R> PhaseState<R, Idle> {
     /// Creates a new idle state.
-    pub fn new(coordinator_state: CoordinatorState, request_rx: RequestReceiver<R>) -> Self {
+    pub fn new(mut coordinator_state: CoordinatorState, request_rx: RequestReceiver<R>) -> Self {
+        // Since some events are emitted very early, the round id must
+        // be correct when the idle phase starts. Therefore, we update
+        // it here, when instantiating the idle PhaseState.
+        coordinator_state.set_round_id(coordinator_state.round_id());
```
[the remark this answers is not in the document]
little-dude: Doh you're right. Definitely not enough tests.

PR #483, opened 2020-07-30
XN-988 Check documentation build

rsaffi:


.github/workflows/rust.yml
```diff
@@ -218,3 +218,20 @@ jobs:
...
+
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: Check the building of docs
+        working-directory: ./rust
+        run: cargo doc --all-features --color always
```
janpetschexain, line 237: this should fail on our current documentation if we don't use the nightly toolchain, because auto link generation is not yet stable.

```suggestion
        run: cargo +nightly doc --all-features --color always
```

.github/workflows/rust.yml
```diff
@@ -218,3 +218,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
```
rsaffi, line 233: @janpetschexain But then we should probably also adjust here, to install the nightly toolchain, shouldn't we? Or we can still install the stable toolchain and just call `cargo doc` with the +nightly flat?

janpetschexain: we have to install nightly, similar to the `cargo-fmt` job.

.github/workflows/rust.yml
```diff
@@ -218,3 +218,20 @@ jobs:
...
         run: docker-compose down
+
+  docs:
+    name: cargo-doc
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install stable toolchain
```
Robert-Steiner, line 229:
```suggestion
      - name: Install nightly toolchain
```

PR #484, opened 2020-07-30
getting started with examples guide

finiteprods:
The project is currently lacking examples for a newcomer to get started. It is also lacking a guide that goes further than what is in the README.

This merge request begins to address this, by making the `test-drive` (which we use for testing, internally) a first example. In a little more detail, we have:

- a new `examples` directory - the new home for the `test-drive` and future example targets.
- an accompanying Getting Started guide, in `src/examples.rs`. This adds a new module (`examples`) which is just pure documentation.
- a small update to the README to reference the above.

Run `cargo doc` to build the Getting Started guide.

As a side-note: as a result of this reorganisation, we should now
```
$ cargo run --example test-drive-net
```
(instead of `--bin` as before).


README.md
```diff
@@ -156,6 +156,9 @@ running on `127.0.0.1:8081`:
 RUST_LOG=xaynet=info cargo run --bin test-drive-net -- -n 20 -u http://127.0.0.1:8081
 ```
 
+For more in-depth details on how to run examples, see the accompanying Getting
+Started guide under [rust/src/examples.rs](./rust/src/examples.rs).
```
little-dude, line 160: Why putting that in `src/`? We could just add this to the examples themself in `rust/examples`.

finiteprods: that was my original thought too, but it seems `cargo doc` doesn't render docs under `examples` (I saw an open issue for addressing that). So instead, it's a module here which is just pure-documentation, which does at least get rendered along with the rest. I actually got this idea from the [`csv`](https://docs.rs/csv/1.0.0-beta.5/csv/index.html) crate (which has pure-documentation modules `cookbook` and `tutorial`).

rust/src/examples.rs
```diff
@@ -0,0 +1,214 @@
...
+This is essentially what is meant by federated learning that is
+*privacy-preserving*, and is a key feature enabled by the PET protocol.
+
+## PET Protocol
+
+It is worth describing the protocol very briefly here, if only to better
+understand some of the configuration settings we will meet later. It is helpful
+to think of each round being divided up into several contiguous phases:
+
+**Start.** At the start of a round, the coordinator generates a collection of random *round
```
janpetschexain, line 49: not sure if we should deviate from the internal phase naming, **start** makes sense from a narrative point of view, but the connection to **idle** might not be directly obvious.

rust/src/examples.rs
```diff
@@ -0,0 +1,214 @@
...
+group type, data type, bound type and model type. The [`ModelSettings`] specify
+the size of the model used. Both of these settings should be decided in advance
+with participants, and agreed upon by both.
+
+## Running
+
+The coordinator can be run as follows:
+
+```ignore
+$ git clone git://github.com/xaynetwork/xaynet
```
janpetschexain, line 146: for whatever reason the double slashes cause the following part of the url to be rendered in grey like a comment. not sure if we can do sth about it, maybe a different interpreter than `ignore` works here?

rust/src/examples.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    Ok(())
+}
+```
+
+As this example is meant solely as a demonstration of the operation of
+the PET protocol, it does not perform any training as such (see other
+forthcoming examples for that). Instead, a "dummy" zero-model is used throughout.
+Nevertheless, note that its length and contents are (respectively) required to
```
janpetschexain, line 198:
```suggestion
Nevertheless, note that its length and contents, respectively, are required to
```

rust/src/examples.rs
```diff
@@ -0,0 +1,214 @@
+/*!
```
janpetschexain, line 1: there are some broken links in this file:
```
warning: `[ApiSettings]` cannot be resolved, ignoring it.
   --> src/examples.rs:112:26
    |
112 | Going from the top, the [`ApiSettings`] include [`bind_address`], which is the
    |                          ^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = note: `#[warn(intra_doc_link_resolution_failure)]` on by default
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[bind_address]` cannot be resolved, ignoring it.
   --> src/examples.rs:112:50
    |
112 | Going from the top, the [`ApiSettings`] include [`bind_address`], which is the
    |                                                  ^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[PetSettings]` cannot be resolved, ignoring it.
   --> src/examples.rs:116:6
    |
116 | The [`PetSettings`] specify various parameters of the PET protocol.
    |      ^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[sum]` cannot be resolved, ignoring it.
   --> src/examples.rs:119:16
    |
119 | important are [`sum`] and [`update`], which are the probabilities assigned to
    |                ^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[update]` cannot be resolved, ignoring it.
   --> src/examples.rs:119:28
    |
119 | important are [`sum`] and [`update`], which are the probabilities assigned to
    |                            ^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[min_sum_count]` cannot be resolved, ignoring it.
   --> src/examples.rs:123:17
    |
123 | - The settings [`min_sum_count`] and [`min_update_count`] specify, respectively,
    |                 ^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[min_update_count]` cannot be resolved, ignoring it.
   --> src/examples.rs:123:39
    |
123 | - The settings [`min_sum_count`] and [`min_update_count`] specify, respectively,
    |                                       ^^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[min_sum_time]` cannot be resolved, ignoring it.
   --> src/examples.rs:128:32
    |
128 | - To complement, the settings [`min_sum_time`] and
    |                                ^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[min_update_time]` cannot be resolved, ignoring it.
   --> src/examples.rs:129:2
    |
129 | [`min_update_time`] specify, respectively, the minimum amount of time (in
    |  ^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[expected_participants]` cannot be resolved, ignoring it.
   --> src/examples.rs:133:4
    |
133 | - [`expected_participants`] should be set to the expected number of participants
    |    ^^^^^^^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[MaskSettings]` cannot be resolved, ignoring it.
   --> src/examples.rs:136:6
    |
136 | The [`MaskSettings`] determines the masking configuration, consisting of the
    |      ^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[ModelSettings]` cannot be resolved, ignoring it.
   --> src/examples.rs:137:56
    |
137 | group type, data type, bound type and model type. The [`ModelSettings`] specify
    |                                                        ^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`

warning: `[Client]` cannot be resolved, ignoring it.
   --> src/examples.rs:201:66
    |
201 | The example creates the expected number of participants (called [`Client`]s
    |                                                                  ^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`
```

rust/src/examples.rs
```diff
@@ -0,0 +1,214 @@
...
+here), connecting to the address where the coordinator should be listening. The
+participants are then started, and run continuously until stopped (with CTRL +
+C).
+
+## Running
+
+The actual `test-drive-net` example is a tidier version of the above, where the
+hard-coded numbers are made configurable. To run:
+
+```ignore
```
Robert-Steiner, line 211:
```suggestion
```text
```
I think you can use `text` here. That cloud fix the problem with the CI

rust/src/examples.rs
finiteprods: I considered this too. But calling it Idle here, I felt might raise a question mark for the reader since the name is only really meaningful together with the context of the state machine, which is not mentioned in this guide. and as the other states aren't explicitly mentioned either, I felt just saying **Start** would be more intuitive here.

rust/src/examples.rs
finiteprods: hopefully @Robert-Steiner 's suggestion sorts this out

rust/src/examples.rs
finiteprods: added the links

rust/src/examples.rs
finiteprods: thanks! this does work-around the smoke-test complaint too

rust/src/examples.rs
janpetschexain: makes sense.

PR #486, opened 2020-07-31
XN-976 CI improvement: code test coverage

rsaffi:


.github/workflows/rust.yml
```diff
@@ -235,3 +235,26 @@ jobs:
...
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+
+      - name: Run cargo-tarpaulin
+        uses: actions-rs/tarpaulin@v0.1.2
+        with:
+          args: '--manifest-path rust/Cargo.toml --all-features --force-clean --lib --exclude-files */tests/* src/vendor/* *.h'
```
little-dude, line 255: Why do we need to specify the `*.h` files? Also I _think_ that to correctly ignore the test code we need `--ignore-tests`. 
```suggestion
          args: '--manifest-path rust/Cargo.toml --all-features --force-clean --lib --ignore-tests --exclude-files src/vendor/*'
```

Not 100% sure though

rsaffi:
> +1 this is great. Could we just squash before merging? grin

Definitely yes. Already intended to do so, due to the many test commits to trigger the CI! :laughing:

PR #487, opened 2020-08-03
add metrics service

Robert-Steiner:
First part of the `Implement data collection to InfluxDB` task.
This PR focus on the definition of the influx data models and the implementation of the metrics service + sender. The integration of the metrics service into the state machine is addressed in a separate PR.  

**metrics/mod.rs**

The `mod.rs` file contains the `MetricsService`, the `MetricsSender` and the functions to generate metrics. 
The `MetricsSender` communicates with the `MetricsService` via an unbounded channel. The unbounded channel 
accepts the type `WriteQuery`. The `WriteQuery` is a struct of the `influxdb` crate. 

My actual plan was to build a wrapper around that struct like:

```rust

enum Metric {
	RoundTotalNumber(RoundTotalNumber);
	Phase(Phase);
}

receiver: UnboundedReceiver<Metric>,
```

However, the `#[derive(InfluxDbWriteable)]` cannot be applied on enums. I would have had to implement the method `into_query` on each variant. So I decided against it. 

I tried to find a solution to encode some properties/information of the metrics into the code. I first started with a couple of structs but that didn't work that well. In the end I used the module system.
Of course, I could have implemented it more generic and save some lines of code. However, I think the abstraction makes it easier to use and less error-prone. 

Structure:

```rust
    use crate::metrics;
    metrics::round_parameters::sum::update(0.3);
    metrics::round_parameters::update::update(0.3);
    metrics::phase::update(PhaseName);
    metrics::phase::error::emit(StateError);
    metrics::masks::total_number::update(1);
    metrics::round::total_number::update(1);
    metrics::round::successful::increment();
    metrics::message::sum::increment(11, PhaseName);
    metrics::message::update::increment(21, PhaseName);
    metrics::message::sum2::increment(31, PhaseName);
    metrics::message::discarded::increment(41, PhaseName);
    metrics::message::rejected::increment(41, PhaseName);
```

The `models.rs` file only contains the influx data models. Through `InfluxDbWriteable` the structs are automatically implement the `into_query` methods.

### DIFF
--- a/rust/src/metrics/mod.rs
+++ b/rust/src/metrics/mod.rs
@@ -216,15 +216,24 @@ pub struct MetricsService {
 }
 
 impl MetricsService {
-    pub fn new(host: &str, db_name: &str) -> (MetricsService, MetricsSender) {
+    pub fn new(url: &str, database: &str) -> (MetricsService, MetricsSender) {
+        let client = Client::new(url, database);
+        Self::new_metrics_service(client)
+    }
+
+    pub fn new_with_auth(
+        url: &str,
+        database: &str,
+        username: &str,
+        password: &str,
+    ) -> (MetricsService, MetricsSender) {
+        let client_auth = Client::new(url, database).with_auth(username, password);
+        Self::new_metrics_service(client_auth)
+    }
+
+    fn new_metrics_service(client: Client) -> (MetricsService, MetricsSender) {
         let (sender, receiver) = unbounded_channel();
-        (
-            MetricsService {
-                client: Client::new(host, db_name),
-                receiver,
-            },
-            MetricsSender(sender),
-        )
+        (MetricsService { client, receiver }, MetricsSender(sender))
     }
 }
 
--- a/rust/src/metrics/mod.rs
+++ b/rust/src/metrics/mod.rs
@@ -4,15 +4,15 @@ use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
 mod models;
 
 pub mod round_parameters {
-    use super::models::{Measurement, Sum, Update};
+    use super::models::{Measurement, RoundParamSum, RoundParamUpdate};
     use influxdb::{InfluxDbWriteable, Timestamp, WriteQuery};
     pub mod sum {
         use super::*;
 
         pub fn update(sum: f64) -> WriteQuery {
-            Sum {
+            RoundParamSum {
                 time: Timestamp::Now.into(),
-                sum,
+                round_param_sum: sum,
             }
             .into_query(Measurement::StateMachine.to_string())
         }
@@ -22,9 +22,9 @@ pub mod round_parameters {
         use super::*;
 
         pub fn update(update: f64) -> WriteQuery {
-            Update {
+            RoundParamUpdate {
                 time: Timestamp::Now.into(),
-                update,
+                round_param_update: update,
             }
             .into_query(Measurement::StateMachine.to_string())
         }
@@ -119,7 +119,7 @@ pub mod message {
         pub fn increment(round_id: u64, phase: PhaseName) -> WriteQuery {
             MessageSum {
                 time: Timestamp::Now.into(),
-                sum: 1,
+                message_sum: 1,
                 round_id,
                 phase: phase as u8,
             }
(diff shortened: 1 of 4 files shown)


Robert-Steiner:
> So instead, we ended up with lots of little `#[derive(InfluxDbWriteable)]` structs, one for each metric, I guess?

The idea of the enum was to have another layer of abstraction. 
Currently we use the type `WriteQuery` which is a type defined in the `influxdb` crate.
The issue is that we don't have control over this type and that leads to some unnecessary complications. 
You can see one example in the tests. Since the fields of `WriteQuery` are private, 
we have to call `build` to actually get something back that we can compare in the assert macro.

Therefore my idea was to introduce another layer of abstraction via an enum.
```
enum Metrics {
	RoundParamSum(RoundParamSum),
	...
}
```

We would still need the structs because they define how the data is transformed into influx data model.

```
#[derive(InfluxDbWriteable)]
pub struct MessageRejected {
    pub time: DateTime<Utc>, 	// timestamp
    pub message_rejected: u8,   // field
    #[tag] 
    pub round_id: u64,      	  // tag
    #[tag]
    pub phase: u8, 	        // tag
}
```

The `into_query` method will transform this data into the influx protocol syntax.

```
state_machine,round_id=1,phase=0 message_rejected=1 1465839830100400200
  |           ------------------ ------------------ -------------------
  |               |                       |                 |
  |               |                       |                 |
+------------+--------+--------------+-----------------+---------+
|measurement |,tag_set|              |field_set|       |timestamp|
+------------+--------+--------------+-----------------+---------+
```
 
But instead of sending a `WriteQuery` we would send a variant of the `Metrics` enum

```diff
- pub struct MetricsSender(UnboundedSender<WriteQuery>);
+ pub struct MetricsSender(UnboundedSender<Metrics>);
```

and call the method `into_query` on the receiver half of the channel (the `run_metric_service` function).

```diff
pub async fn run_metric_service(mut metics_service: MetricsService) {
    loop {
        match metics_service.receiver.recv().await {
-            Some(write_query) => { 
+            Some(metric) => { 			              <---- Type Metrics
                let _ = metics_service
                    .client
-                    .query(&write_query)
+                    .query(&metric.into_query())         <---- call into_query
                    .await
                    .map_err(|e| error!("{}", e));
            }
            None => {
                warn!("All senders have been dropped!");
                return;
            }
        }
    }
}
```

However for this to work, we need to implement `into_query` on the enum or we need to use the match pattern in the `run_metric_service` function.

```rust 

impl Metrics {
	fn into_query(self) -> WriteQuery {
		match metric {
			RoundParamSum(metric) => metric.into_query()
			...
		}
	}
}

// or

pub async fn run_metric_service(mut metics_service: MetricsService) {
    loop {
        match metics_service.receiver.recv().await {
            Some(metric) => { 			      
            
                   
            	  let query = match metric {
	             RoundParamSum(metric) => metric.into_query(),
		     ...
	          };
					
					
                let _ = metics_service
                    .client
                    .query(&query())        
                    .await
                    .map_err(|e| error!("{}", e));
            }
            None => {
                warn!("All senders have been dropped!");
                return;
            }
        }
    }
}
```

In both cases we need to alter the match pattern manually once we add/change or remove a metric which is quite annoying I think. 
Therefore I used `WriteQuery` for now.
But my plan is, to create a PR on the influxdb crate which implements `#[derive(InfluxDbWriteable)]` for enums.

rust/src/metrics/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+                message_rejected: 1,
+                round_id,
+                phase: phase as u8,
+            }
+            .into_query(Measurement::StateMachine.to_string())
+        }
+    }
+}
+
+pub async fn run_metric_service(mut metics_service: MetricsService) {
```
little-dude, line 187: One downside of this implementation is that we process the request one by one. If the queries are slow, the queries will accumulate in the channel. Maybe we could make that channel bounded to apply some backpressure. On the longer run, we could use tower to benefit from timeouts, backpressure and concurrent requests.

Robert-Steiner:
@finiteprods @little-dude thanks for the review👍

@little-dude I agree that is a good idea to use `bounded` channels.  
However, I have one question. When I use a `bounded` channel I need to use` try_send` instead of `send` otherwise the metrics channel can block the execution of the state machine, right?

little-dude:
Yes exactly

PR #488, opened 2020-08-03
Integrate metrics sender into state machine

Robert-Steiner:
The second part of the `Implement data collection to InfluxDB` task.

**Integration of the MetricsSender into the state machine.**

My first idea was to put the metrics sender into the CoordinatorState so that we don't have to add another field to the `PhaseState`. However, I didn't like it because the metrics sender has nothing to do with the coordinator state. Therefore, I ended up putting the metrics sender into the `PhaseState`.

In my opinion the same applies for `EventPublisher`. It may make sense to decouple the `EventPublisher` from the `CoordinatorState` in a separate PR. We might have two structs: a `state` struct and an `io` struct. Both are fields of a `coordinator` struct which then used in the state machine.

```rust
pub struct State {
    /// The credentials of the coordinator.
    pub keys: EncryptKeyPair,
    /// Internal ID used to identify a round
    pub round_id: u64,
    /// The round parameters.
    pub round_params: RoundParameters,
    /// The minimum of required sum/sum2 messages.
    pub min_sum_count: usize,
    /// The minimum of required update messages.
    pub min_update_count: usize,
    /// The minimum time (in seconds) reserved for processing sum/sum2 messages.
    pub min_sum_time: u64,
    /// The minimum time (in seconds) reserved for processing update messages.
    pub min_update_time: u64,
    /// The maximum time (in seconds) permitted for processing sum/sum2 messages.
    pub max_sum_time: u64,
    /// The maximum time (in seconds) permitted for processing update messages.
    pub max_update_time: u64,
    /// The number of expected participants.
    pub expected_participants: usize,
    /// The masking configuration.
    pub mask_config: MaskConfig,
    /// The size of the model.
    pub model_size: usize,
}

pub struct IO<R> {
    /// The request receiver half.
    pub request_rx: RequestReceiver<R>,
    /// The event publisher.
    pub events: EventPublisher,
    /// The metrics sender half.
    pub metrics_tx: MetricsSender, 
}

pub struct Coordinator<R> {
    pub state: State,
    pub io: IO<R>
}


pub struct PhaseState<R, S> {
    /// The inner state.
    pub(in crate::state_machine) inner: S,
    /// The Coordinator state.
    pub(in crate::state_machine) coordinator: Coordinator<R>,
}
```

**Future improvements**

My original plan was to have something like a log macro e.g. `metric!(round_id, 1)` which you can add anywhere in the code without having to carry a metrics sender around. However, I have no glue how to do that. I know that the `metrics-rs` crate does just that, but it is at a very early stage of development. It could be possible to do this with `#[proc_macro_hack]`.


Robert-Steiner:
I was able to implement a [Function-like procedural macros](https://doc.rust-lang.org/reference/procedural-macros.html#function-like-procedural-macros) for the code `sender.send(metric)`
The macro will only add this line if the `metrics` feature is enabled. You can find an example [here](https://github.com/xaynetwork/xaynet/pull/488/files#diff-0f9fdb67ec3d985e033d4ccf9649d660).

little-dude:
@Robert-Steiner should I review now or do you think it's worth waiting for you to rebase against the changes in #492 ?

Robert-Steiner:
@little-dude I think it makes sense to merge #492 first. Then I can rebase and finish the implementation of the missing message metrics.
I will drop a message when it is ready for review.
@finiteprods Since you've already started the review, feel free to post your comments.

rsaffi:
#build

rust/src/state_machine/phases/sum.rs
```diff
@@ -95,6 +103,18 @@ where
...
+        #[cfg(feature = "metrics")]
+        let Self {
+            inner: Sum {
+                sum_dict,
+                seed_dict,
+            },
+            coordinator_state,
+            request_rx,
+            metrics_tx,
+        } = self;
```
finiteprods, line 116: any reason here it's not
```rust
let Self {
            inner: Sum {
                sum_dict,
                seed_dict,
            },
            coordinator_state,
            request_rx,
            #[cfg(feature = "metrics")] metrics_tx,
        } = self;
```
?

Robert-Steiner:
> looks good to me so far! left a few comments that caught me eye. BTW when trying to build it earlier, I got several
> 
> ```
> error[E0658]: procedural macros cannot be expanded to statements
> ```
> 
> errors - but perhaps this is expected.

You right, function-like procedural macros were added recently. I forgot to mention that you need at least rust v1.45.0. (`rustup update` should do the job)

rust/examples/metrics-macro.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[tokio::main]
+async fn main() {
+    // adds `sender.send(metrics::round::total_number::update(1));`if the feature `metrics` is enabled
+    // `cargo run --example metrics-macro --features metrics` prints `Ok(ValidQuery("state_machine round_total_number=1i 1596543817214696000"))`
+    // however `cargo run --example metrics-macro --features metrics` prints nothing
+    #[cfg(feature = "metrics")]
+    let sender = Sender::new();
+
+    metrics!(
```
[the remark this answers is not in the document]
Robert-Steiner: That was actually my plan to eliminate the dependency of the sender. It is possible but I haven't yet figured out how it works. I took most of the ideas from the `metrics-rs` crate (in particular from the `refactor/metrics-v2` branch). As I understood it, you can solve it with a (global) singleton. I've seen the use of "unsafe" so I've avoided implementing it in this PR😄. If you are interested in their implementation you can take a look at the `metrics-macros/src/lib.rs` and `metrics/src/recorder.rs` file. But I agree, it would be a nice improvement because then we can kick the metrics sender out of the state machine and use the `metrics!` macro anywhere in the code like a `log` macro.

rust/src/bin/main.rs
```diff
@@ -40,17 +48,47 @@ async fn main() {
...
-            println!("shutting down: REST server terminated");
+            warn!("shutting down: REST server terminated");
         }
+        _ =  signal::ctrl_c() => {}
+    }
+
+    // The moment the message_handler is dropped, the sender half of the request channel is also
+    // dropped, which means that the state machine shuts itself down.
+    warn!("shutting down service");
+    let _ = state_machine_handle.await;
```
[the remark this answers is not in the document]
Robert-Steiner: The state machine can only be terminated if all request senders have been dropped. Since only the HTTP service has a sender, there is no way the state machine can terminate earlier than the HTTP service. Of course there is also the possibility of panic but in this case the whole program would just crash.

rust/src/state_machine/phases/sum.rs
Robert-Steiner: I tried but to my surprise it doesn't work in the destructuring pattern.

Robert-Steiner:
I would like to briefly summarize again what the idea behind the macro is and how it works.
The idea is to reduce the usage of `#[cfg(feature = "metrics")]`.
For example, for each `sender.send(metrics)` line we need to add `#[cfg(feature = "metrics")]` as well.

```rust
#[cfg(feature = "metrics")]
sender.send(metrics)
```

It would be more elegant if we could reduce the two lines into one line.

```rust
metrics!(sender, metric::update());
```

First, I tried to solve this with `marcos_rules!`

```rust
macro_rules! metrics {
    ($sender:expr, $metric:expr) => {
        #[cfg(feature = "metrics")]
        $sender.send($metric);
    };
}

metrics!(sender, metric::update());
```

However it has two downsides. 
1. we always create an object (`metric::update()`) even if we don't use the metric feature.
2. the metrics module + its dependencies must be available in both cases (with/without the metrics feature) otherwise we get a compiler error.

Next, I tried to solve it with `proc-macros`. 
`proc-macros` are more powerful than `marcos_rules!`. They cannot be called in the crate in which they are defined, which is why we have to create the additional crate `xaynet-metrics-macro`. I think the reason for this need is that `proc-macros` have to be compiled before the actual code is compiled. 

Now the trick is that we use conditional compilation in the `proc-macros` as well. 

```rust
#[proc_macro]
pub fn metrics(input: TokenStream) -> TokenStream {
    #[cfg(not(feature = "metrics"))]
    return TokenStream::new();

    #[cfg(feature = "metrics")]
    {
        use quote::quote;
        use syn::parse_macro_input;
        let metric::Send { sender, metrics } = parse_macro_input!(input as metric::Send);

        let gen = quote! {
                #(#sender.send(#metrics);)*
        };
        gen.into()
    }
}
```

So when we call `cargo build` (without the `metrics` feature) the line `metrics!(sender, metric::update());` is replaced by `()` before the crate `xaynet` is compiled. 

However, this approach also has its disadvantages. We need to publish an additional crate and it is a bit hacky.
Therefore I'm fine to postpone the macro and go with 

```rust 
#[cfg(feature = "metrics")]
sender.send(metrics)
```

first.

rsaffi:
#build

rust/Cargo.toml
```diff
@@ -42,6 +42,9 @@ async-trait = "0.1.35"
 influxdb = { version = "0.1.0", features = ["derive"], optional = true }
 chrono = { version = "0.4.13", optional = true }
 
+[dependencies.xaynet-metrics-macro]
+path = "xaynet-metrics-macro"
```
little-dude, line 46: Not 100% sure but I think we need `optional = true`. Otherwise, will it not be always pulled in?

little-dude: Although I think we actually need it even without the `metrics` feature flag. So it's probably the line 
```toml
metrics = ["influxdb", "chrono", "xaynet-metrics-macro/metrics"]
``` that should be changed back to 
```toml
metrics = ["influxdb", "chrono"]
```
Correct?

rust/src/state_machine/phases/update.rs
```diff
@@ -133,8 +157,26 @@ impl Handler for PhaseState<Update> {
...
+                metrics!(
+                    self.metrics_tx,
+                    metrics::message::update::increment(
+                        self.coordinator_state.round_id,
+                        Self::NAME
+                    )
+                );
+                self.handle_update(update_req)
+            }
+            _ => {
```
little-dude, line 170: This patterns comes back a few times. I wonder if there would be a way to use a `Rejection` trait or to extend the existing `Handler` trait to get rid of the repetition.

rust/xaynet-metrics-macro/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[dependencies]
+syn = "1.0.37"
+quote = "1.0"
+
+[lib]
+proc-macro = true
+
+[features]
+metrics = []
```
little-dude, line 15: I'm curious about how this works. Are feature flags passed to the dependencies? Specifically, if we compile `xaynet` with the `metrics` feature flag, then cargo will also build `xaynet-metrics-macro` with the `metrics` feature? That doesn't seem right to me.

rust/xaynet-metrics-macro/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+        Ok(Send { sender, metrics })
+    }
+}
+
+#[proc_macro]
+pub fn metrics(input: TokenStream) -> TokenStream {
+    let Send { sender, metrics } = parse_macro_input!(input as Send);
+
+    TokenStream::from(quote! {
+            #[cfg(feature = "metrics")]
```
little-dude, line 48: :+1:

rust/src/bin/main.rs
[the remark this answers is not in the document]
little-dude: > The state machine can only be terminated if all request senders have been dropped

As of today, this is true, but we may introduce other termination conditions in the future. And in that case, we might run into the case @finiteprods mentioned. But I agree about the clean shutdown issue. Maybe we could have a custom drop implementation that performs the shutdown?

rsaffi:
#build

rust/src/bin/main.rs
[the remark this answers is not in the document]
Robert-Steiner: I agree. I moved the state machine back into the `select!`. I really like the idea to replace the shutdown state with a custom drop implementation👍

rust/src/state_machine/phases/update.rs
Robert-Steiner: I added a `reject_request` method to the `Handler` trait

rsaffi:
#build

rust/src/state_machine/phases/update.rs
Robert-Steiner: It doesn't work the way I think it will 😔

I first tried adding `reject_request` to the `Handler` trait

```rust
pub trait Handler {
    fn reject_request(&mut self) -> Result<(), PetError> {
        metrics!(
            self.shared.io.metrics_tx,
            metrics::message::rejected::increment(self.shared.state.round_id, Self::NAME)
        );
        Err(PetError::InvalidMessage)
    }
}
```

but here I get an error because I cannot access `self.shared` or `Self:NAME` (which makes sense).

then I tried

```rust
impl<S> Handler for PhaseState<S>
where
    Self: Phase,
{
    /// Rejects a request by returning a [`PetError::InvalidMessage`] back to the caller.
    fn reject_request(&mut self) -> Result<(), PetError> {
        metrics!(
            self.shared.io.metrics_tx,
            metrics::message::rejected::increment(self.shared.state.round_id, Self::NAME)
        );
        Err(PetError::InvalidMessage)
    }
}
```
and get:
```
error[E0119]: conflicting implementations of trait `state_machine::phases::Handler` for type `state_machine::phases::PhaseState<state_machine::phases::idle::Idle>`:
   --> src/state_machine/phases/mod.rs:168:1
    |
168 | / impl<S> Handler for PhaseState<S>
169 | | where
170 | |     Self: Phase,
171 | | {
...   |
179 | |     }
180 | | }
    | |_^ conflicting implementation for `state_machine::phases::PhaseState<state_machine::phases::idle::Idle>`
    | 
   ::: src/state_machine/phases/idle.rs:23:1
    |
23  |   impl Handler for PhaseState<Idle> {
    |   --------------------------------- first implementation here
```
then I tried to add a `Rejection` trait

```rust
pub trait Rejection {
    /// Rejects a request by returning a [`PetError`] back to the caller.
    fn reject_request(&mut self) -> Result<(), PetError>;
}

impl<S> Rejection for PhaseState<S>
where
    Self: Phase,
{
    /// Rejects a request by returning a [`PetError::InvalidMessage`] back to the caller.
    fn reject_request(&mut self) -> Result<(), PetError> {
        metrics!(
            self.shared.io.metrics_tx,
            metrics::message::rejected::increment(self.shared.state.round_id, Self::NAME)
        );
        Err(PetError::InvalidMessage)
    }
}


impl Handler for PhaseState<Sum>
where
    Self: Rejection,
{
    fn handle_request(&mut self, req: StateMachineRequest) -> Result<(), PetError> {
        match req {
            StateMachineRequest::Sum(sum_req) => {
...
                self.handle_sum(sum_req)
            }
            _ => self.reject_request(),
        }
    }
}
```

but this also didn't work

```
error[E0275]: overflow evaluating the requirement `state_machine::phases::PhaseState<state_machine::phases::sum::Sum>: state_machine::phases::Rejection`
  --> src/state_machine/phases/sum.rs:34:1
   |
34 | / impl Handler for PhaseState<Sum>
35 | | where
36 | |     Self: Rejection,
37 | | {
...  |
54 | |     }
55 | | }
   | |_^
   |
   = help: see issue #48214
```

So I dicided to add a `reject_request` method to `impl<S> PhaseState<S> where Self: Phase,`

rust/src/state_machine/phases/mod.rs
```diff
@@ -57,10 +60,19 @@ pub trait Phase {
...
 }
 
-/// A trait that must be implemented by a state to handle a request.
+/// A trait that must be implemented by a state to handle and reject a request.
 pub trait Handler {
     /// Handles a request.
     fn handle_request(&mut self, req: StateMachineRequest) -> Result<(), PetError>;
+
+    /// Rejects a request by returning a [`PetError::InvalidMessage`] back to the caller.
+    fn reject_request(&mut self) -> Result<(), PetError> {
```
little-dude: It feels a bit weird to put that here imo. Not sure but could we instead move that to the `process_single` method?

```rust
impl<S> PhaseState<S>
where
    Self: Handler + Phase,
{
    async fn process_single(&mut self) -> Result<(), StateError> {
        let (req, resp_tx) = self.next_request().await?;
        let span = req.span();
        let _span_guard = span.enter();
        let res = self.handle_request(req.into_inner());

        // if res is an InvalidMessage, do the `metrics!` stuff here
        // ...
        // ...

        // This may error out if the receiver has already be dropped but
        // it doesn't matter for us.
        let _ = resp_tx.send(res.map_err(Into::into));
        Ok(())
    }
}
```

rust/src/state_machine/phases/update.rs
little-dude: I just commented about that. Maybe we don't even need a trait actually and this could just go to the `PhaseState::process_single` impl. (Not sure though since I haven't tried it)

rsaffi:
#build

rust/src/state_machine/phases/mod.rs
Robert-Steiner: I like that idea but it would be an empty `if` if we compile it without the `metrics` feature. But the compiler will probably optimize that away. 
I will add the macro to the `process_single` method.

rsaffi:
#build

Robert-Steiner:
Updates:

- renamed `xain-metrics-macro` to `xaynet-macros` (just in case we decide in the future to remove the `metrics!` macro, we can reuse that crate for other macros) 
- added metrics/macro documentation
- removed macro example (I embedded the example into the macro documentation)
- added attributes `#![cfg_attr(docsrs, feature(doc_cfg))]` and `#[cfg_attr(docsrs, doc(cfg(feature = "metrics")))]` to `rust/src/lib.rs`. They trigger that feature note
Screen Shot 2020-08-12 at 15 00 34

is displayed. It only works if it is built by docs.rs. It can be configured to work locally but it doesn't work nicely with our setup since it is a nightly feature.

rsaffi:
> Updates:
> 
> * renamed `xain-metrics-macro` to `xaynet-macros` (just in case we decide in the future to remove the `metrics!` macro, we can reuse that crate for other macros)
> * added metrics/macro documentation
> * removed macro example (I embedded the example into the macro documentation)
> * added attributes `#![cfg_attr(docsrs, feature(doc_cfg))]` and `#[cfg_attr(docsrs, doc(cfg(feature = "metrics")))]` to `rust/src/lib.rs`. They trigger that feature note is displayed. It only works if it is built by docs.rs. It can be configured to work locally but it doesn't work nicely with our setup since it is a nightly feature.

Very, very nice! :clap:

rust/src/state_machine/phases/mod.rs
little-dude: > But the compiler will probably optimize that away.

For sure yeah.

rsaffi:
#build

Last one prior to it being merged (but please, if you already feel like merging it, go ahead!)

rust/xaynet-macros/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+///     {
+///         sender.send(metrics::round::total_number::update(1)),
+///         sender.send(metrics::masks::total_number::update(1, 1, PhaseName::Idle)),
+///     };
+/// }
+/// ```
+///
+/// ## Sender
+///
+/// A `Sender` must implement the method `pub fn send(&self, metics: T)` where `T`
```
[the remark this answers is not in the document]
Robert-Steiner: good catch!
thanks for the review 👍

PR #489, opened 2020-08-04
metrics feature flag

finiteprods:
This merge request updates the manifest to support an "opt-in" for a forthcoming feature `metrics`. Making this feature optional means we can avoid the overhead (both at run time and compile time) associated with this feature when we don't need it on.

There is also a small test app `bin/hello-metrics.rs` to illustrate how to use it in our library code, which will be removed in the final PR.
By default, without the feature:
```
$ cargo run --bin hello-metrics
Hello World!
```
With the feature:
```
$ cargo run --features "metrics" --bin hello-metrics
Hello World!
Influx says Hi!
```


rust/Cargo.toml
```diff
@@ -38,6 +38,9 @@ uuid = { version = "0.8.1", features = ["v4"]}
 rayon = "1.3.0"
 async-trait = "0.1.35"
 
+# optional dependencies
+influxdb = { version = "0.1.0", features = ["derive"], optional = true }
+
```
Robert-Steiner, line 43: Can you please add `chrono = "0.4.13"` here as well. We need this library to create the timestamps of the metric data points.

finiteprods: sure!

PR #491, opened 2020-08-06
CI change: allow for the Cronjob of Rust Audit to fail (but still be notified on Slack)

rsaffi:


.github/workflows/rust-audit-cron.yml
```diff
@@ -22,13 +22,12 @@ jobs:
...
       - name: Run rust-audit
         id: rust-audit
-        continue-on-error: true
         run: |
           cargo audit --deny-warnings -f rust/Cargo.lock
 
       - name: Notify on Slack
         uses: 8398a7/action-slack@v3
-        if: steps.rust-audit.outcome == 'failure'
+        if: ${{ failure() }}
```
rsaffi, line 30: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

PR #492, opened 2020-08-06
refactor tracing

little-dude:
use of the `Traced` struct has two downsides:

- Because we want to decouple the tracing logic from the application
  logic, we introduced generics all over the place for our code to
  work both with T and Traced<T>.
- We currently pass the `Traced.span()` down all the services chain, but
  `Span`s are more of less immutable and declared in macros at compile
  time, so just passing this span down is useless: we cannot at any
  field to it.

This refactoring is  twofold:

1. Remove all the genericity around tracing. This resulted in a lot of
   simplifications, especially in the state machine, but also in the
   pet message services since we could get rid of the
   `_PetMessageHandler` trait.

2. Replace Traced<T> by a Request<T> that is going to be used
   everywhere. Note that we currently just use `Request` in for the
   PET message services and the state machine, but it should be
   straightforward to use anywhere else. Unlike `Traced<T>`,
   `Request<T>` allows us to enrich the span attached to the request:
   `Request::map` creates a child span a the request is mapped.


little-dude:
I still need to update the tests but it's ready for reviewing

rust/src/services/messages/message_parser.rs
```diff
@@ -58,16 +59,17 @@ impl MessageParserService {
...
-/// It contains the parsed message.
-pub type MessageParserResponse = Result<MessageOwned, MessageParserError>;
+pub struct RawMessage<T: AsRef<[u8]>>(T);
+
+impl<T> Traceable for RawMessage<T>
+where
+    T: AsRef<[u8]>,
+{
+    fn make_span(&self) -> Span {
+        error_span!("raw_message", payload_len = self.0.as_ref().len())
```
Robert-Steiner, line 70: What is the reason for using the `error_span!` instead of using `info_span!`?

rust/src/utils/request.rs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        let Request {
+            mut metadata,
+            inner,
+        } = self;
+        let mapped = f(inner);
+
+        // self.span() is the parent of the span associated to the
+        // inner type
+        let new_span = metadata.span().in_scope(|| mapped.make_span());
```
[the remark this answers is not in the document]
little-dude: The span is created within the scope of the `metadata.span()`, so it's a child of `metadata.span()`. Then we attach this child span to the metadata. It may seem like the "original span" is gone, but actually, it is kept alive because it has a child.

rust/src/services/messages/message_parser.rs
little-dude: If the log level is `trace`, `warn` or `info`, there's no difference. But if the log level is `warn` or `error`, the span created by `info_span!` would be empty. By using `error_span!` we're basically saying: "whatever the log level is, make this a non-empty span".

Robert-Steiner: Ahh I see
Thank you!

rust/src/utils/request.rs
```diff
@@ -0,0 +1,129 @@
...
+{
+    /// Create a new request
+    pub fn new(t: T) -> Self {
+        Self {
+            inner: t,
+            metadata: RequestMetadata::new(),
+        }
+    }
+
+    pub fn from_parts(metadata: RequestMetadata, inner: T) -> Self {
```
finiteprods, line 72: ditto

rust/src/utils/request.rs
```diff
@@ -0,0 +1,129 @@
...
+        Self { metadata, inner }
+    }
+
+    pub fn metadata(&self) -> RequestMetadata {
```
finiteprods, line 76: ditto

PR #493, opened 2020-08-06
metrics settings

finiteprods:
Adds a settings struct for `[metrics.influxdb]` settings

```rust
pub struct InfluxSettings {
    /// The URL where InfluxDB is running.
    ///
    /// # Examples
    ///
    /// **TOML**
    /// ```text
    /// [metrics.influxdb]
    /// url = "http://localhost:8086"
    /// ```
    ///
    /// **Environment variable**
    /// ```text
    /// XAYNET_METRICS__INFLUXDB__URL=http://localhost:8086
    /// ```
    pub url: String,

    /// The InfluxDB database name.
    ///
    /// # Examples
    ///
    /// **TOML**
    /// ```text
    /// [metrics.influxdb]
    /// db = "test"
    /// ```
    ///
    /// **Environment variable**
    /// ```text
    /// XAYNET_METRICS__INFLUXDB__DB=test
    /// ```
    pub db: String,
}

```

### DIFF
--- a/configs/docker-release.toml
+++ b/configs/docker-release.toml
@@ -23,3 +23,7 @@ model_type = "M3"
 
 [model]
 size = 4
+
+[metrics]
+store_url = "http://influxdb:8086"
+store_name = "metrics"
--- a/configs/docker-release.toml
+++ b/configs/docker-release.toml
@@ -24,6 +24,6 @@ model_type = "M3"
 [model]
 size = 4
 
-[metrics]
-store_url = "http://influxdb:8086"
-store_name = "metrics"
+[metrics.influxdb]
+url = "http://influxdb:8086"
+db = "metrics"
--- a/rust/src/settings.rs
+++ b/rust/src/settings.rs
@@ -34,6 +34,7 @@ pub struct Settings {
     pub mask: MaskSettings,
     pub log: LoggingSettings,
     pub model: ModelSettings,
+    #[validate]
     pub metrics: MetricsSettings,
 }
 
@@ -437,16 +438,18 @@ pub struct ModelSettings {
     pub size: usize,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Validate)]
 /// Metrics settings.
 pub struct MetricsSettings {
+    #[validate]
     /// Settings for the InfluxDB backend.
     pub influxdb: InfluxSettings,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Validate)]
 /// InfluxDB settings.
 pub struct InfluxSettings {
+    #[validate(url)]
     /// The URL where InfluxDB is running.
     ///
     /// # Examples
--- a/rust/src/settings.rs
+++ b/rust/src/settings.rs
@@ -452,7 +452,7 @@ pub struct MetricsSettings {
     ///
     /// **Environment variable**
     /// ```text
-    /// XAYNET_API__STORE_URL=http://localhost:8086
+    /// XAYNET_METRICS__STORE_URL=http://localhost:8086
     /// ```
     pub store_url: String,
 
@@ -468,7 +468,7 @@ pub struct MetricsSettings {
     ///
     /// **Environment variable**
     /// ```text
-    /// XAYNET_API__STORE_NAME=test
+    /// XAYNET_METRICS__STORE_NAME=test
     /// ```
     pub store_name: String,
 }
--- a/rust/src/settings.rs
+++ b/rust/src/settings.rs
@@ -440,37 +440,44 @@ pub struct ModelSettings {
 #[derive(Debug, Deserialize)]
 /// Metrics settings.
 pub struct MetricsSettings {
-    /// The URL address of the metrics store.
-    ///
-    /// # Examples
-    ///
-    /// **TOML**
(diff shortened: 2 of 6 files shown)


rust/src/settings.rs
```diff
@@ -436,6 +437,42 @@ pub struct ModelSettings {
...
+    ///
+    /// **TOML**
+    /// ```text
+    /// [metrics]
+    /// store_url = "http://localhost:8086"
+    /// ```
+    ///
+    /// **Environment variable**
+    /// ```text
+    /// XAYNET_API__STORE_URL=http://localhost:8086
```
little-dude, line 455: Shouldn't these be `XAYNET__METRICS_FOO` instead?

configs/docker-release.toml
```diff
@@ -23,3 +23,7 @@ model_type = "M3"
 
 [model]
 size = 4
+
+[metrics]
+store_url = "http://influxdb:8086"
```
little-dude, line 28: I suspect that some settings will be specific to the metrics backend we're using. Maybe we have different settings for each backend:

```toml
[metrics.influxdb]
url = "http://influxdb:8086"
store_name = "metrics"
```

rust/src/settings.rs
finiteprods: yes indeed, well caught!

configs/docker-release.toml
finiteprods: fair point, I think namespacing `[metrics.influxdb]` is indeed useful

rust/src/settings.rs
```diff
@@ -436,6 +437,42 @@ pub struct ModelSettings {
...
+    /// ```text
+    /// [metrics]
+    /// store_name = "test"
+    /// ```
+    ///
+    /// **Environment variable**
+    /// ```text
+    /// XAYNET_API__STORE_NAME=test
+    /// ```
+    pub store_name: String,
```
Robert-Steiner: I think it will be clearer to a user if we use the influxdb specific term `database_name` or `db_name` instead of `store_name`.

rust/src/settings.rs
```diff
@@ -436,6 +437,42 @@ pub struct ModelSettings {
...
+    /// ```text
+    /// [metrics]
+    /// store_url = "http://localhost:8086"
+    /// ```
+    ///
+    /// **Environment variable**
+    /// ```text
+    /// XAYNET_API__STORE_URL=http://localhost:8086
+    /// ```
+    pub store_url: String,
```
Robert-Steiner: Do you know if it's easy to check if the value is a valid URL?

For example you can use the type `std::net::SocketAddr` for ip addresses. 
When the configuration file is parsed, `SocketAddr::FromStr` is called and it will automatically fail if it is not a valid IP address. We used that behaviour for `ApiSettings::bind_address`. Maybe there is something similar for urls in the std lib.

However, if it requires an additional library, I'm not sure it's worth it.

little-dude: We can easily check whether a URL is valid, whether is contains a hostname, what the scheme is, etc. But we cannot check that the domain is valid in that it resolves to an IP address without making a network call, and I think we should refrain from doing network calls when parsing a config file, because they can hang.

rust/src/settings.rs
finiteprods: yes, at the time I was erring on the side of being agnostic to the backend, but really these settings are influx-specific (as per @little-dude 's comment above). the naming should be better now together with the new nested `[metrics.influxdb]` setting.

finiteprods:
> Looks good👍
> Do you know if we can make these settings optional? I mean for the whole settings object we could use `#[cfg(feature = "metrics")]` but I think it is nicer if we can just make it optional.
> 
> If you like you can also add the fields `username` and `password`. Both have the type `String`. But here it is important that both fields are optional because influx allows communication without authentication.

thanks for reviewing. would you be happy if we postpone the "optionality" of the settings to another PR? because I had in mind other settings (in PET, especially) which should also be optional, so it could all be tackled there.

rust/src/settings.rs
finiteprods: one thing we can certainly do I think is `#[validate(url)]`. I'll add that now...

Robert-Steiner: Right, I just thought of checking that scheme is valid. 
But maybe it's not that important for now (just an idea I had in mind).

Robert-Steiner:
> > Looks good👍
> > Do you know if we can make these settings optional? I mean for the whole settings object we could use `#[cfg(feature = "metrics")]` but I think it is nicer if we can just make it optional.
> > If you like you can also add the fields `username` and `password`. Both have the type `String`. But here it is important that both fields are optional because influx allows communication without authentication.
> 
> thanks for reviewing. would you be happy if we postpone the "optionality" of the settings to another PR? because I had in mind other settings (in PET, especially) which should also be optional, so it could all be tackled there.

Sounds good to me 👍

little-dude:
> Do you know if we can make these settings optional? I mean for the whole settings object we could use `#[cfg(feature = "metrics")]` but I think it is nicer if we can just make it optional.

I think we shouldn't hide it completely, because that means the config wouldn't parse based on whether the feature is there or not. A more user-friendly approach imo would be to log a warning when there's a metrics section in the config, but the metrics feature is not enabled.

PR #495, opened 2020-08-10
Move coordinator state and I/O into the Shared struct

Robert-Steiner:
The implantation of the suggestion in #488

If everyone likes it I would merge this pr first. It will save me a lot of `#[cfg(feature = "metrics")]` annotations in #488 :)

### DIFF
--- a/rust/src/state_machine/phases/mod.rs
+++ b/rust/src/state_machine/phases/mod.rs
@@ -21,6 +21,7 @@ pub use self::{
 use crate::{
     state_machine::{
         coordinator::CoordinatorState,
+        events::EventPublisher,
         requests::{RequestReceiver, ResponseSender, StateMachineRequest},
         StateMachine,
     },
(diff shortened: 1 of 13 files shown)


rust/src/state_machine/phases/mod.rs
```diff
@@ -62,17 +63,61 @@ pub trait Handler {
...
+pub struct IO {
+    /// The request receiver half.
+    pub(in crate::state_machine) request_rx: RequestReceiver,
+    /// The event publisher.
+    pub(in crate::state_machine) events: EventPublisher,
+}
+
+/// A struct that contains the coordinator state and the I/O interfaces.
+#[derive(Debug)]
+pub struct Shared {
```
finiteprods, line 77: it's not very clear to me what "shared" means here. perhaps a more descriptive name (and docstring)?

Robert-Steiner:
```suggestion
/// A struct that contains the coordinator state and the I/O interfaces that is shared and accessible by all PhaseStates.
#[derive(Debug)]
pub struct Shared {
```
maybe like this?

finiteprods: Yes, I think it’s better

PR #502, opened 2020-08-24
Replace `Proxy` by the `ApiClient` trait

little-dude:
### replace `Proxy` by the `ApiClient` trait

The trait has two implementors:

- `InMemoryApiClient` that queries the coordinator services via
  in-memory channels
- `HttpApiClient` that queries the coordinator services via HTTP

The second commit renames `crate::client::request` into `create::client::api`.

### DIFF
--- a/rust/src/client/api/mod.rs
+++ b/rust/src/client/api/mod.rs
@@ -1,9 +1,4 @@
-//! Provides functionality to enable clients to communicate with a XayNet
-//! service over HTTP.
-//!
-//! See the [client module] documentation since this is a private module anyways.
-//!
-//! [client module]: ../index.html
+//! This module provides clients for the Xaynet coordinator API.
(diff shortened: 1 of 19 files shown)


rust/src/client/api/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    state_machine::coordinator::RoundParameters,
+    SumDict,
+    SumParticipantPublicKey,
+    UpdateSeedDict,
+};
+
+/// An interface that API clients implement
+#[async_trait]
+pub trait ApiClient {
+    type Error: ::std::fmt::Debug + ::std::error::Error + 'static;
```
Robert-Steiner, line 22: Nice!

Robert-Steiner (approved):
Really good!

PR #501, opened 2020-08-21
prepare-cargo-workspaces Move RoundSeed and RoundParameters out of coordinator.rs

Robert-Steiner:

PR #503, opened 2020-08-24
Split the crate into smaller ones

little-dude:
Split the `xaynet` crate into:

- `xaynet_core`
- `xaynet_client`
- `xaynet_ffi`
- `xaynet_server`

To do so, we needed to isolate a few types that are common to several of the aforementioned crates:

- moving the `RoundParameters` and `RoundSeed` types out of `xaynet::state_machine` (https://github.com/xaynetwork/xaynet/pull/501)
- replace the `Proxy` enum (which depends on `xaynet::services`) by an `ApiClient` trait so that the client code doesn't depend on types defined on the server side (https://github.com/xaynetwork/xaynet/pull/502)


rust/xaynet-client/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+derive_more = { version = "0.99.7", default-features = false, features = ["from"] }
+serde = { version = "1.0.111", features = ["derive"] }
+bytes = "0.5.4"
+sodiumoxide = "0.2.5"
+bincode = "1.2.1"
+thiserror = "1.0.19"
+tracing = "0.1.16"
+async-trait = "0.1.35"
+xaynet-core = { path = "../xaynet-core" }
+xaynet-server = { path = "../xaynet-server" }
```
little-dude, line 23: This can actually be a dev-dependency because it is only needed for the in-memory client, which is only compiled for the tests:

```
❯ rg server
Cargo.toml
23:xaynet-server = { path = "../xaynet-server" }

src/api/in_memory.rs
10:use xaynet_server::services::{FetchError, Fetcher, PetMessageError, PetMessageHandler};
```

Atm we don't even use it in the tests though, so I'm not even sure it's worth keeping but that may be a debate for another PR.

Fwiw I found this using [`cargo-udeps`](https://github.com/est31/cargo-udeps), which is a pretty useful tool.

Robert-Steiner: good catch 👍
I fixed it 
Yeah I agree. I think the idea was to keep it for integration tests later.

rust/xaynet-core/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "xaynet-core"
+version = "0.9.0"
+authors = ["Xayn Engineering <engineering@xaynet.dev>"]
+edition = "2018"
+description = "The Xayn Network project is building a privacy layer for machine learning so that AI projects can meet compliance such as GDPR and CCPA. The approach relies on Federated Learning as enabling technology that allows production AI applications to be fully privacy compliant."
```
little-dude, line 6: Maybe add something like:

> `xaynet_core` provides building blocks that are common to the Xayn Networks backend and client.

rust/xaynet-ffi/src/api.rs
```diff
@@ -50,10 +50,8 @@ use tokio::{
     time::Duration,
```
little-dude, line 50: I think we could move all this to `lib.rs` no? It seems a bit artificial to have only that one module.

rust/xaynet-ffi/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "xaynet-ffi"
+version = "0.9.0"
+authors = ["Xayn Engineering <engineering@xaynet.dev>"]
+edition = "2018"
+description = "The Xayn Network project is building a privacy layer for machine learning so that AI projects can meet compliance such as GDPR and CCPA. The approach relies on Federated Learning as enabling technology that allows production AI applications to be fully privacy compliant."
```
little-dude, line 6: Maybe add:

> `xaynet_ffi` provides an implementation of a Xayn Network client with C bindings.

rust/xaynet-client/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "xaynet-client"
+version = "0.9.0"
+authors = ["Xayn Engineering <engineering@xaynet.dev>"]
+edition = "2018"
+description = "The Xayn Network project is building a privacy layer for machine learning so that AI projects can meet compliance such as GDPR and CCPA. The approach relies on Federated Learning as enabling technology that allows production AI applications to be fully privacy compliant."
```
little-dude, line 6: > `xaynet_client` provides an implementation of a Xayn Network client

rust/xaynet-core/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "xaynet-core"
+version = "0.9.0"
+authors = ["Xayn Engineering <engineering@xaynet.dev>"]
+edition = "2018"
+description = "`xaynet_core` provides building blocks that are common to the Xayn Networks backend and client."
```
finiteprods, line 6:
```suggestion
description = "`xaynet_core` provides building blocks that are common to the Xayn Network backend and client."
```

finiteprods (commented):
this is great! i would be happy to approve it once the outstanding test is dealt with :+1:

Robert-Steiner (approved):

PR #506, opened 2020-08-26
Add mobile client error handling

Robert-Steiner:
Summary:
- remove `unwrap` in places where it is not safe to`unwrap`

### DIFF
--- a/rust/examples/mobile-client.rs
+++ b/rust/examples/mobile-client.rs
@@ -103,7 +103,7 @@ fn main() -> Result<(), ()> {
         .init();
 
     // create a new client
-    let client = MobileClient::init(&opt.url, get_participant_settings());
+    let client = MobileClient::init(&opt.url, get_participant_settings()).unwrap();
     // serialize the current client state (and save it on the phone)
     let mut bytes = client.serialize();
 
@@ -119,10 +119,13 @@ fn main() -> Result<(), ()> {
 // perform the participant task (this function should be triggered regularly on the phone while the
 // app is active or in a background task)
 fn perform_task(url: &str, bytes: &[u8], model: Model) -> Vec<u8> {
-    let mut client = MobileClient::deserialize(url, bytes);
+    let mut client = MobileClient::restore(url, bytes).unwrap();
     client.set_local_model(model);
-    client = client.perform_task();
-    println!("global model: {:?}", client.get_global_model());
+    client = match client.try_to_proceed() {
+        Ok(client) => client,
+        Err((client, _)) => client,
+    };
+    println!("global model: {:?}", client.get_global_model().unwrap());
     let new_bytes = client.serialize();
     println!("size serialized: {:?}", &bytes.len());
     new_bytes
--- a/rust/xaynet-client/src/mobile_client/mod.rs
+++ b/rust/xaynet-client/src/mobile_client/mod.rs
@@ -39,6 +39,14 @@ pub struct MobileClient {
 }
 
 impl MobileClient {
+    /// Initializes a fresh client. This method only needs to called once.
+    ///
+    /// To serialize and restore a client, please use the [`MobileClient::serialize`] and
+    /// [`MobileClient::restore`]
+    ///
+    /// # Errors
+    ///
+    /// Fails if the crypto module cannot be initialized.
     pub fn init(
         url: &str,
         participant_settings: ParticipantSettings,
@@ -53,6 +61,12 @@ impl MobileClient {
         Ok(Self::new(url, client_state))
     }
 
+    /// Restores a client from its serialized state.
+    ///
+    /// # Errors
+    ///
+    /// Fails if the serialized state is corrupted and the client cannot be restored
+    /// or if the crypto module cannot be initialized.
     pub fn restore(url: &str, bytes: &[u8]) -> Result<Self, MobileClientError> {
         let client_state: ClientStateMachine = bincode::deserialize(bytes)?;
         Ok(Self::new(url, client_state))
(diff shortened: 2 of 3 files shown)


rust/examples/mobile-client.rs
```diff
@@ -103,7 +103,7 @@ fn main() -> Result<(), ()> {
         .init();
 
     // create a new client
-    let client = MobileClient::init(&opt.url, get_participant_settings());
+    let client = MobileClient::init(&opt.url, get_participant_settings()).unwrap();
```
little-dude, line 106: It would be nice to add a comment explaining why unwrapping is OK here and in the other instances of this.

rust/examples/mobile-client.rs
```diff
@@ -119,10 +119,10 @@ fn main() -> Result<(), ()> {
 // perform the participant task (this function should be triggered regularly on the phone while the
 // app is active or in a background task)
 fn perform_task(url: &str, bytes: &[u8], model: Model) -> Vec<u8> {
-    let mut client = MobileClient::deserialize(url, bytes);
+    let mut client = MobileClient::deserialize(url, bytes).unwrap();
     client.set_local_model(model);
-    client = client.perform_task();
-    println!("global model: {:?}", client.get_global_model());
+    client = client.perform_task().unwrap();
+    println!("global model: {:?}", client.get_global_model().unwrap());
```
little-dude, line 125: Unrelated to this PR but afaik most models will be at list around 1MB large, we may not want to print them.

rust/examples/mobile-client.rs
little-dude: Actually, we should not unwrap here. The init can fail as far as I can see. Shouldn't we exit gracefully with an error message?

rust/xaynet-client/src/mobile_client/mod.rs
```diff
@@ -8,26 +8,51 @@ use crate::{
...
     mask::Model,
+    InitError,
 };
 
+#[derive(Debug, Error)]
+/// Mobile client errors
+pub enum MobileClientError {
+    #[error("failed to deserialise mobile client: {0}")]
+    /// Failed to deserialise mobile client.
+    DeserialiseErr(#[from] bincode::Error),
```
little-dude, line 23: I'd remove the `Err` suffix on the variants. The `Error` is already part of the enum name.

little-dude: Also I think we should follow `serde` convention of using the US the spelling of "(de)seriali*z*ation", since it's the de facto standard in the Rust ecosystem.

rust/examples/mobile-client.rs
Robert-Steiner: This file is an example that I only use to test the mobile client implementation. My plan is to remove it once we moved the sdk into the repository.

rust/xaynet-client/src/mobile_client/mod.rs
Robert-Steiner: Yeah you right. I have changed it

Robert-Steiner:
I did one more small change to `perform_task`. If the creation of the runtime fails, the current state of the client is returned along with the error. Before, I just returned the error which would mean we would have to create a new client

rust/examples/mobile-client.rs
little-dude: Oh sorry I missed that.

rust/xaynet-client/src/mobile_client/mod.rs
```diff
@@ -85,15 +85,22 @@ impl MobileClient {
         Ok(global_model)
     }
 
-    pub fn perform_task(self) -> Result<Self, MobileClientError> {
+    pub fn perform_task(self) -> Result<Self, (Self, MobileClientError)> {
+        let mut runtime = match Self::runtime() {
+            Ok(runtime) => runtime,
+            // We don't want to loose the current client because of a runtime error.
+            // Therefore we return the error as well as the current client.
+            Err(err) => return Err((self, err.into())),
```
little-dude, line 93: Mmh I understand why we're doing that but it feels awkward. I think it would at least deserve an explanation in the docstring.

Robert-Steiner: I added the documentaion for all public methods :
In addtion, I renamed two methods `deserialize` -> `restore` and `perfrom_task` -> `try_to_proceed`.

rust/xaynet-client/src/mobile_client/mod.rs
```diff
@@ -27,6 +27,9 @@ pub enum MobileClientError {
     #[error("failed to initialize runtime: {0}")]
     /// Failed to initialize runtime.
     Runtime(#[from] std::io::Error),
+    #[error("failed to request API: {0}")]
```
little-dude, line 30: API request failed
```suggestion
    #[error("API request failed: {0}")]
```

rust/xaynet-client/src/mobile_client/mod.rs
```diff
@@ -27,6 +27,9 @@ pub enum MobileClientError {
...
+    /// Failed to request API.
```
little-dude, line 31:
```suggestion
    /// API request failed
```

rust/xaynet-client/src/mobile_client/mod.rs
```diff
@@ -79,13 +103,27 @@ impl MobileClient {
...
 
-    pub fn perform_task(self) -> Result<Self, (Self, MobileClientError)> {
+    /// Tries to proceed with the current client task.
+    /// This will consume the current state of the client and produces a new one.
+    ///
+    /// # Errors
+    ///
+    /// Fails if the runtime cannot be initialized.
+    /// In this case the state of the client remains unchanged and is returned
+    /// together with the error.
```
little-dude, line 125:
```suggestion
    /// along with the error.
```

little-dude (approved):

PR #508, opened 2020-08-27
Simplify messages

little-dude:
This PR simplifies the message types in preparation of the upcoming changes needed to support multipart messages:

- [X] remove non-`Owned` message types since they are not used
- [x] remove the certificate from the message header since it's currently not used, and it's unclear whether it will actually be needed
- [X] remove the coordinator public key from the message header since it's unused


Robert-Steiner:
For me it looks good👍
It makes reading the code easier. However I'm not sure if we better wait for @janpetschexain opinion because I was not so involved in that part of the code.

little-dude:
I just started with the multipart support so no need to merge right now. If too many changes queue up, I'll merge, but otherwise I'll wait for Monday.

Robert-Steiner (approved):

little-dude:
Merging because I'm about to open a new PR based on this one.

PR #513, opened 2020-08-31
Refactor messages and add support for multipart messages

little-dude:
The three commits can/should be reviewed independently.


little-dude:
#build

rust/xaynet-core/src/message/payload/chunk.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// A message chunk.
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct Chunk {
+    /// Chunk ID
+    id: u16,
+    /// ID of the message this chunk belongs to
+    message_id: u16,
+    /// `true` if this is the last chunk of the message, `false` otherwise
+    last: bool,
```
Robert-Steiner, line 35: Is there a way to limit the amount of chunks a participant can send?

rust/xaynet-client/src/mobile_client/participant/mod.rs
```diff
@@ -64,14 +64,15 @@ pub struct Participant<Task> {
...
+    /// The message is signed with the participant secret signing
+    /// key. `pk` is the coordinator public key, used to encrypt the
+    /// final message
     pub fn seal_message(&self, pk: &CoordinatorPublicKey, message: &Message) -> Vec<u8> {
-        let message_seal = MessageSeal {
-            recipient_pk: pk,
-            sender_sk: &self.state.keys.secret,
-        };
-        message_seal.seal(message)
+        let mut buf = vec![0; message.buffer_length()];
```
Robert-Steiner, line 73: Does it makes sense to add the `seal_message` method to `Message`? For me it feels a bit low level to work with bytes of the message.

Robert-Steiner (commented):
I doesn't feel so qualified to review this part of the codebase but for me it looks good.
I really like the simplification in the tower service!

rust/xaynet-core/src/message/payload/chunk.rs
little-dude: No. With this the server doesn't know the full message size. However we still have 24 "reserved" bytes, so we could maybe have a 16 bits long field for the total number of chunks. We'd still have 8 bits left.

rust/xaynet-client/src/mobile_client/participant/mod.rs
little-dude: I considered it but I had to make a choice between two APIs, but felt that the trade-offs should be made by the caller instead:

```
impl Message
  // This allocates two vec: one for serializing the message, one for the encrypted data. It's user friendly, and can't panic
  fn seal_1(&self, sk: SecretSigningKey, pk: PublicEncryptKey) -> Vec<u8> {
    let mut buf = vec![0; self.buffer_length()];
    self.to_bytes(&mut buf, sk);
    pk.encrypt(&buf[..])
  }

  // This allocates just one vec for the encrypted data. It's more consistent with the existing to_bytes method but it can panic.
  fn seal_2(&self, buf: &mut [u8], sk: SecretSigningKey, pk: PublicEncryptKey) -> Vec<u8> {
    self.to_bytes(&mut buf, sk);
    pk.encrypt(&buf[..])
  }
```

Apart from that, the caller may decide to do encryption differently (using the new [streaming API](https://docs.rs/sodiumoxide/0.2.6/sodiumoxide/crypto/secretstream/index.html) for instance). So in the end, implementing that method here means we'd have to take some decisions I'd prefer the caller to take.
-

Robert-Steiner (approved):

PR #517, opened 2020-09-14
enable integration tests in CI

Robert-Steiner:
Summary:

* enabled integration tests in ci
(to mark a test as "integration test" you have to add the prefix integration to the function name e.g. integration_set_and_get_coordinator_state)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -120,11 +120,10 @@ jobs:
         working-directory: ./rust
         env:
           RUSTFLAGS: "-D warnings"
-        run: cargo test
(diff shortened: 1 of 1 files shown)


little-dude:
We don't actually have such tests currently right? And just to clarify, by "integration" you mean "tests that have external dependencies" such as redis, influxdb, etc?

Robert-Steiner:
That's right but we will have some after the redis integration. 
> And just to clarify, by "integration" you mean "tests that have external dependencies" such as redis, influxdb, etc?
exactly

little-dude (approved):

PR #518, opened 2020-09-15
remove expected-participants parameter

finiteprods:
Previously, the coordinator used an `expected_participants` parameter (for the expected number of participants) to calculate the model scalar. Now that scalars are determined by the participants (as part of the [generalised scalar extension](#496 )), this parameter is no longer needed.


Robert-Steiner (approved):
Looks good to me👍
Can you please remove the field in the config files as well

Robert-Steiner (approved):

PR #519, opened 2020-09-15
Redis integration

Robert-Steiner:
Moves the `Sum` and `Seed` dict into Redis.

I tried to split the changes into 3 commits to make it easier to review.

# Error handling 

The error handling is kept simple as I'd love to hear your ideas before I implement anything more sophisticated.

If a Redis error occurs during the message processing, we ignore the message.
- [Sum message](https://github.com/xaynetwork/xaynet/pull/519/files#diff-a07f40c3b4b740367f7d4a72897eb2d0R135)
- [Update message](https://github.com/xaynetwork/xaynet/pull/519/files#diff-b15272399ff491c84d93d4ca104eaab1R229)

If a Redis error occurs during the message processing, we ignore the message.

If a Redis error occurs when writing / reading the state that relates to the state machine itself (e.g. `get_seed_dict`, `set_coordinator_state`), we move into the error state.

- [`get_seed_dict`](https://github.com/xaynetwork/xaynet/pull/519/files#diff-b15272399ff491c84d93d4ca104eaab1R70)
- [`set_coordinator_state`](https://github.com/xaynetwork/xaynet/pull/519/files#diff-f0a081898820862a4f5ba67d9ec8d408R49)

The state machine remains in the error state until the Redis client has successfully re-established the connection to the Redis server. 

The [ConnectionManager](https://docs.rs/redis/0.17.0/redis/aio/struct.ConnectionManager.html) who is responsible for reconnecting has an unusual behavior.

> The reconnect code will atomically swap the current (dead) connection with a future that will eventually resolve to a MultiplexedConnection or to a RedisError

> All commands that are issued after the reconnect process has been initiated, will have to await the connection future.

> If reconnecting fails, all pending commands will be failed as well. A new reconnection attempt will be triggered if the error is an I/O error.

Therefore, I had to create this [loop](https://github.com/xaynetwork/xaynet/pull/519/files#diff-7d8af3981570c8116cf43cef52e6344fR61)

### DIFF
--- a/rust/xaynet-server/src/state_machine/mod.rs
+++ b/rust/xaynet-server/src/state_machine/mod.rs
@@ -115,12 +115,13 @@ use self::{
     },
     requests::{RequestReceiver, RequestSender},
 };
-
+use crate::{
+    settings::{MaskSettings, ModelSettings, PetSettings},
+    storage::{redis, redis::RedisError},
+};
 use derive_more::From;
 use thiserror::Error;
 use xaynet_core::{mask::UnmaskingError, InitError};
-
-use crate::settings::{MaskSettings, ModelSettings, PetSettings};
 
 #[cfg(feature = "metrics")]
 use crate::metrics::MetricsSender;
@@ -139,6 +140,9 @@ pub enum StateMachineError {
 
     #[error("the request could not be processed due to an internal error")]
     InternalError,
+
+    #[error("Redis failed: {0}")]
+    Redis(#[from] RedisError),
 }
 
 pub type StateMachineResult = Result<(), StateMachineError>;
@@ -203,6 +207,7 @@ where
         pet_settings: PetSettings,
         mask_settings: MaskSettings,
         model_settings: ModelSettings,
+        redis: redis::Client,
         #[cfg(feature = "metrics")] metrics_tx: MetricsSender,
     ) -> Result<(Self, RequestSender, EventSubscriber), InitError> {
         // crucial: init must be called before anything else in this module
@@ -221,6 +226,7 @@ where
             coordinator_state,
             event_publisher,
             req_receiver,
+            redis,
             #[cfg(feature = "metrics")]
             metrics_tx,
         );
--- a/rust/xaynet-server/src/state_machine/phases/error.rs
+++ b/rust/xaynet-server/src/state_machine/phases/error.rs
@@ -1,8 +1,13 @@
-use crate::state_machine::{
-    phases::{Idle, Phase, PhaseName, PhaseState, Shared, Shutdown},
-    RoundFailed,
-    StateMachine,
+use crate::{
+    state_machine::{
+        phases::{Idle, Phase, PhaseName, PhaseState, Shared, Shutdown},
+        RoundFailed,
+        StateMachine,
+    },
+    storage::redis::RedisError,
 };
+use std::time::Duration;
+use tokio::time::delay_for;
 
 #[cfg(feature = "metrics")]
 use crate::metrics;
@@ -18,6 +23,8 @@ pub enum StateError {
     RoundError(#[from] RoundFailed),
     #[error("state failed: phase timeout: {0}")]
     TimeoutError(#[from] tokio::time::Elapsed),
+    #[error("state failed: Redis failed: {0}")]
+    Redis(#[from] RedisError),
 }
 
 impl PhaseState<StateError> {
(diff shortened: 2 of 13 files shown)


rust/xaynet-server/src/state_machine/mod.rs
```diff
@@ -151,6 +152,9 @@ pub enum StateMachineError {
 
     #[error("the request could not be processed due to an internal error")]
     InternalError,
+
+    #[error("dependency failed: {0}")]
+    DependencyFailed(#[from] RedisError),
```
little-dude, line 157: Maybe we could just name this one `Redis` because it contains a redis error anyway

rust/xaynet-server/src/state_machine/phases/error.rs
```diff
@@ -18,6 +23,8 @@ pub enum StateError {
     RoundError(#[from] RoundFailed),
     #[error("state failed: phase timeout: {0}")]
     TimeoutError(#[from] tokio::time::Elapsed),
+    #[error("state failed: dependency failed: {0}")]
```
little-dude, line 26: Same as above: we might just as well name this `Redis`

rust/xaynet-server/src/state_machine/phases/update.rs
```diff
@@ -223,75 +220,50 @@ impl PhaseState<Update> {
...
     /// # Error
-    /// Fails if it contains invalid keys or it is a repetition.
-    fn add_local_seed_dict(
+    /// Fails if the local dict cannot be added due to a Redis error.
+    async fn add_local_seed_dict(
         &mut self,
         pk: &UpdateParticipantPublicKey,
         local_seed_dict: &LocalSeedDict,
     ) -> Result<(), StateMachineError> {
-        if local_seed_dict.keys().len() == self.inner.frozen_sum_dict.keys().len()
```
little-dude, line 232: If I'm reading `update_seed_dict` correctly, we're actually removing the following checks

Checking that the seed dictionary sent by the participant has the right length:
```rust
local_seed_dict.keys().len() == self.inner.frozen_sum_dict.keys().len()
```

All of the keys of the seed dictionary sent by the participant have a counterpart in the frozen sum dict:

```rust
local_seed_dict.keys().all(|pk| self.inner.frozen_sum_dict.contains_key(pk))
```

The update participant didn't already send an update:

```rust
self.inner.seed_dict.values().next().map_or(true, |dict| !dict.contains_key(pk))
```

The first two checks can be summarized as "the update dict contains exactly one entry for each of the sum participants". I think this check is important, otherwise we may end up with an invalid seed dict. How doable is this with Redis?

little-dude (commented):
Looks very good odewise. However I'm concerned by the removal and sanity checks when updating the seed dict.

rust/xaynet-server/src/state_machine/phases/update.rs
Robert-Steiner: yep, you are right. These checks are what my comment in #416 referred to: 
>
        // This means, that Redis does not abort a transaction if for example the
        // command "hset_nx" returns `0` (value already exists).
        //
        // As a result, this method is successful even though one or all of
        // the following points apply:
        // - `sum_pk` is not in the `sum_dict`
        // - `update_pk` already exists (however, in this case the seed is not updated)
        // - the `LocalSeedDict` has a different length/update_pks than the `sum_dict`.
        //
        // We can create our own transactions via lua scripts. In a script we can
        // perform all the checks first and only write the data if the checks were successful.
        // However, in our case, the checks can be quite expensive.
        // Therefore we should agree on which checks are necessary.

And one solution (not sure if this is the only one) is to use lua scripts. I tried to implement these checks in #409. If you want to take a look, [here is the line where it starts](https://github.com/xaynetwork/xaynet/pull/409/files#diff-a2ad71b28c4ae09ddfa8e791350b705aR10)

little-dude: Ok so it's definitely possible to re-add these checks. They shouldn't be too costly if running in a lua script. Do you have short term plans for adding such a script?

Robert-Steiner: The lua scripts are task I would like to tackle next. I haven't added them yet because our sprint is so short and I didn't know how long the integration would take.

little-dude (approved):

Robert-Steiner:
#build

PR #520, opened 2020-09-15
refactor xaynet_server services

little-dude:
This refactoring started when I tried to split the message parsing
service in several components, some which I could re-use for
implementing a multipart message service. I tried to make some
simplification, and after some yak shaving ended up making quite a lot
of changes. This should make it much easier to implement the multipart
message service though.

### Add a coordinator public key validation service

This is a check we forgot to implement in the past. See also:

- http://world.std.com/~dtd/#sign_encrypt
- https://github.com/xaynetwork/xaynet/pull/514

### `PetError` removal

Only one out of the three variants is used in `xaynet_core` and
`xaynet_server`, and two in `xaynet_client`, so there is no point
sharing that error type accross the crates.

### Remove the tracing machinery

Removed `Request<T>` and the `Traceable` traits. These abstractions
were too pervasive and made the code difficult to maintain. Moreover,
out tracing implementation relied on `tower_tracing` which still
hasn't been released. Removing all these abstractions does makes the
code much more readable, and should make it easier to implement the
multipart service.

### Flatten the service error type

Many services were defined as follow:

```rust
impl Service<MyReq> for MyType {
  type Response = Result<MyResp, MyError>;
  type Error = MyServiceError;
  type Future = MyFuture

  // ..
}
```

The future returned by the service then resolved into a
`Result<Result<MyResp, MyError>, MyServiceError>`. The idea was that
the outer result would be for errors that are specific to the service
machinery, while the inner result type would be for business logic
only.

There are several problems this approach. First it induces complexity
and we quickly end up fighting the type checker when we start messing
around with the services. Second, there's not always a clear
distinction between what is business logic and what is a service
implementation detail. Finally, that's a maintenance burden because it
forces us to maintain several error types, and convert between them.

Therefore, for message services, we unified the error type under
`ServiceError`. Now services are defined as:

```rust
impl Service<MyReq> for MyType {
  type Response = MyResp;
  type Error = MyError;
  type Future = MyFuture

  // ..
}
```

### Split out the decryption service

Our work for supporting multipart messages led us to consider
delegating the encryption layer to TLS at some point. Although there
is no short term plan for that, moving the decryption logic into its
own service lays the ground for this. Moreover, it allowed us to
simplify our tests, because we can now test the message parsing
services with non-encrypted messages


Robert-Steiner (approved):
Good work!

The Refactoring is already paying off. The code was easy to review👍
Regarding the StateMachineError, not sure whether we should split it up into several types later (like messages related errors, state machine errors, ...) because I probably also introduce another error type (redis errors like connection errors) in my redis branch. 

However, (as you mentioned in the pr description) we have seen that it also induces complexity. Maybe something we should as a ticket to the error handling story.

little-dude:
@Robert-Steiner I agree, we should sit together to figure out a error handling strategy. In this case though, introducing an error type for the state machine instead of using `PetError` seems to go in the direction you'd like to head to, no?

finiteprods (approved):
Looks good, the simplification to the tracing and error types very welcome!

little-dude:
> Looks good, the simplification to the tracing and error types very welcome!

Indeed but we also lost some tracing in the process. I'd like to re-introduce that in a less invasive way later.

PR #524, opened 2020-09-18
refactor MaskObject into vector and scalar parts

finiteprods:
The changes introduced for the [generalised scalar extension](#496) had the unfortunate side effect of adding unwelcome duplication to the code base - effectively, a pair of `MaskObject`s were passed through the system, either a (masked model, masked scalar) pair, or their corresponding masks. This in turn lead to the need for a pair of `Aggregation`s, a pair of `MaskDict`s, and so on. Worse, one is a vector and the other is a scalar, but the types don't reflect this at all.

This merge request introduces a new, more type-safe `MaskObject`:

```rust
pub struct MaskObject {
    pub vector: MaskMany,
    pub scalar: MaskOne,
}
```
consisting of a `MaskMany` (the new name for the _old_ `MaskObject`) and a `MaskOne` (containing exactly one data value):

```rust
pub struct MaskOne {
    pub data: BigUint,
    pub config: MaskConfig,
}
``` 

As a result, the scalar-related code is better encapsulated away, for the most part. We get back the tidier code (roughly similar to before #496) - in particular, the new scaling correction step of the generalised extension becomes an implementation detail of the unmasking, as it probably should.

A separate `MaskOne` type also allows a `MaskConfig` to be associated with a scalar - previously, scalars could only be masked according to the same masking configuration as the weights. Although currently, masking configurations are very geared towards the masking of weights, this is nevertheless a necessary first step in achieving more flexibility in the masking of scalars.

### a few extra details...
1. The above refactoring also allows the existing unit tests to more properly cover the scalar extension code. This in turn revealed an issue with (probably) the masking of scalars, as unmasked models no longer match (within the _tolerance_ we expect) the original model weights. As such, the tolerance checks of the masking tests are temporarily relaxed ([commit](https://github.com/xaynetwork/xaynet/pull/524/commits/3710bc2ed26bcfe6806b3500acdbd6ee03d24656)). These should be reverted once the issue has been resolved.
2. The change in structure of `MaskObject` necessitates changes in its (de)serialization. As with other types we serialize, a corresponding "buffer" type `MaskObjectBuffer` is introduced, and similarly `MaskManyBuffer`. But currently, a `MaskOne` is serialized by converting it into a `MaskMany` which was convenient to some extent for purposes of code reuse, but perhaps sub-optimal in other ways. For example, a `MaskOne` can be encoded more compactly than a `MaskMany` - there is always a single data value, so there is no need for a `LENGTH` field. This is an optimization that can be tackled later.
3. Some comments in reviews (pairing of `MaskConfig`s into a type, more consistent naming) will also be dealt with in another PR, probably in conjunction with 1. above.

**Todos**:
- [x] tests still to fix
- [x] conflicts / rebase
- [x] complete the _More to follow_

### DIFF
--- a/rust/xaynet-client/src/participant.rs
+++ b/rust/xaynet-client/src/participant.rs
@@ -217,10 +217,10 @@ impl Participant {
             let mask = seed.derive_mask(mask_len, mask_config, mask_config);
 
             mask_agg
-                .validate_aggregation(&mask.vector)
+                .validate_aggregation(&mask)
                 .map_err(|_| PetError::InvalidMask)?;
 
-            mask_agg.aggregate(mask.vector);
+            mask_agg.aggregate(mask);
         }
         Ok(mask_agg.into())
     }
--- a/rust/xaynet-client/src/participant.rs
+++ b/rust/xaynet-client/src/participant.rs
@@ -9,8 +9,8 @@ use std::default::Default;
 use xaynet_core::{
     crypto::{ByteObject, EncryptKeyPair, SigningKeyPair},
     mask::{
-        Aggregation, BoundType, DataType, GroupType, MaskConfig, MaskMany, MaskObject, MaskSeed,
-        Masker, Model, ModelType,
+        Aggregation, BoundType, DataType, GroupType, MaskConfig, MaskObject, MaskSeed, Masker,
+        Model, ModelType,
     },
     message::{Message, Sum, Sum2, Update},
     CoordinatorPublicKey, InitError, LocalSeedDict, ParticipantPublicKey, ParticipantSecretKey,
@@ -149,12 +149,11 @@ impl Participant {
         mask_len: usize,
     ) -> Result<Message, PetError> {
         let mask_seeds = self.get_seeds(seed_dict)?;
-        let (model_mask, scalar_mask) =
-            self.compute_global_mask(mask_seeds, mask_len, dummy_config())?;
+        let mask = self.compute_global_mask(mask_seeds, mask_len, dummy_config())?;
         let payload = Sum2 {
             sum_signature: self.sum_signature,
-            model_mask,
-            scalar_mask,
+            model_mask: mask.vector,
+            scalar_mask: mask.scalar.into(),
         };
         Ok(Message::new_sum2(self.pk, coordinator_pk, payload))
     }
@@ -206,27 +205,24 @@ impl Participant {
         mask_seeds: Vec<MaskSeed>,
         mask_len: usize,
         mask_config: MaskConfig,
-    ) -> Result<(MaskMany, MaskMany), PetError> {
+        // FIXME also need scalar mask config somehow
+    ) -> Result<MaskObject, PetError> {
         if mask_seeds.is_empty() {
             return Err(PetError::InvalidMask);
         }
 
-        let mut model_mask_agg = Aggregation::new(mask_config, mask_len);
-        let mut scalar_mask_agg = Aggregation::new(mask_config, 1);
+        // HACK reuse model mask config for now
+        let mut mask_agg = Aggregation::new(mask_config, mask_config, mask_len);
         for seed in mask_seeds.into_iter() {
-            let (model_mask, scalar_mask) = seed.derive_mask(mask_len, mask_config);
+            let mask = seed.derive_mask(mask_len, mask_config, mask_config);
 
-            model_mask_agg
-                .validate_aggregation(&model_mask)
-                .map_err(|_| PetError::InvalidMask)?;
-            scalar_mask_agg
-                .validate_aggregation(&scalar_mask)
+            mask_agg
+                .validate_aggregation(&mask.vector)
                 .map_err(|_| PetError::InvalidMask)?;
 
-            model_mask_agg.aggregate(model_mask);
-            scalar_mask_agg.aggregate(scalar_mask);
+            mask_agg.aggregate(mask.vector);
         }
-        Ok((model_mask_agg.into(), scalar_mask_agg.into()))
+        Ok(mask_agg.into())
     }
 }
 
--- a/rust/xaynet-client/src/participant.rs
+++ b/rust/xaynet-client/src/participant.rs
@@ -5,17 +5,31 @@
 //! [client module]: ../index.html
 
 use std::default::Default;
-// TODO may remove MaskMany later
 use xaynet_core::{
     crypto::{ByteObject, EncryptKeyPair, SigningKeyPair},
     mask::{
-        Aggregation, BoundType, DataType, GroupType, MaskConfig, MaskObject, MaskSeed, Masker,
-        Model, ModelType,
+        Aggregation,
+        BoundType,
+        DataType,
+        GroupType,
+        MaskConfig,
+        MaskObject,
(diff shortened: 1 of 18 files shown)


rust/xaynet-core/src/mask/masking.rs
```diff
@@ -76,10 +94,10 @@ impl Into<MaskObject> for Aggregation {
 #[allow(clippy::len_without_is_empty)]
 impl Aggregation {
     /// Creates a new, empty aggregator for masks or masked models.
-    pub fn new(config: MaskConfig, object_size: usize) -> Self {
+    pub fn new(config_many: MaskConfig, config_one: MaskConfig, object_size: usize) -> Self {
```
little-dude, line 97: Not sure if it's worth it but maybe we could have a 

```rust
struct AggregationConfig {
  model: MaskConfig,
  scalar: MaskConfig,
}
```

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -122,12 +145,22 @@ impl Aggregation {
...
 
-        if self.nb_models > self.object.config.model_type.max_nb_models() {
+        if self.nb_models > self.object.vector.config.model_type.max_nb_models() {
             return Err(UnmaskingError::TooManyModels);
         }
 
-        if self.object.config != mask.config || self.object_size != mask.data.len() {
-            return Err(UnmaskingError::MaskMismatch);
+        if self.nb_models > self.object.scalar.config.model_type.max_nb_models() {
+            return Err(UnmaskingError::TooManyScalars);
```
little-dude, line 153: Can `self.object.vector.config.model_type.max_nb_models()` and `self.object.scalar.config.model_type.max_nb_models()` actually differ? Seems like the max is actually capped to the min of these values. More generally, can the scalar masking config and model masking config have incompatibilities?

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -39,14 +50,21 @@ pub enum UnmaskingError {
...
+    // TODO rename Model -> Vector; or use MaskMany/One terminology
+    #[error("the object to aggregate is invalid")]
+    InvalidObject,
 
     #[error("too many models were aggregated for the current unmasking configuration")]
     TooManyModels,
 
-    #[error("the model to aggregate is incompatible with the current aggregated model")]
+    #[error("too many scalars were aggregated for the current unmasking configuration")]
+    TooManyScalars,
```
little-dude, line 61: Is it worth distingushing between `TooManyModels` and `TooManyScalars`? Could they be unified into an `UnmaskingError::TooManyMasks` for instance?

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -122,12 +145,22 @@ impl Aggregation {
...
-        if self.object.config != mask.config || self.object_size != mask.data.len() {
-            return Err(UnmaskingError::MaskMismatch);
+        if self.nb_models > self.object.scalar.config.model_type.max_nb_models() {
+            return Err(UnmaskingError::TooManyScalars);
+        }
+
+        if self.object.vector.config != mask.vector.config
+            || self.object_size != mask.vector.data.len()
+        {
+            return Err(UnmaskingError::MaskManyMismatch);
```
little-dude, line 159: The naming is a bit inconsistent between model/scalar vs many/one. One one side we have `TooManyScalars` and `TooManyModels`, and here `MaskManyMismatch` and `MaskOneMismatch`. There are other places where that is the case. Is the motivation for introducing the many/one terminology to be broader than the ML specific term "model"? I like the idea but I think we should have a clear separation of where we use a specific terminology. For instance we could say that in `xaynet-core` there shouldn't be any mention of "model", while in `xaynet-client` and `xaynet-server` which are domain specific, we should only use the model/scalar terminology (it would even be better if we could find a more specific term than "scalar" imo).

On the same topic, I have a slight preference for vector/scalar over many/one. It's not domain specific but fits better the nature of the data we're referring to. Wdyt?

little-dude (commented):
Very nice. My only concern is the naming, which I find slightly inconsistent sometimes.

rust/xaynet-client/src/participant.rs
```diff
@@ -5,7 +5,7 @@
 //! [client module]: ../index.html
 
 use std::default::Default;
-
+// TODO may remove MaskMany later
```
Robert-Steiner, line 8: What does this comment refer to?

Robert-Steiner (commented):
Really nice refactoring👍
This will make the extracting of the mask dictionary into Redis much easier

rust/xaynet-core/src/mask/masking.rs
finiteprods: I think this would further increase tidiness, thanks. will add this in a later PR if that's ok.

rust/xaynet-core/src/mask/masking.rs
finiteprods: I agree this could probably done in one check. more generally, since the scalar could be masked with a different config to weights, we should do the same check to make sure it admits all of them for aggregating. will address separately.

rust/xaynet-core/src/mask/masking.rs
finiteprods: I guess distinguishing the cases has the advantage when we get errors, e.g. to narrow down which mask config contributed to the error.

rust/xaynet-client/src/participant.rs
finiteprods: a stale comment, now removed. Thanks for spotting!

rust/xaynet-core/src/mask/masking.rs
finiteprods: yes I agree the naming isn't the best at the moment. OT1H I had introduced new terminology, OTOH I was also trying to not deviate too much from the existing names here for these errors (which are not named the best either!). Let me revise this separately if that's ok and we can continue the discussion there. And yes, the idea of Many vs One was to more abstractly capture the structure of masked models / scalars and model masks / scalar masks. Vector / Scalar would also have worked but scalar clashes with the more domain specific use of the term, as you also noted.

rust/xaynet-server/src/state_machine/phases/sum2.rs
```diff
@@ -249,19 +211,21 @@ mod test {
         let msg =
             updater.compose_update_message(coord_keys.public, &sum_dict, scalar, model.clone());
         let masked_model = utils::masked_model(&msg);
-        let masked_scalar = utils::masked_scalar(&msg);
+        //let masked_scalar = utils::masked_scalar(&msg);
```
Robert-Steiner, line 214: I think we can completely remove the commented out lines

rust/xaynet-core/src/mask/masking.rs
little-dude: So this could be tackled with https://github.com/xaynetwork/xaynet/pull/524#discussion_r490782543 right?

Robert-Steiner (commented):
The changes look good to me

little-dude (approved):
:+1: Very nice cleanup. Could we just tidy the git history a bit before we merge?

PR #526, opened 2020-09-22
Redis script for seed update operation

Robert-Steiner:
**Summary**

- added lua script for seed update operation
- added `SeedDictUpdateError`
- refactored `DeleteSumParticipant`(I will adjust `AddSumParticipant` in a separate pr)
- increased test coverage for redis client
- annotated unused client methods with `#[cfg(test)]` (they are still useful for testing)

### DIFF
--- a/rust/xaynet-server/src/storage/impls.rs
+++ b/rust/xaynet-server/src/storage/impls.rs
@@ -1,7 +1,8 @@
 use crate::state_machine::coordinator::CoordinatorState;
-use derive_more::{From, Into};
+use derive_more::{Deref, From, Into};
 use paste::paste;
 use redis::{ErrorKind, FromRedisValue, RedisError, RedisResult, RedisWrite, ToRedisArgs, Value};
+use thiserror::Error;
 use xaynet_core::{
     crypto::{ByteObject, PublicEncryptKey, PublicSigningKey},
     mask::{EncryptedMaskSeed, MaskMany},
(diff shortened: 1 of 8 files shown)


rust/xaynet-server/src/storage/impls.rs
```diff
@@ -206,21 +208,115 @@ impl FromRedisValue for AddSumParticipant {
...
 }
 
-impl FromRedisValue for DeleteSumParticipant {
-    fn from_redis_value(v: &Value) -> RedisResult<DeleteSumParticipant> {
+impl<'a> ToRedisArgs for &'a LocalSeedDictWrite<'a> {
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + RedisWrite,
+    {
+        let args: Vec<(PublicSigningKeyWrite, EncryptedMaskSeedWrite)> = self
```
little-dude, line 239: Do we actually need the type annotation here? Seems like the `(PublicSigningKeyWrite::from(pk),EncryptedMaskSeedWrite::from(seed))` in the `map` should be enought?

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -258,7 +197,60 @@ impl Connection {
...
+        update_pk: &UpdateParticipantPublicKey,
+        local_seed_dict: &LocalSeedDict,
+    ) -> RedisResult<SeedDictUpdate> {
+        let script = Script::new(
+            r#"
+                local update_pk = ARGV[1]
+
+                -- check if the local seed dict has the same length as the sum_dict
+
+                -- KEYS is a list (table) of key value pairs ([update_pk_1, seed_1, update_pk_2, seed_2])
```
little-dude, line 212: Where does `KEYS` come from?

little-dude: Sorry I found the answer in the redis documentation: https://redis.io/commands/eval

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -258,7 +197,60 @@ impl Connection {
...
+                for i = 1, #KEYS, 2 do
+                    redis.call("HSETNX", KEYS[i], update_pk, KEYS[i + 1])
+                end
+
+                return 0
+            "#,
+        );
+
+        script
+            .key(LocalSeedDictWrite::from(local_seed_dict))
```
little-dude, line 244: I'm a little confused by this. Is this `LocalSeedDictWrite` automatically converted into multiple keys like `EVAL "<script>" key1 key2 key3 arg`? Or is it just one key that has the "list" type, like `EVAL "<script>" "key1,key2,key3" arg`?

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -258,7 +197,60 @@ impl Connection {
...
+                -- check if one pk of the local seed dict already exists in seed_dict
+                local exist_in_seed_dict = redis.call("SADD", "update_participants", update_pk)
+                -- SADD returns 0 if the key already exists
+                if exist_in_seed_dict == 0 then
+                    return -3
+                end
+
+                -- update the seed dict
+                for i = 1, #KEYS, 2 do
+                    redis.call("HSETNX", KEYS[i], update_pk, KEYS[i + 1])
```
little-dude, line 236: Although this _shouldn't_ fail, maybe we could check for an error here as well? Even though I don't really know how we'd handle it, I think it would be nice to just log it, so that we know it happened.

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: there are multiple keys: `EVAL "<script>" key1 key2 key3 arg`
I have kept the usage of `KEYS` and `ARGV` even though we can't use it for redis-cluster, because this way it is easier to distinguish between the `local seed dict` and `the update_pk`.

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: Good point👍 I will add that

rust/xaynet-server/src/storage/impls.rs
Robert-Steiner: Unfortunately yes, otherwise I get an error at `args.write_redis_args(out)`
But I can write
```rust
            .map(|(pk, seed)| {
                (
                    From::from(pk),
                    From::from(seed),
                )
            })
```
instead of 
```rust
            .map(|(pk, seed)| {
                (
                    PublicSigningKeyWrite::from(pk),
                    EncryptedMaskSeedWrite::from(seed),
                )
            })
```
to make it a bit shorter

little-dude (approved):

PR #528, opened 2020-09-23
XN-904 tls server auth

janpetschexain:
Optional TLS server authentication:
- enable TLS on server and client side, make certificates optional (i.e. http if no certs are provided and https if certs are provided)
- update the ffi code for the mobile client
- update the server and client examples (trivially without certs for now)
- add some error handling on client initialization
- make certificate inputs configurable via paths for server and client
- gate the certificates behind a `tls` feature flag

We need to generate some actual certificates to run this stuff in a testing session, but there is a separate ticket for that.

### DIFF
--- a/rust/examples/test-drive-net.rs
+++ b/rust/examples/test-drive-net.rs
@@ -50,7 +50,11 @@ async fn main() -> Result<(), ClientError<HttpApiClientError>> {
 
     let mut clients = Vec::with_capacity(opt.nb_client as usize);
     for id in 0..opt.nb_client {
-        let mut client = Client::new(opt.period, id, HttpApiClient::new(&opt.url))?;
+        let mut client = Client::new(
+            opt.period,
+            id,
+            HttpApiClient::new(&opt.url, None).map_err(ClientError::Api)?,
+        )?;
         client.local_model = Some(model.clone());
         let join_hdl = tokio::spawn(async move {
             tokio::select! {
--- a/rust/examples/test-drive-net.rs
+++ b/rust/examples/test-drive-net.rs
@@ -38,20 +38,12 @@ struct Opt {
     nb_client: u32,
 
     #[structopt(
-        short = "der",
+        short,
         long,
         parse(from_os_str),
-        help = "The list of trusted DER encoded TLS server certificates"
+        help = "The list of trusted DER and PEM encoded TLS server certificates"
     )]
-    der_certificates: Option<Vec<PathBuf>>,
-
-    #[structopt(
-        short = "pem",
-        long,
-        parse(from_os_str),
-        help = "The list of trusted PEM encoded TLS server certificates"
-    )]
-    pem_certificates: Option<Vec<PathBuf>>,
+    certificates: Option<Vec<PathBuf>>,
 }
 
 /// Test-drive script of a (local, but networked) federated
@@ -71,8 +63,7 @@ async fn main() -> Result<(), ClientError<HttpApiClientError>> {
     let model = Model::from_primitives(vec![0; len].into_iter()).unwrap();
 
     let certificates =
-        HttpApiClient::certificates_from_paths(opt.der_certificates, opt.pem_certificates)
-            .map_err(ClientError::Api)?;
+        HttpApiClient::certificates_from(&opt.certificates).map_err(ClientError::Api)?;
     let mut clients = Vec::with_capacity(opt.nb_client as usize);
     for id in 0..opt.nb_client {
         let mut client = Client::new(
--- a/rust/examples/test-drive-net.rs
+++ b/rust/examples/test-drive-net.rs
@@ -37,13 +37,16 @@ struct Opt {
     #[structopt(default_value = "10", short, help = "The number of clients")]
     nb_client: u32,
 
-    #[structopt(
-        short,
-        long,
-        parse(from_os_str),
-        help = "The list of trusted DER and PEM encoded TLS server certificates"
+    #[cfg_attr(
+        feature = "tls",
+        structopt(
+            short,
+            long,
+            parse(from_os_str),
+            help = "The list of trusted DER/PEM encoded TLS server certificates"
+        )
     )]
-    certificates: Option<Vec<PathBuf>>,
+    certificates: Vec<PathBuf>,
 }
 
 /// Test-drive script of a (local, but networked) federated
@@ -62,14 +65,21 @@ async fn main() -> Result<(), ClientError<HttpApiClientError>> {
     let len = opt.len as usize;
     let model = Model::from_primitives(vec![0; len].into_iter()).unwrap();
 
+    #[cfg(feature = "tls")]
     let certificates =
         HttpApiClient::certificates_from(&opt.certificates).map_err(ClientError::Api)?;
+
     let mut clients = Vec::with_capacity(opt.nb_client as usize);
     for id in 0..opt.nb_client {
         let mut client = Client::new(
             opt.period,
             id,
-            HttpApiClient::new(&opt.url, certificates.clone()).map_err(ClientError::Api)?,
+            HttpApiClient::new(
+                &opt.url,
+                #[cfg(feature = "tls")]
+                certificates.clone(),
+            )
+            .map_err(ClientError::Api)?,
         )?;
         client.local_model = Some(model.clone());
         let join_hdl = tokio::spawn(async move {
--- a/rust/examples/test-drive-net.rs
+++ b/rust/examples/test-drive-net.rs
@@ -1,9 +1,12 @@
+use std::path::PathBuf;
+
 #[macro_use]
 extern crate tracing;
 
 use structopt::StructOpt;
 use tokio::signal;
 use tracing_subscriber::*;
+
 use xaynet_client::{
     api::{HttpApiClient, HttpApiClientError},
     Client,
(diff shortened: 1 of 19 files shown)


rust/examples/test-drive-net.rs
```diff
@@ -20,16 +23,35 @@ struct Opt {
...
         short,
         help = "The time period at which to poll for service data, in seconds"
     )]
     period: u64,
+
     #[structopt(default_value = "10", short, help = "The number of clients")]
     nb_client: u32,
+
+    #[structopt(
+        short = "der",
```
little-dude, line 41: Do we need a setting for der certs and one for pem certs? Couldn't we just check the extension?

little-dude (commented):
The `xaynet_server` changes look good to me. For the client side though, I don't understand why we need to:

- differentiate between der and pem certificate
- specify potentially multiple certificates

rust/examples/test-drive-net.rs
janpetschexain: good idea, i'll change that.

janpetschexain:
> The `xaynet_server` changes look good to me. For the client side though, I don't understand why we need to:
> 
> * differentiate between der and pem certificate

haven't thought of extension checking, i'll update this.

> * specify potentially multiple certificates

a likely use case would be a client that does federated learning and federated analytics and talks to two different coordinators for that reason. i know that the current client doesn't support this yet, but this way it is easier to extend in the future.

rust/xaynet-client/src/api/http.rs
```diff
@@ -21,13 +23,68 @@ pub struct HttpApiClient {
...
+        let client = if let Some(certificates) = certificates {
+            let mut builder = ClientBuilder::new().use_rustls_tls();
+            for certificate in certificates {
+                builder = builder.add_root_certificate(certificate);
+            }
+            builder.build().map_err(HttpApiClientError::Http)?
+        } else {
+            ClientBuilder::new()
+                .use_rustls_tls()
+                .danger_accept_invalid_certs(true)
```
Robert-Steiner, line 46: No sure if I understand that `else` branch correctly. 
If we don't use certificated, we allow invalid certificates? 
Can't we just say: if no certificate is provided, only http is supported?

rust/xaynet-client/src/mobile_client/mod.rs
```diff
@@ -49,6 +54,8 @@ impl MobileClient {
     /// Fails if the crypto module cannot be initialized.
     pub fn init(
         url: &str,
+        der_certificates: Option<Vec<PathBuf>>,
```
Robert-Steiner, line 57: Maybe we should move them at the end of the parameters list because both are optional

rust/xaynet-server/src/settings.rs
```diff
@@ -566,3 +568,140 @@ where
 
     deserializer.deserialize_str(EnvFilterVisitor)
 }
+
+#[derive(Debug, Deserialize, Validate)]
+#[validate(schema(function = "validate_tls"))]
+/// TLS settings.
+pub struct TlsSettings {
```
Robert-Steiner, line 575: I would prefer to put these into the api settings.

rust/xaynet-server/src/rest.rs
```diff
@@ -1,22 +1,29 @@
...
 /// Starts a HTTP server at the given address, listening to GET requests for
 /// data and POST requests containing PET messages.
 ///
 /// * `addr`: address of the server.
+/// * `tls_settings`: optional certificate and key for TLS server authentication.
 /// * `fetcher`: fetcher for responding to data requests.
 /// * `pet_message_handler`: handler for responding to PET messages.
 pub async fn serve<F>(
     addr: impl Into<SocketAddr> + 'static,
+    tls_settings: TlsSettings,
```
Robert-Steiner, line 26: If we put the tls setting into the api setting, we can simplify this part a bit

```
pub async fn serve<F>(
    api_settings: ApiSettings,
    fetcher: F,
    pet_message_handler: PetMessageHandler,
)
```

rust/xaynet-server/src/settings.rs
```diff
@@ -566,3 +568,140 @@ where
...
+        match (cert, key) {
+            (Some(cert), Some(key)) => Some((PathBuf::from(cert), PathBuf::from(key))),
+            (None, None) => None,
+            _ => unreachable!("mixed cases are prevented by validation"),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Robert-Steiner, line 634: 👍

Robert-Steiner (commented):
Looks good so far 
I added some comments

I think we need to sync up with @rsaffi because i'm not sure if this change works out of the box in our k8s cluster. In our current setup `nginx` handles tls if i'm not mistaken.

rsaffi:
> Looks good so far
> I added some comments
> 
> I think we need to sync up with @rsaffi because i'm not sure if this change works out of the box in our k8s cluster. In our current setup `nginx` handles tls if i'm not mistaken.

I briefly chatted about this with @janpetschexain. So: on our k8s cluster right now we do have SSL termination for the coordinator ingress, and we're relying on letsencrypt certificate for that. But that's one thing.
As @janpetschexain pointed out, not everybody will be relying on k8s and the ease of deploying a valid certificate, so it still makes sense to provide the functionality inside the coordinator, not only for its external ingress service.

There's a lot related to using self-signed certs, so other topics come into play, like [Certificate Pinning](https://carvesystems.com/news/cert_pin/) (on the client's side).

This will all come together with the implementation of our own CA (soon).

rust/xaynet-client/src/api/http.rs
janpetschexain: this issue comes from the fact that `reqwest` now has the `tls` feature enabled, which causes the `Client` to have a vector of certificates. i *assume*  that every connection will fail if this vector is empty, therefore i set it to accept everything (as it would if there was no tls enabled).

i agree that this is somewhat suboptimal and potentially confusing. a better solution would be to introduce a feature flag for tls ourselves and to gate the respective parts behind config attributes. this would also allow us to remove the `Option<>`s around the certificate paths to clean up some code. should we do so right away or in a future ticket?

rust/xaynet-client/src/mobile_client/mod.rs
janpetschexain: agree.

rust/xaynet-server/src/settings.rs
janpetschexain: i'll move them.

rust/xaynet-server/src/rest.rs
janpetschexain: agree.

rust/xaynet-client/src/api/http.rs
Robert-Steiner: Good point. A feature flag on the client side would make sense I think👍 
> should we do so right away or in a future ticket? 

I'm fine if we do it later.

rust/xaynet-client/src/api/http.rs
```diff
@@ -21,13 +23,65 @@ pub struct HttpApiClient {
...
+    pub fn certificates_from(
+        paths: &Option<Vec<PathBuf>>,
+    ) -> Result<Option<Vec<Certificate>>, HttpApiClientError> {
+        if let Some(paths) = paths {
+            if paths.is_empty() {
+                Ok(None)
+            } else {
+                paths
+                    .iter()
+                    .map(|path| -> Result<Certificate, HttpApiClientError> {
```
little-dude, line 67: I think it's worth moving this closure into a free standing function. `load_cert()` or something like this.

rust/xaynet-client/src/api/http.rs
```diff
@@ -21,13 +23,65 @@ pub struct HttpApiClient {
...
+                            |extension| match extension.to_str() {
+                                Some("der") => Certificate::from_der(&encoding)
+                                    .map_err(HttpApiClientError::Http),
+                                Some("pem") => Certificate::from_pem(&encoding)
+                                    .map_err(HttpApiClientError::Http),
+                                _ => Err(HttpApiClientError::UnexpectedCertificate),
+                            },
+                        )
+                    })
+                    .collect::<Result<Vec<_>, HttpApiClientError>>()
```
little-dude, line 80: Nice I didn't know `collect` could do that. I assume it stops on the first error?

rust/xaynet-client/src/api/http.rs
```diff
@@ -21,13 +23,65 @@ pub struct HttpApiClient {
...
+            client,
             address: address.into(),
+        })
+    }
+
+    /// Reads DER and PEM certificates from given paths.
+    pub fn certificates_from(
+        paths: &Option<Vec<PathBuf>>,
+    ) -> Result<Option<Vec<Certificate>>, HttpApiClientError> {
+        if let Some(paths) = paths {
```
little-dude, line 61: Personal taste, but I think it'd be nicer to get rid of all the nesting with early returns and unwraps.

```rust
if paths.is_none() {
    return Ok(None);
}

let paths = paths.unwrap();
if paths.is_empty() {
    return Ok(None);
}

paths.iter().map(.....)
```

rust/xaynet-client/src/api/http.rs
janpetschexain: it short-circuits and returns the first encountered error, that's pretty neat indeed :) should work similarly with vecs and options as well.

little-dude: `Iterator` is a little gem ^^

rust/xaynet-client/src/api/http.rs
janpetschexain: agree.

rust/xaynet-client/src/api/http.rs
janpetschexain: i now cleaned it up and simplified this a bit during the feature flag introduction.

rust/xaynet-client/src/api/http.rs
janpetschexain: i couldn't let go, so it is part of this pr now as well :D looks nicer imo

little-dude (approved):

rust/xaynet-server/Cargo.toml
```diff
@@ -40,10 +40,9 @@ bincode = "1.3.1"
 thiserror = "1.0.20"
 anyhow = "1.0.32"
 bitflags = "1.2.1"
-warp = "0.2.5"
+warp = { version = "0.2.5" }
```
Robert-Steiner, line 43:
```suggestion
warp = "0.2.5"
```

Robert-Steiner (approved):

finiteprods (approved):

rust/xaynet-server/Cargo.toml
janpetschexain: changed.

PR #530, opened 2020-09-25
Implement message parsing from byte streams

little-dude:
- rename `FromBytes::from_bytes` into `FromBytes::from_byte_slice`
- add `FromBytes::from_byte_stream`
- add a chunkable iterator under `xaynet_core::messages::utils`, that is used internally

### DIFF
--- a/rust/xaynet-core/src/mask/object/serialization.rs
+++ b/rust/xaynet-core/src/mask/object/serialization.rs
@@ -16,7 +16,7 @@ use crate::{
     },
     message::{
         traits::{FromBytes, ToBytes},
-        utils::range,
+        utils::{range, ChunkableIterator},
         DecodeError,
     },
 };
@@ -102,6 +102,21 @@ impl<T: AsRef<[u8]>> MaskManyBuffer<T> {
             ));
         }
 
+        let total_expected_length = self.try_len()?;
+        if len < total_expected_length {
+            return Err(anyhow!(
+                "invalid buffer length: expected {} bytes but buffer has only {} bytes",
+                total_expected_length,
+                len
+            ));
+        }
+        Ok(())
+    }
+
+    /// Return the expected length of the underlying byte buffer,
+    /// based on the masking config field of numbers field. This is
+    /// similar to [`len`] but cannot panic.
+    fn try_len(&self) -> Result<usize, DecodeError> {
         let config =
             MaskConfig::from_byte_slice(&self.config()).context("invalid MaskObject buffer")?;
         let bytes_per_number = config.bytes_per_number();
@@ -111,15 +126,7 @@ impl<T: AsRef<[u8]>> MaskManyBuffer<T> {
                 "invalid MaskObject buffer: invalid masking config or numbers field"
             ));
         }
-        let total_expected_length = NUMBERS_FIELD.end + data_length;
-        if len < total_expected_length {
-            return Err(anyhow!(
-                "invalid buffer length: expected {} bytes but buffer has only {} bytes",
-                total_expected_length,
-                len
-            ));
-        }
-        Ok(())
+        Ok(NUMBERS_FIELD.end + data_length)
     }
 
     /// Gets the expected number of bytes of this buffer wrt to the masking configuration.
@@ -248,6 +255,38 @@ impl FromBytes for MaskMany {
         }
 
         Ok(MaskMany { data, config })
+    }
+
+    fn from_byte_stream<I: Iterator<Item = u8> + ExactSizeIterator>(
+        iter: &mut I,
+    ) -> Result<Self, DecodeError> {
+        let config = MaskConfig::from_byte_stream(iter)?;
+        if iter.len() < 4 {
+            return Err(anyhow!("byte stream exhausted"));
+        }
+        let numbers = u32::from_byte_stream(iter)
+            .context("failed to parse the number of item in mask object")?;
+        let bytes_per_number = config.bytes_per_number();
+
+        let data_len = numbers as usize * bytes_per_number;
+        if iter.len() < data_len {
+            return Err(anyhow!(
+                "mask object is {} bytes long but byte stream only has {} bytes",
+                data_len,
+                iter.len()
+            ));
+        }
+
+        let mut data = Vec::with_capacity(numbers as usize);
+        let mut buf = vec![0; bytes_per_number];
+        for chunk in iter.take(data_len).chunks(bytes_per_number).into_iter() {
+            for (i, b) in chunk.enumerate() {
+                buf[i] = b;
+            }
+            data.push(BigUint::from_bytes_le(buf.as_slice()));
+        }
+
+        Ok(MaskObject { data, config })
     }
 }
 
@@ -382,6 +421,14 @@ pub(crate) mod tests {
     }
 
     #[test]
+    fn deserialize_stream() {
+        assert_eq!(
+            MaskObject::from_byte_stream(&mut bytes().into_iter()).unwrap(),
+            object()
+        );
+    }
+
+    #[test]
     fn serialize_1() {
         let mut buf = vec![0xff; 14];
         object_1().to_bytes(&mut buf);
@@ -391,5 +438,13 @@ pub(crate) mod tests {
     #[test]
     fn deserialize_1() {
         assert_eq!(MaskMany::from_byte_slice(&bytes_1()).unwrap(), object_1());
+    }
+
+    #[test]
+    fn deserialize_stream_1() {
+        assert_eq!(
+            MaskObject::from_byte_stream(&mut bytes_1().into_iter()).unwrap(),
+            object_1()
+        );
     }
 }
--- a/rust/xaynet-core/src/mask/object/serialization.rs
+++ b/rust/xaynet-core/src/mask/object/serialization.rs
@@ -265,7 +265,7 @@ impl FromBytes for MaskMany {
             return Err(anyhow!("byte stream exhausted"));
         }
         let numbers = u32::from_byte_stream(iter)
-            .context("failed to parse the number of item in mask object")?;
+            .context("failed to parse the number of items in mask object")?;
         let bytes_per_number = config.bytes_per_number();
 
         let data_len = numbers as usize * bytes_per_number;
@@ -286,7 +286,7 @@ impl FromBytes for MaskMany {
             data.push(BigUint::from_bytes_le(buf.as_slice()));
         }
 
-        Ok(MaskObject { data, config })
+        Ok(MaskMany { data, config })
     }
 }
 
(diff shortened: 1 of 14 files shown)


rust/xaynet-core/src/mask/object/serialization.rs
```diff
@@ -202,6 +210,38 @@ impl FromBytes for MaskObject {
 
         Ok(MaskObject { data, config })
     }
+
+    fn from_byte_stream<I: Iterator<Item = u8> + ExactSizeIterator>(
+        iter: &mut I,
+    ) -> Result<Self, DecodeError> {
+        let config = MaskConfig::from_byte_stream(iter)?;
+        if iter.len() < 4 {
```
Robert-Steiner, line 218: I'm not sure that this line will ever be executed because by calling `MaskConfig::from_byte_stream(iter)?`
`MaskConfigBuffer::check_buffer_length` would already return the error `invalid buffer length`.

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
```diff
@@ -0,0 +1,480 @@
...
+where
+    I: Iterator,
+{
+    /// The iterator we're chunking
+    iter: I,
+    /// Size of each chunk. Note that the last chunk may be smaller
+    chunk_size: usize,
+    /// Number of chunks that have been yielded
+    nb_chunks: usize,
+    /// Next item from `iter`. By buffering it, we can now when `iter`
```
Robert-Steiner, line 101: a word missing here

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
```diff
@@ -0,0 +1,480 @@
...
+{
+    /// Return a new `Inner` with the given iterator and chunk size
+    fn new(mut iter: I, chunk_size: usize) -> Self {
+        if chunk_size == 0 {
+            panic!("invalid chunk size (must be > 0)")
+        }
+        let next = iter.next().map(|elt| (0, elt));
+        Self {
+            iter,
+            // index: 0,
```
Robert-Steiner, line 143: Are these lines still important?

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
```diff
@@ -0,0 +1,480 @@
...
+        start..end
+    }
+
+    /// Return `true` if the iterator we're chunking is exhausted
+    fn exhausted(&self) -> bool {
+        self.inner.borrow().next.is_none()
+    }
+
+    /// Get the `index`-th item from the underlying iterator. If the
+    /// iterator already advanced beyon `index`, `None` is
```
Robert-Steiner, line 217:
```suggestion
    /// iterator already advanced beyond `index`, `None` is
```

Robert-Steiner (commented):
For me it looks good! 
I really like the code documentation with the examples 👍

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
little-dude: I think I meant to write "know"

rust/xaynet-core/src/mask/object/serialization.rs
little-dude: No, because `iter` is mutated. `MaskConfig::from_byte_stream(iter)?` consumes a few bytes, and then we check that there are at least 4 bytes left in `iter`.

rust/xaynet-core/src/message/traits.rs
```diff
@@ -301,6 +313,43 @@ impl FromBytes for LocalSeedDict {
...
+
+        let mut dict = LocalSeedDict::new();
+        let entries = iter.take(len - 4).chunks(ENTRY_LENGTH);
+        for mut chunk in entries.into_iter() {
+            let key = SumParticipantPublicKey::from_byte_stream(&mut chunk)
+                .context("invalid entry: cannot parse public key")?;
+            let value = EncryptedMaskSeed::from_byte_stream(&mut chunk)
+                .context("invalid entry: cannot parse encrypted mask seed")?;
+            // This should really not happen, but it's worth checking
+            // because out chunkable iterator panics if the chunks are
```
little-dude, line 340:
```suggestion
            // because our chunkable iterator panics if the chunks are
```

rust/xaynet-core/src/mask/object/serialization.rs
Robert-Steiner: Ah you are right👍 my bad

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
little-dude: No that's a leftover. I removed them.

rust/xaynet-core/src/message/mod.rs
```diff
@@ -29,7 +29,11 @@
 pub(crate) mod message;
 pub(crate) mod payload;
 pub(crate) mod traits;
-pub(crate) mod utils;
+
+// FIXME: I'd like to make this `pub(crate)` but then the doc-tests in
+// utils::chunkable_iterator cannot be compiled
+#[doc(hidden)]
+pub mod utils;
```
janpetschexain, line 36: i'd say it is probably better to mark the code in the doctests as `ignore` and to move their content to some unit tests where we can run them on `pub(crate)` imports. otherwise we would expose internal details, which seems to be a worse tradeoff than some test duplication.

rust/xaynet-core/src/message/payload/update.rs
```diff
@@ -280,11 +279,11 @@ pub(in crate::message) mod tests_helpers {
     use std::convert::TryFrom;
 
     use super::*;
-    use crate::{
-        crypto::ByteObject,
-        mask::{object::MaskMany, seed::EncryptedMaskSeed},
-        SumParticipantPublicKey,
+    pub(in crate::message) use crate::mask::object::serialization::tests::{
+        mask_many as masked_model,
+        mask_one as masked_scalar,
```
finiteprods, line 284: I think it would be better to bring in `mask_object` instead and use that in the below

finiteprods (commented):
just a few minor comments re the mask serialization tests, so far.

rust/xaynet-core/src/message/mod.rs
```diff
@@ -29,7 +29,11 @@
 pub(crate) mod message;
 pub(crate) mod payload;
 pub(crate) mod traits;
-pub(crate) mod utils;
+
+// FIXME: I'd like to make this `pub(crate)` but then the doc-tests in
+// utils::chunkable_iterator cannot be compiled
```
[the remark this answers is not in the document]
little-dude: nope =/

rust/xaynet-core/src/mask/object/serialization.rs
```diff
@@ -249,6 +256,38 @@ impl FromBytes for MaskMany {
...
+            return Err(anyhow!("byte stream exhausted"));
+        }
+        let numbers = u32::from_byte_stream(iter)
+            .context("failed to parse the number of item in mask object")?;
```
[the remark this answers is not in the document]
little-dude: In the code we use `numbers` because it's short, but it really refers to the number of weights in the mask. For errors, I think it's a little clearer if we're more explicit.

rust/xaynet-core/src/message/payload/update.rs
little-dude: done

rust/xaynet-core/src/mask/object/serialization.rs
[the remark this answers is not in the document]
finiteprods: ah true. then I suggest
```suggestion
            .context("failed to parse the number of items in mask object")?;
```

rust/xaynet-core/src/message/payload/update.rs
finiteprods: thanks, why the thumbs down?

rust/xaynet-core/src/message/traits.rs
```diff
@@ -301,6 +313,43 @@ impl FromBytes for LocalSeedDict {
...
+                "expected {} bytes, but only {} left",
+                len - 4,
+                iter.len()
+            ));
+        }
+
+        let mut dict = LocalSeedDict::new();
+        let entries = iter.take(len - 4).chunks(ENTRY_LENGTH);
+        for mut chunk in entries.into_iter() {
+            let key = SumParticipantPublicKey::from_byte_stream(&mut chunk)
```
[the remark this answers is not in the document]
little-dude: I don't know, I think `by_ref` "raison d'etre" is method chaining:

```rust
sutff.
    .into_iter()
    .by_ref()
    .foo()
    .bar()
```

is better than

```rust
(&mut stuff.into_iter()).foo().bar()
```

In this specific case though we're not doing that, and I'd argue that `by_ref` is actually not as explicit as `&mut`. In the standard library, `ref` often refers to shared references (eg `Deref::deref`, `AsRef::as_ref`). Here that convention is not followed, and I actually opened the doc to double check that the return type was a mutable reference

rust/xaynet-core/src/message/payload/update.rs
little-dude: Sorry, that was a misclick, I wanted to thumb this up actually :smile:

rust/xaynet-core/src/mask/object/serialization.rs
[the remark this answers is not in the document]
little-dude: oops. Yes thank you

rust/xaynet-core/src/message/traits.rs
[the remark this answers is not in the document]
janpetschexain: i agree that `by_mut_ref()` would be way better. your chaining point is also valid, in the end it's up to you to choose the format.

rust/xaynet-core/src/message/traits.rs
```diff
@@ -376,3 +425,61 @@ mod tests {
         assert_eq!(bytes, expected);
     }
 }
+
+impl FromBytes for u16 {
```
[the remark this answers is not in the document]
little-dude: done

rust/xaynet-core/src/message/mod.rs
little-dude: I'm not too concerned about test duplication, but the problem is if we make changes to the iterator, we may forget to update the docstring and not notice that it's broken.

little-dude: I did as you suggested in the end.

janpetschexain (approved):

finiteprods (approved):
the implementation of chunkable iterators looks very clean and nicely documented to me.

perhaps you could just comment on the significance of using `ExactSizeIterator`s, which wasn't clear to me (I think one benefit is it helps with parsing e.g. we can abort as soon as `len()` is not what we were expecting)?

also, do you foresee the requirement of the implementation that chunks be consumed strictly in the order they're produced, to be a limitation anytime soon?

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
```diff
@@ -0,0 +1,478 @@
...
+        let _c2 = chunks_iter.next().unwrap();
+        let mut c3 = chunks_iter.next().unwrap();
+
+        assert_eq!(c3.next().unwrap(), 4);
+    }
+
+    // This test case illustrates a weird behavior of our iterator:
+    // everything being lazy, we can create chunks that start *beyond*
+    // what our main iterator can provide in theory. Attempting to
+    // consume such iterators should panic
```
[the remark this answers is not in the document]
little-dude: Yeah it's quirky so I though it was worth documenting as a test.

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
```diff
@@ -0,0 +1,478 @@
...
+    /// returned. If the requested `index` hasn't been reached yet,
+    /// this method panics. This is to enforce the invariant that all
+    /// chunks must be consumed in order.
+    ///
+    /// # Examples
+    ///
+    /// ```should_panic
+    /// # use xaynet_core::message::utils::IntoChunks;
+    /// let iter = vec![0, 1, 2, 3, 4, 5].into_iter();
+    /// let chunk_size = 2;
```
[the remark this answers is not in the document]
little-dude: Indeed.

rust/xaynet-core/src/message/utils/chunkable_iterator.rs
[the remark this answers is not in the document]
little-dude: A lot of the quirkiness around this type would be solved when rust provides generic associated types. The motivation given in the RFC for that feature is exactly what we want: chunks that are bound to the lifetime of `&self` in `Iterator::next`:


```rust
trait StreamingIterator {
    type Item<'a>;
    fn next<'a>(&'a mut self) -> Option<Self::Item<'a>>;
}
```

As a type theory nerd you might enjoy reading the RFC. It's one of the most awaited features: https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md

PR #531, opened 2020-09-25
Add comment about the redis url configuration in k8s

Robert-Steiner:


rsaffi (changes requested):
Can you please reference that `XAYNET_REDIS__URL` is defined as a Kubernetes secret and exposed to the pod as such? Just to make it cristal clear...

rsaffi (approved):
LGTM

PR #532, opened 2020-09-28
Fix message serialization

little-dude:
@Robert-Steiner noticed that the test-drive example was broken, because the messages were rejected by the message parser service in the update phase. The reason is that the tag field in the update and sum2 messages was set to `1` which is for sum messages. The first commit fixes this. The second commit make `Message::to_bytes` a little more typesafe by relying the `Message.payload` field instead of `Message.tag` when possible.

### DIFF
--- a/rust/xaynet-core/src/message/message.rs
+++ b/rust/xaynet-core/src/message/message.rs
@@ -516,7 +516,7 @@ impl Message {
             participant_pk,
             coordinator_pk,
             is_multipart: false,
-            tag: Tag::Sum,
+            tag: Tag::Sum2,
             payload: message.into(),
         }
     }
@@ -533,7 +533,7 @@ impl Message {
             participant_pk,
             coordinator_pk,
             is_multipart: false,
-            tag: Tag::Sum,
+            tag: Tag::Update,
             payload: message.into(),
         }
     }
--- a/rust/xaynet-core/src/message/message.rs
+++ b/rust/xaynet-core/src/message/message.rs
@@ -478,9 +478,11 @@ pub struct Message {
     pub coordinator_pk: PublicEncryptKey,
     /// Wether this is a multipart message
     pub is_multipart: bool,
-    /// The type of message. This information is only partially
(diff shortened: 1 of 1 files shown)


rust/xaynet-core/src/message/message.rs
```diff
@@ -478,9 +478,12 @@ pub struct Message {
...
     /// Wether this is a multipart message
     pub is_multipart: bool,
-    /// The type of message. This information is only partially
-    /// carried by the `payload` field, because in the case of a
-    /// multipart message, the payload is of type `Payload::Chunk`.
+    /// The type of message. This information is already partially
+    /// partially redundand with the `payload` field. So when
+    /// serializing the message, this field is ignored if the payload
+    /// is a [`Payload::Sum`], [`Payload::Update`], or
+    /// [`Payload::Sum2`]. However, it is taken as is for for
```
Robert-Steiner, line 485:
```suggestion
    /// [`Payload::Sum2`]. However, it is taken as is for
```

rust/xaynet-core/src/message/message.rs
```diff
@@ -478,9 +478,12 @@ pub struct Message {
     pub coordinator_pk: PublicEncryptKey,
     /// Wether this is a multipart message
     pub is_multipart: bool,
-    /// The type of message. This information is only partially
-    /// carried by the `payload` field, because in the case of a
-    /// multipart message, the payload is of type `Payload::Chunk`.
+    /// The type of message. This information is already partially
```
Robert-Steiner, line 481:
```suggestion
    /// The type of message. This information is already
```

Robert-Steiner (approved):
LGTM!

rust/xaynet-core/src/message/message.rs
little-dude: Actually _partially_ is important here. When the payload is of type `Payload::Chunk` we don't know the actual type of the message

little-dude: Ah but I've put partially twice, I see.

PR #534, opened 2020-09-29
XN-1146 Setup SSL certificate data in Kubernetes

rsaffi:
This mounts the letsencrypt certificate (key and crt) into `/app/ssl/tls.key` and `/app/ssl/tls.crt`, respectively.
Whenever the letsencrypt certificates are renewed, the files inside the container are automatically updated.

### DIFF
--- a/k8s/coordinator/development/cert-volume-mount.yaml
+++ b/k8s/coordinator/development/cert-volume-mount.yaml
@@ -1,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: coordinator-deployment
+spec:
+  template:
+    spec:
+      volumes:
+        - name: tls-certificate
+          secret:
+            secretName: dev-coordinator
+            items:
+              - key: tls.key
+                path: tls.key
+                mode: 0600
+              - key: tls.crt
+                path: tls.crt
+                mode: 0644
+      containers:
+        - name: coordinator
+          volumeMounts:
+            - name: tls-certificate 
+              mountPath: "/app/ssl"
+              readOnly: true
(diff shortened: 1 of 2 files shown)


janpetschexain (approved):

k8s/coordinator/development/cert-volume-mount.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      volumes:
+        - name: tls-certificate
+          secret:
+            secretName: dev-coordinator
+            items:
+              - key: tls.key
+                path: tls.key
+                mode: 0400
+              - key: tls.crt
```
Robert-Steiner, line 16:
```suggestion
              - key: tls.pem
```

k8s/coordinator/development/cert-volume-mount.yaml
```diff
@@ -0,0 +1,24 @@
...
+                path: tls.pem
+                mode: 0444
```
Robert-Steiner, line 18: Why not `400`?

k8s/coordinator/development/cert-volume-mount.yaml
rsaffi: Actually the *key* is stored by our letsencrypt issuer as `tls.crt` inside the secrets, so we can't change the key to `tls.pem`, otherwise it'll reference a non-existing key. 

What can be done (and it's what I did) is to mount that `tls.crt` key to a `tls.pem` file. :)

Robert-Steiner: Ahh I see 👍

k8s/coordinator/development/cert-volume-mount.yaml
rsaffi: The certificate file itself can be world readable without any problems. I just followed what's common practice by having it be `0444` (it's actually usually `0644`, but since I already mount the secret as a read-only file, I went the extra mile and made it be `0444` to reflect that).

Robert-Steiner (approved):

PR #535, opened 2020-09-29
Refactor add sum participant result

Robert-Steiner:
A small PR that aligns the sum participant result with the `SeedDictUpdate` result. 

Next sprint, I would like to rethink the error handling in the state machine and maybe change it because the error handling slowly becomes a bit confusing.

### DIFF
--- a/rust/xaynet-server/src/storage/redis.rs
+++ b/rust/xaynet-server/src/storage/redis.rs
@@ -32,7 +32,6 @@
 use crate::{
     state_machine::coordinator::CoordinatorState,
     storage::impls::{
-        AddSumParticipant,
         EncryptedMaskSeedRead,
         LocalSeedDictWrite,
         MaskObjectRead,
@@ -42,6 +41,7 @@ use crate::{
         PublicSigningKeyRead,
         PublicSigningKeyWrite,
         SeedDictUpdate,
+        SumDictAdd,
     },
 };
 use redis::{aio::ConnectionManager, AsyncCommands, IntoConnectionInfo, RedisResult, Script};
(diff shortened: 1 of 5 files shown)


rust/xaynet-server/src/storage/redis.rs
```diff
@@ -630,7 +630,10 @@ mod tests {
             .add_sum_participant(pk, epk)
             .await
             .unwrap();
-        assert_eq!(key_already_exist, AddSumParticipant::AlreadyExists);
+        assert!(matches!(
+            key_already_exist.into_inner().unwrap_err(),
+            SumDictAddError::AlreadyExists
+        ));
```
janpetschexain, line 636: i was wondering what's the reason for `matches!`? wouldn't
```rust
assert_eq!(
    key_already_exist.into_inner().unwrap_err(),
    SumDictAddError::AlreadyExists,
);
```
result in the same outcome?

janpetschexain (approved):
lgtm

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: That's right, but I would have to add `#[derive(PartialEq)]` to the `SeedDictUpdateError` enum. With `matches!` it works without `PartialEq`.

janpetschexain: i see, fine for me to do it this way. another way would be to annotate the struct with `#[cfg_attr(test, derive(PartialEq))]`, then it will only derive if the code is compiled for tests.

little-dude (approved):
:+1:

PR #538, opened 2020-09-30
Redis mask dict preparation

Robert-Steiner:
**Summary**
- prepares the Redis client for storing mask dict 
- added lua script for increment mask count
- added PET errors for the increment mask count operation
- fixed/added tests

### DIFF
--- a/rust/xaynet-server/src/storage/impls.rs
+++ b/rust/xaynet-server/src/storage/impls.rs
@@ -5,7 +5,7 @@ use redis::{ErrorKind, FromRedisValue, RedisError, RedisResult, RedisWrite, ToRe
 use thiserror::Error;
 use xaynet_core::{
     crypto::{ByteObject, PublicEncryptKey, PublicSigningKey},
-    mask::{EncryptedMaskSeed, MaskMany},
+    mask::{EncryptedMaskSeed, MaskObject},
     LocalSeedDict,
 };
 
@@ -163,12 +163,12 @@ macro_rules! impl_bincode_redis_traits {
 impl_bincode_redis_traits!(CoordinatorState);
 
 #[derive(From, Into, Serialize, Deserialize)]
-pub(crate) struct MaskObjectRead(MaskMany);
+pub(crate) struct MaskObjectRead(MaskObject);
 
 impl_bincode_redis_traits!(MaskObjectRead);
 
 #[derive(From, Serialize)]
-pub(crate) struct MaskObjectWrite<'a>(&'a MaskMany);
+pub(crate) struct MaskObjectWrite<'a>(&'a MaskObject);
 
 impl ToRedisArgs for MaskObjectWrite<'_> {
     fn write_redis_args<W>(&self, out: &mut W)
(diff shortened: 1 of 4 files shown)


janpetschexain:
there seems to be some overlappings with #537 , should we wait until that is merged and rebased before we review this one?

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -343,25 +349,51 @@ impl Connection {
...
-        //   represented as string.
-        //
-        // We ignore the return value because we are not interested in it. We will use the method
-        // `get_best_masks` instead.
+        let script = Script::new(
+            r#"
+                -- lua lists (tables) start at 1
+                local sum_pk = ARGV[1]
+
+                -- check if client participated in sum phase and
```
little-dude, line 363: seems like this comment is truncated.

rust/xaynet-server/src/storage/impls.rs
```diff
@@ -5,7 +5,7 @@ use redis::{ErrorKind, FromRedisValue, RedisError, RedisResult, RedisWrite, ToRe
 use thiserror::Error;
 use xaynet_core::{
     crypto::{ByteObject, PublicEncryptKey, PublicSigningKey},
-    mask::{EncryptedMaskSeed, MaskMany},
+    mask::{EncryptedMaskSeed, MaskObject},
```
little-dude, line 8: Looks like a rebasing issue?

little-dude (approved):
60491ba looks good to me :100: 
As @janpetschexain said the timeout changes have been pulled in though. Maybe we could just cherry-pick 60491ba while we take a decision on those?

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: Oh yes, good catch! I will fix it

rust/xaynet-server/src/storage/impls.rs
Robert-Steiner: rebasing issue because it was previously `MaskMany` or because is now `MaskObject`?

rust/xaynet-server/src/storage/impls.rs
```diff
@@ -334,3 +334,34 @@ pub enum SumDictDeleteError {
...
+        self.0
+    }
+}
+
+impl FromRedisValue for MaskDictIncr {
+    fn from_redis_value(v: &Value) -> RedisResult<MaskDictIncr> {
+        match *v {
+            Value::Int(-2) => Ok(MaskDictIncr(Err(MaskDictIncrError::MaskAlreadySubmitted))),
+            Value::Int(-1) => Ok(MaskDictIncr(Err(MaskDictIncrError::UnknownSumPk))),
+            Value::Int(0) => Ok(MaskDictIncr(Ok(()))),
```
finiteprods, line 352: thought that occurred to me: it would be nice to be able to map "error codes" to error variants right at the point of definition, by making the error codes discriminants of the enum variants i.e.
```rust
pub enum MaskDictIncrError {
    UnknownSumPk = -1,
    MaskAlreadySubmitted = -2,
}
```
but it seems it's not straightforward to convert from an int to an enum, there's at least [some boilerplate](https://stackoverflow.com/a/57578431) required. so I'm not sure whether it's worth it. but might be worth considering.

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -292,22 +298,48 @@ impl Connection {
...
+                -- check if the client participated in sum phase
+                --
+                -- Note: we cannot delete the sum_pk in the sum_dict because we
+                -- need the sum_dict later to delete the seed_dict
+                local sum_pk_exist = redis.call("HEXISTS", "sum_dict", sum_pk)
+                if sum_pk_exist == 0 then
+                    return -1
+                end
+
+                -- check if sum participated has not already submitted a mask
```
finiteprods, line 321: do you mean "sum participant"?
while we're here, also check the comment on line 249:
```
-- This condition and should never apply.
```

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -292,22 +298,48 @@ impl Connection {
...
+        );
+
+        script
+            .key(MaskObjectWrite::from(mask))
+            .arg(PublicSigningKeyWrite::from(sum_pk))
+            .invoke_async(&mut self.connection)
             .await
     }
 
     /// Retrieves the two masks with the highest score.
```
finiteprods, line 341: why *two* masks?

finiteprods (commented):
looks good. just a few minor comments.

rust/xaynet-server/src/storage/impls.rs
Robert-Steiner: Not a bad idea🤔 Maybe we can use `#[derive(TryInto)]` to hide the boilerplate code. I will try it out and give an update on the result.

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: In the state unmask, we check whether the two masks with the highest score do not have the same score.

rust/xaynet-server/src/storage/redis.rs
Robert-Steiner: fixed

PR #539, opened 2020-09-30
Storing mask dict in redis

Robert-Steiner:
**Summary**
- storing the mask dict in redis
- fix tests
- extended Redis Connection struct with `get_mask_submitted_set` method to be abe to emit the metric `metrics::masks::total_number`


rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -39,16 +32,31 @@ impl Phase for PhaseState<Unmask> {
...
-                self.shared.state.round_id,
-                Self::NAME
-            )
-        );
+        #[cfg(feature = "metrics")]
+        {
+            let redis = &mut self.shared.io.redis;
+            let metrics_tx = &mut self.shared.io.metrics_tx;
+
+            match redis.connection().await.get_number_of_unique_masks().await {
```
little-dude, line 40: Could this be `spawned`? I don't think we need to really wait for it.

rust/xaynet-server/src/state_machine/phases/sum2.rs
```diff
@@ -120,47 +107,38 @@ impl PhaseState<Sum2> {
...
-        if let Some(count) = self.inner.mask_dict.get_mut(&mask) {
-            *count += 1;
-        } else {
-            self.inner.mask_dict.insert(mask, 1);
-        }
+        self.shared
+            .io
+            .redis
+            .connection()
+            .await
```
little-dude, line 130: I know this is not new from this PR but it didn't strike me before: why do we need to `await` for the connection? Don't we have a pool of pre-established connections?

little-dude (approved):
Looks good to me :+1: I just left a question of a suggestion regarding the metrics. Nice to see the test coverage increasing too.

rust/xaynet-server/src/state_machine/phases/unmask.rs
Robert-Steiner: Good idea👍

rust/xaynet-server/src/state_machine/phases/sum2.rs
Robert-Steiner: The `await` comes form the `Semaphore` that limits the number of concurrent requests. However, the multiplex connection is only a [single connection](https://github.com/mitsuhiko/redis-rs/issues/218#issuecomment-509216100).
For connection pool we would need something like [`deadpool-redis`](https://crates.io/crates/deadpool-redis) or [`bb8-redis`](https://crates.io/crates/bb8-redis/0.3.1) ([ref of the comment](https://github.com/mitsuhiko/redis-rs/pull/388#issuecomment-702239648))

little-dude: I see. All good then. I was just afraid we were re-establishing a new connection every time.

PR #540, opened 2020-10-01
message splitting logic

little-dude:
I decided to split my PR into smaller chunks. This is the module that implements the logic for splitting messages in chunks.

### DIFF
--- a/rust/xaynet-client/src/lib.rs
+++ b/rust/xaynet-client/src/lib.rs
@@ -53,12 +53,11 @@ use tokio::time;
 
 use xaynet_core::{crypto::ByteObject, mask::Model, CoordinatorPublicKey, InitError};
 
+pub mod api;
 #[doc(hidden)]
 pub mod mobile_client;
-
-pub mod api;
-
 mod participant;
+pub(crate) mod utils;
 pub use participant::{Participant, Task};
 
 #[derive(Clone, Debug)]
--- a/rust/xaynet-client/src/utils/mod.rs
+++ b/rust/xaynet-client/src/utils/mod.rs
@@ -1,0 +1,1 @@
+mod multipart;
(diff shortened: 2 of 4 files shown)


rust/xaynet-client/src/multipart.rs
```diff
@@ -0,0 +1,130 @@
...
+        let end = cmp::min(start + self.max_chunk_size, self.data.as_ref().len());
+        let range = start..end;
+        &self.data.as_ref()[range]
+    }
+}
+
+/// A helper that performs division with ceil.
+fn ceiling_div(n: usize, d: usize) -> usize {
+    (n + d - 1) / d
+}
```
janpetschexain, line 54: this will panic if `d == 0`. it's never used like that in above's code and this is a private function, but some extra docs never hurt.

janpetschexain (approved):

rust/xaynet-client/src/multipart.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Get the total number of chunks
+    pub fn nb_chunks(&self) -> usize {
+        let data_len = self.data.as_ref().len();
+        ceiling_div(data_len, self.max_chunk_size)
+    }
+
+    /// Get the chunk with the given ID.
+    pub fn get_chunk(&self, id: usize) -> &'a [u8] {
+        if id >= self.nb_chunks() {
+            panic!("no chunk with ID {}", id);
```
Robert-Steiner, line 42: Is there a reason why you decided to panic here instead of returning an `Option`?

Robert-Steiner (approved):
LGTM👍

rust/xaynet-client/src/multipart.rs
little-dude: Done

rust/xaynet-client/src/multipart.rs
little-dude: Simplicity. This is for internal use only, and we never call this with an invalid `id`. If we don't panic here, we'll probably just unwrap() at the caller site anyway. That also spares us for adding another error type.

little-dude: I added a panic section to the docstring though

PR #541, opened 2020-10-01
add a helper for running futures concurrently

little-dude:
This will allow clients to start sending multiple chunks concurrently.

[`FuturesUnordered`](https://docs.rs/futures-util/0.3.5/futures_util/stream/futures_unordered/struct.FuturesUnordered.html)
allows us to poll multiple futures, but if we want to really run them
_concurrently_ we need to spawn them first. So this type is just a thin
wrapper around `FuturesUnordered`.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -974,9 +974,9 @@ checksum = "3dcaa9ae7725d12cdb85b3ad99a434db70b468c09ded17e012d86b5c1010f7a7"
 
 [[package]]
 name = "futures"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "90fa4cc29d25b0687b8570b0da86eac698dcb525110ad8b938fe6712baa711ec"
+checksum = "da9052a1a50244d8d5aa9bf55cbc2fb6f357c86cc52e46c62ed390a7180cf150"
 dependencies = [
  "futures-channel",
  "futures-core",
@@ -989,9 +989,9 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "31ebc390c6913de330e418add60e1a7e5af4cb5ec600d19111b339cafcdcc027"
+checksum = "f2d31b7ec7efab6eefc7c57233bb10b847986139d88cc2f5a02a1ae6871a1846"
 dependencies = [
  "futures-core",
  "futures-sink",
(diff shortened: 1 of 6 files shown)


rust/xaynet-client/src/utils/concurrent_futures.rs
```diff
@@ -0,0 +1,131 @@
...
+mod tests {
+    use std::time::Duration;
+
+    use futures::stream::StreamExt;
+    use tokio::time::delay_for;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test() {
```
janpetschexain, line 86: just out of interest: can the timings of pushing futures always be guaranteed to be precise down to the millisecond? the timings in this test are quite tight, if pushing new futures after another has resolved is delayed for only 1ms (for example for test environments with heavy load), then this test will likely yield a different order of values.

janpetschexain (approved):

rust/xaynet-client/src/utils/concurrent_futures.rs
little-dude: The interval between the futures is ~10ms, not 1ms, so I think it's enough for tests.

janpetschexain: the way i understood the timeline is the following: in the beginning all 4 futures are queued and none is spawned, then:
- 0ms: futures 1 and 2 are spawned, futures 3 and 4 are queued
- 10ms: future 1 resolves, future 2 is still pending, future 3 is spawned, future 4 is still queued
- 22ms: future 3 is resolved, future 2 is still pending, future 4 is spawned, none is queued any more
- 23ms: future 4 is resolved, future 2 is still pending
- 25ms: future 2 is resolved

this leaves only a two milliseconds between resolving the last two futures. if anything is delayed in between, this might be quite tight.

little-dude: Oh yes you're right. I should probably rework this a bit indeed.

janpetschexain:
rebased. we should probably move this to the `e2e` package later on (#568).

rust/xaynet-client/src/utils/concurrent_futures.rs
janpetschexain: fixed.

Robert-Steiner (approved):
@janpetschexain thanks for the rebase 👍

little-dude:
What are you guys going to use this for? I forgot why I did that in the first place

janpetschexain:
> What are you guys going to use this for? I forgot why I did that in the first place

@Robert-Steiner wants to use it for the e2e tests, but they are not mergeable yet. since we are cleaning up things for the next release, we want to merge this so it doesn't get lost.

little-dude:
Forgot to mention why I didn't want to use `FuturesUnordered` or `BufferUnordered`: https://discord.com/channels/500028886025895936/500336333500448798/704402952235909200

> @little-dude if you want futures to execute **concurrently** you may want to spawn anyway... you can use Semaphore to limit the # of in-flight.
> `FuturesUnordered` & keep all futures on the same thread

(Commenting because I'm having a similar use case again and I was wondering whether I should use `FuturesUnordered` ^^)

PR #542, opened 2020-10-01
fix extended coverage of masking tests

finiteprods:
#524 introduced various changes to tidy the API especially around the generalised scalar extension. As such, the masking tests (several hundred tests generated by the macros `test_masking`, `test_aggregation` and `test_masking_and_aggregation`) now cover the (un)masking and aggregation of the scalar as well.

An issue was discovered in #524 where the unmasked scalar sum appeared larger than expected. This lead to "closeness checks" of model weights to exceed the desired tolerance, and so those tests (at least, those based on `test_masking` and `test_masking_and_aggregation`) were temporarily relaxed. This PR contains a fix for that issue, which was due to an inconsistency in the use of the PRNG in masking vs unmasking.

All masking tests are now reinstated, although a related issue has been found with a few test cases of `test_masking_and_aggregation`. Again, closeness checks break for those cases - for now, a slightly fewer number of models are aggregated to keep the difference within tolerance.

On a separate note, this PR also contains a more complete implementation of `MaskObjectBuffer`.

### DIFF
--- a/rust/xaynet-core/src/mask/masking.rs
+++ b/rust/xaynet-core/src/mask/masking.rs
@@ -11,6 +11,7 @@ use num::{
     clamp,
     rational::Ratio,
 };
+
 use rand::SeedableRng;
 use rand_chacha::ChaCha20Rng;
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


janpetschexain (approved):
apart from XN-1180 this is good to go, but this shouldn't prevent merging this pr.

two things that might either be added here or in a follow up ticket:
- the naming should be consistified, currently `many`/`vector`/`n` and `one`/`scalar`/`1` are used interchangeably in various places
- the unmasking can be optimized memory wise, currently we do to `collect` allocations, but if the order of operations is changed to "unmask scalar -> unmask model and apply correction directly in the map -> collect" we can save a whole vector allocation for the model

finiteprods:
Thanks for reviewing @janpetschexain and pointing out the memory optimization. Since there is already a plan to deal with more consistent naming in a PR soon, let's also roll the memory optimization into there.

PR #543, opened 2020-10-02
Implement `MessageEncoder`

little-dude:
The two commits can and should probably be reviewed separately.

The `MessageEncoder` is an iterator that takes a message payload of   
type `Payload::{Sum,Update,Sum2}`, and yields serialized and signed   
messages.                                                             
                                                                      
Given a maximum payload length, the encoder automatically chunks the  
messages that are too big. For payloads that are small enough to fix  
in a single message, the iterator only yields one item. For payload   
that needs to be split in several messages, the encoder yields several
items.                                                                
                                                                      
Thus, it offers a common serialization interface for both multipart   
and non-multipart messages.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2861,6 +2861,7 @@ dependencies = [
  "bincode",
  "bytes",
  "derive_more",
+ "num",
  "reqwest",
  "serde 1.0.116",
  "sodiumoxide",
--- a/rust/xaynet-client/Cargo.toml
+++ b/rust/xaynet-client/Cargo.toml
@@ -25,6 +25,7 @@ reqwest = { version = "0.10.8", default-features = false }
 [dev-dependencies]
 tower-test = "0.3.0"
 tokio-test = "0.2.1"
+num = { version = "0.3.0", features = ["serde"] }
 
 [features]
 default = []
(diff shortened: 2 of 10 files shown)


rust/xaynet-client/src/mobile_client/client.rs
```diff
@@ -46,6 +57,27 @@ impl<Type> ClientState<Type> {
...
+        // Unwrapping is fine because this only errors out if the
+        // payload if a Chunk, which we never create in the client.
+        let encoder = MessageEncoder::<'_, Type>::new(
+            &self.participant,
+            payload,
+            self.round_params.pk,
+            MAX_PAYLOAD_SIZE,
+        )
+        .unwrap();
+        for part in encoder {
```
Robert-Steiner, line 75: Just for my understanding.
In the event that the client is terminated in the middle of the loop, the client would send the messages that are already received by the coordinator in the next attempt again + the missing ones? Can the coordinator already handle this case?

Robert-Steiner (approved):
This PR and #552 have the same changes it that intended?

I really like the changes especially how you reduce the duplicate code in the client👍

little-dude:
> This PR and #552 have the same changes it that intended?

Oops no. The changes to the mobile client are not supposed to be in here. This PR is just for the `MessageEncoder`. I'll remove the extra commit and merge, then rebase #552

rust/xaynet-client/src/mobile_client/client.rs
little-dude: >  the client would send the messages that are already received by the coordinator in the next attempt again + the missing ones?

Yes.

>  Can the coordinator already handle this 

Yes, it'll just ignore the parts if already has.

PR #544, opened 2020-10-02
S3 client implementation

Robert-Steiner:
**Summary**
- added s3 client
- added tests

As I already mentioned it would be nice to have an API that is similar to the redis client. For example, in the s3 client we don't have a limit on concurrent requests because currently we only write the global model at the end of the round.
However, if we also want to persist the local models, we will definitely need that.
In addition, it would be nice to define traits that are implemented by the clients so that we use the traits in the state machine instead of the concrete types. That would give us the ability to for example easily swap out the Redis client for in-memory storage (as we had before).

As with the redis client, the s3 client also has no functionality for retries and timeouts because we decided to solve it later. In my opinion, we should investigate in a poc whether we can possibly also use `tower` in this case instead of implementing these functions ourselves. 
As you can see, there is a lot of room for improvement, but I kept it simple for now because we still have other important topics to solve. I hope that is fine.

### DIFF
--- a/rust/xaynet-server/src/storage/s3.rs
+++ b/rust/xaynet-server/src/storage/s3.rs
@@ -1,0 +1,290 @@
+use crate::settings::{S3BucketsSettings, S3Settings};
+use rusoto_core::{credential::StaticProvider, request::TlsError, HttpClient, RusotoError};
+use rusoto_s3::{
+    CreateBucketError,
+    CreateBucketOutput,
+    CreateBucketRequest,
+    DeleteObjectsError,
+    ListObjectsV2Error,
+    PutObjectError,
+    PutObjectOutput,
+    PutObjectRequest,
+    S3Client,
+    StreamingBody,
+    S3,
+};
+use std::sync::Arc;
+use thiserror::Error;
+
+use xaynet_core::mask::Model;
+
+type S3Result<T> = Result<T, S3Error>;
+
+#[derive(Debug, Error)]
+pub enum S3Error {
+    #[error("upload error: {0}")]
+    Upload(#[from] RusotoError<PutObjectError>),
+    #[error("create bucket error: {0}")]
+    CreateBucket(#[from] RusotoError<CreateBucketError>),
+    #[error("list objects error: {0}")]
+    ListObjects(#[from] RusotoError<ListObjectsV2Error>),
+    #[error("delete objects error: {0}")]
+    DeleteObjects(#[from] RusotoError<DeleteObjectsError>),
+    #[error("serialization failed")]
+    Serialization(#[from] bincode::Error),
+    #[error("empty response error")]
+    EmptyResponse,
+    #[error(transparent)]
+    HttpClient(#[from] TlsError),
+}
+
+#[derive(Clone)]
+pub struct Client {
+    buckets: Arc<S3BucketsSettings>,
+    s3_client: S3Client,
+}
+
+impl Client {
+    /// Creates a new S3 client. The client creates and maintains one bucket for storing global models.
+    ///
+    /// To connect to AWS-compatible services such as Minio, you need to specify a custom region.
+    /// ```
+    /// use rusoto_core::Region;
+    /// use xaynet_server::{
+    ///     settings::{S3BucketsSettings, S3Settings},
+    ///     storage::s3::Client,
+    /// };
+    ///
+    /// let region = Region::Custom {
+    ///     name: String::from("minio"),
+    ///     endpoint: String::from("http://127.0.0.1:9000"), // URL of minio
+    /// };
+    ///
+    /// let s3_settings = S3Settings {
+    ///     region,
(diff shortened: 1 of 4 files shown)


rust/xaynet-server/src/storage/s3.rs
```diff
@@ -0,0 +1,289 @@
...
+    Delete,
+    DeleteObjectsOutput,
+    DeleteObjectsRequest,
+    ListObjectsV2Output,
+    ListObjectsV2Request,
+    ObjectIdentifier,
+};
+
+#[cfg(test)]
+impl Client {
```
janpetschexain, line 147: this impl could be moved into the `tests` module below. but if it is also needed for other tests elsewhere it probably  should be moved to some test helper module.

rust/xaynet-server/src/storage/tests/mod.rs
```diff
@@ -132,3 +142,7 @@ pub async fn write_local_seed_entries(
 
     update_result
 }
+
+pub fn create_global_model(byte_size: usize) -> Model {
```
janpetschexain, line 146: `byte_size` is quite misleading, since the `Model::from_primitives()` will cast the `Vec<i32/64>` into a `Vec<Ratio<BigInt>>` where the byte size depends on the chosen mask config. maybe `len` or `nb_elements` is a better name.

little-dude (approved):
:+1:

rust/xaynet-server/src/storage/s3.rs
```diff
@@ -0,0 +1,289 @@
...
+        Ok(Self {
+            buckets: Arc::new(settings.buckets),
+            s3_client: S3Client::new_with(dispatcher, credentials_provider, settings.region),
+        })
+    }
+
+    /// Uploads a global model.
+    pub async fn upload_global_model(&self, key: &str, global_model: &Model) -> S3Result<()> {
+        // As key for the global model we use the seed of the round in which the global model was
+        // created. We also might store the round number for easier debugging?
```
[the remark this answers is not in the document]
Robert-Steiner: That is a good idea👍 I will change it

rust/xaynet-server/src/storage/s3.rs
Robert-Steiner: I agree but I would like to keep it here to make adding the feature flag easier in #567 .
this way I only need to add the flag to `mod  s3` in `rust/xaynet-server/src/storage/mod.rs`

janpetschexain: i'm not sure if i get this, because adding the flag to the `s3` module will also enable/disable the submodules which includes `s3::tests`. but if it's easier this way, then i'm ok with that.

Robert-Steiner: ah sry my bad, I thought the idea was to add this to `rust/xaynet-server/src/storage/tests/mod.rs`
That make sense I can do that👍

rust/xaynet-server/src/storage/s3.rs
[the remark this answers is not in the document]
Robert-Steiner: Updated the comment + I'll move this comment in #567 to the place where we save the model

janpetschexain (approved):

rust/xaynet-server/src/storage/s3.rs
```diff
@@ -0,0 +1,290 @@
...
+                    objects: identifiers,
+                    ..Default::default()
+                },
+                ..Default::default()
+            };
+
+            self.s3_client.delete_objects(req).await.map_err(From::from)
+        }
+
+        // Returns all objects keys for the given bucket.
```
finiteprods, line 216:
```suggestion
        // Returns all object keys for the given bucket.
```

rust/xaynet-server/src/storage/s3.rs
```diff
@@ -0,0 +1,290 @@
...
+        Ok(Self {
+            buckets: Arc::new(settings.buckets),
+            s3_client: S3Client::new_with(dispatcher, credentials_provider, settings.region),
+        })
+    }
+
+    /// Uploads a global model.
+    pub async fn upload_global_model(&self, key: &str, global_model: &Model) -> S3Result<()> {
+        // As key for the global model we use the round_id and the seed of the round in which the
+        // global model was created.
```
finiteprods, line 88: should this be part of the docstring? since the caller wouldn't otherwise know this (or is this just for the test? In which case maybe this comment belongs there...).

finiteprods (approved):
minor comments you needn't address, otherwise looks very nice.

rust/xaynet-server/src/storage/s3.rs
Robert-Steiner: you are right, I will move this line to the place where we store the model (in the unmasking phase) in #567 
After the restoring story I will add some module documentation to the s3/redis client.

PR #546, opened 2020-10-06
S3 settings

Robert-Steiner:
**Summary**
- added S3 settings

### DIFF
--- a/k8s/coordinator/development/config.toml
+++ b/k8s/coordinator/development/config.toml
@@ -34,3 +34,9 @@ db = "metrics"
 # `XAYNET_REDIS__URL` depends on the enviroment variable `REDIS_AUTH`,
 # which is defined as a Kubernetes secret and exposed to the coordinator pod.
 # See: k8s/coordinator/base/deployment.yaml
+
+[s3]
+# The access_key and secret_access_key are configured via the enviroment variables
+# `XAYNET_S3__ACCESS_KEY` and `XAYNET_S3__SECRET_ACCESS_KEY`.
+# See: k8s/coordinator/base/deployment.yaml
+region = ["minio", "http://minio:9000"]
--- a/rust/xaynet-server/src/settings.rs
+++ b/rust/xaynet-server/src/settings.rs
@@ -6,8 +6,13 @@
 use std::{fmt, path::PathBuf};
 
 use config::{Config, ConfigError, Environment};
+use fancy_regex::Regex;
 use redis::{ConnectionInfo, IntoConnectionInfo};
-use serde::de::{self, Deserializer, Visitor};
+use rusoto_core::Region;
+use serde::{
+    de::{self, value, Deserializer, Visitor},
+    Deserialize,
+};
 use thiserror::Error;
 use tracing_subscriber::filter::EnvFilter;
 use validator::{Validate, ValidationError, ValidationErrors};
@@ -28,7 +33,6 @@ pub enum SettingsError {
 ///
 /// Each section in the configuration file corresponds to the identically named settings field.
 pub struct Settings {
-    #[validate]
     pub api: ApiSettings,
     #[validate]
     pub pet: PetSettings,
@@ -38,6 +42,8 @@ pub struct Settings {
     #[validate]
     pub metrics: MetricsSettings,
     pub redis: RedisSettings,
+    #[validate]
+    pub s3: S3Settings,
 }
 
 impl Settings {
@@ -281,7 +287,7 @@ fn validate_fractions(s: &PetSettings) -> Result<(), ValidationError> {
     }
 }
 
-#[derive(Debug, Validate, Deserialize, Clone)]
+#[derive(Debug, Deserialize, Clone)]
 /// REST API settings.
 pub struct ApiSettings {
     /// The address to which the REST API should be bound.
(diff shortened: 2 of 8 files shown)


Robert-Steiner:
#build

Robert-Steiner:
#build

Robert-Steiner:
#build

rsaffi:
I seem to have missed this, sorry! Is this already ready-for-review or still wip?

Robert-Steiner:
It is ready 🙂

k8s/coordinator/development/config.toml
```diff
@@ -34,3 +34,9 @@ db = "metrics"
 # `XAYNET_REDIS__URL` depends on the enviroment variable `REDIS_AUTH`,
 # which is defined as a Kubernetes secret and exposed to the coordinator pod.
 # See: k8s/coordinator/base/deployment.yaml
+
+[s3]
+# The access_key and secret_access_key are configured via the enviroment variables
+# `XAYNET_S3__ACCESS_KEY` and `XAYNET_S3__SECRET_ACCESS_KEY`.
+# See: k8s/coordinator/base/deployment.yaml
+region = ["minio", "http://minio:9000"]
```
rsaffi, line 42: What exactly is with passing the `region`? Is this the official way of informing the endpoint of the S3 endpoint?
I would assume that there would be at least 4 parameters:
* Bucket
* Endpoint
* accessKey
* secretKey

Robert-Steiner: Yes, that is the [official way](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#create-bucket-intro) 
The endpoint is implicit given via the [region code](https://docs.aws.amazon.com/general/latest/gr/rande.html). However if you don't use S3 directly but an AWS-compatible services then you have to specify the endpoint.

I added the the bucket name to the settings but made them optional (uses a default value if not present)

rust/xaynet-server/src/settings.rs
```diff
@@ -558,6 +564,181 @@ where
...
+            formatter.write_str("sequence of \"name Optional<endpoint>\"")
+        }
+
+        // FIXME: a copy of https://rusoto.github.io/rusoto/src/rusoto_core/region.rs.html#221-251\
+        // I haven't managed to create a sequence and call `self.visit_seq(seq)`.
+        fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            let mut seq = value.split_whitespace();
```
little-dude, line 714: Could we just `collect` the iter to create a seq and then call `self.visit_seq(seq)` ?
Something like:

```rust
        fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
        where
            E: de::Error,
        {
            let mut seq = value.split_whitespace().collect();
            self.visit_seq(seq)
        }
```

little-dude (commented):
Thanks for all the comments and documentation. I like the PR overall but I find the region de-serialization implementation a bit complex.

rust/xaynet-server/src/settings.rs
```diff
@@ -558,6 +564,181 @@ where
...
+    }
+}
+
+// A small wrapper to support the list type for environment variable values.
+// config-rs always converts a environment variable value to a string
+// https://github.com/mehcode/config-rs/blob/master/src/env.rs#L114 .
+// Strings however, are not supported by the deserializer of rusoto_core::Region (only sequences).
+// Therefore we use S3RegionVisitor to implement `visit_str` and thus support
+// the deserialization of rusoto_core::Region from strings.
+fn deserialize_s3_region<'de, D>(deserializer: D) -> Result<Region, D::Error>
```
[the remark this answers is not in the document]
Robert-Steiner: I have added some tests. It adds a lot of code tbh. Maybe we should spilt the settings.rs into multiple files in the near future. Unfortunately I have to use `serial` here because I set environment variables in some tests.
Let me know what you think🙂

rust/xaynet-server/src/settings.rs
Robert-Steiner: Yes, that was my plan. The issue is that the deserializer of `Region` expects a sequence of `(String, Optional<String>)` so unfortunately I can't use a `vec` here. I tried a tuple, but I wasn't able to create a sequence from a tuple.

rust/xaynet-server/src/settings.rs
```diff
@@ -558,6 +564,181 @@ where
...
+    struct S3RegionVisitor;
+
+    impl<'de> Visitor<'de> for S3RegionVisitor {
+        type Value = Region;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("sequence of \"name Optional<endpoint>\"")
+        }
+
+        // FIXME: a copy of https://rusoto.github.io/rusoto/src/rusoto_core/region.rs.html#221-251\
```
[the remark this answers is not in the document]
Robert-Steiner: updated the url

rust/xaynet-server/src/settings.rs
little-dude: This seems to compile:

```rust
        fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
        where
            E: de::Error,
        {
            let seq = value.split_whitespace();
            self.visit_seq(serde::de::value::SeqDeserializer::new(seq))
        }
```

little-dude: Ah but the tests fail:

```
---- settings::tests::test_s3_custom_region_env stdout ----
thread 'settings::tests::test_s3_custom_region_env' panicked at 'called `Result::unwrap()` on an `Err` value: invalid type: string "http://localhost:8000", expected option', xaynet-server/src/settings.rs:1152:57
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- settings::tests::test_s3_custom_region_toml stdout ----
thread 'settings::tests::test_s3_custom_region_toml' panicked at 'called `Result::unwrap()` on an `Err` value: invalid type: string "http://localhost:8000", expected option', xaynet-server/src/settings.rs:1107:57
```

little-dude: Ok let's forget about this. I don't think we can do better.

little-dude (approved):
:+1:

rust/xaynet-server/src/settings.rs
```diff
@@ -558,6 +564,181 @@ where
...
+    }
+}
+
+// Validates the bucket name
+// [Rules for AWS bucket naming](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html)
+fn validate_s3_bucket_name(bucket_name: &str) -> Result<(), ValidationError> {
+    // https://stackoverflow.com/questions/50480924/regex-for-s3-bucket-name#comment104807676_58248645
+    // I had to use fancy_regex here because the std regex does not support `look-around`
+    let re =
+        Regex::new(r"(?!^(\d{1,3}\.){3}\d{1,3}$)(^[a-z0-9]([a-z0-9-]*(\.[a-z0-9])?)*$(?<!\-))")
```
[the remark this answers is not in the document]
Robert-Steiner: I created a story for refactoring the settings and add this as one of the tasks

PR #552, opened 2020-10-12
Mobile client multipart

little-dude:
This is a minimalistic implementation of multipart messages for the mobile client. The following limitations will be addressed in follow-up PRs:

- there is no retry mechanism. If a chunk cannot be sent, the client doesn't try to re-send it.
- chunks are sent sequentially. We could make it possible to send several chunks in parallel
- if a chunk cannot be sent, the client doesn't try to send the next chunks
- the maximum chunk size is currently hard-coded

This is based on https://github.com/xaynetwork/xaynet/pull/543

### DIFF
--- a/rust/xaynet-client/src/mobile_client/client.rs
+++ b/rust/xaynet-client/src/mobile_client/client.rs
@@ -64,7 +64,7 @@ impl<Type> ClientState<Type> {
         payload: Payload,
     ) -> Result<(), ClientError<T::Error>> {
         // Unwrapping is fine because this only errors out if the
-        // payload if a Chunk, which we never create in the client.
+        // payload is a Chunk, which we never create in the client.
         let encoder = MessageEncoder::<'_, Type>::new(
             &self.participant,
             payload,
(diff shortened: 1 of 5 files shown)


Robert-Steiner (approved):
Already reviewed the changes in #543

rust/xaynet-client/src/mobile_client/client.rs
```diff
@@ -46,6 +57,27 @@ impl<Type> ClientState<Type> {
...
         ClientState::<Awaiting>::new(self.participant.reset(), self.round_params)
     }
+
+    async fn send_message<T: ApiClient>(
+        &self,
+        api: &mut T,
+        payload: Payload,
+    ) -> Result<(), ClientError<T::Error>> {
+        // Unwrapping is fine because this only errors out if the
+        // payload if a Chunk, which we never create in the client.
```
finiteprods, line 67:
```suggestion
        // payload is a Chunk, which we never create in the client.
```

finiteprods (approved):
looks good :+1:

PR #556, opened 2020-10-13
fix upload of bench artifacts

little-dude:


.github/workflows/rust.yml
```diff
@@ -141,7 +141,7 @@ jobs:
       - uses: actions/upload-artifact@v1
```
Robert-Steiner, line 141: I saw there is also a version 2 
`actions/upload-artifact@v2`

Robert-Steiner (approved):

.github/workflows/rust.yml
little-dude: Ok let me update this then

PR #557, opened 2020-10-13
mask object and unmasking improvements

finiteprods:
This merge request implements some minor refactoring for further improvements to the mask object.

- Renames throughout for more consistency. In particular, `MaskMany` and `MaskOne` are renamed `MaskVect` and `MaskUnit`, respectively, as are the accessor fields
```rust
pub struct MaskObject {
    pub vect: MaskVect,
    pub unit: MaskUnit,
}
```
- Reorganised `mask/object/serialization.rs` into
```
- mask
    - object
        - serialization
            - mod.rs
            - vect.rs
            - unit.rs // in a separate PR
```
for a cleaner separation of the serialization code for mask object and its constituents. Also acts as a precursor to more improvements to mask unit serialization, which will be in `unit.rs` in another PR.

- Removed unnecessary "scalar accessor" code in the `Sum2Buffer` and `UpdateBuffer`. The sum2 and update payloads themselves also (de)serialize more directly by working with mask objects rather than the constituent parts.

- (somewhat separate from the above) Reordered the steps of the unmasking for more memory efficiency.

TODO
- [ ] tidy up commits

### DIFF
--- a/rust/xaynet-core/src/mask/masking.rs
+++ b/rust/xaynet-core/src/mask/masking.rs
@@ -21,7 +21,7 @@ use crate::{
     mask::{
         config::MaskConfig,
         model::{float_to_ratio_bounded, Model},
-        object::{MaskMany, MaskObject, MaskOne},
+        object::{MaskObject, MaskOne, MaskVect},
         seed::MaskSeed,
     },
 };
@@ -399,7 +399,7 @@ impl Masker {
                 (shifted + rand_int) % &order
             })
             .collect();
-        let masked_model = MaskMany::new(config_model, masked_weights);
+        let masked_model = MaskVect::new(config_model, masked_weights);
 
         // mask the scalar
         // PANIC_SAFE: shifted scalar is guaranteed to be non-negative
@@ -629,7 +629,7 @@ mod tests {
                         let integers = iter::repeat_with(|| generate_integer(&mut prng, &order))
                             .take($len as usize)
                             .collect::<Vec<_>>();
-                        let model = MaskMany::new(config, integers);
+                        let model = MaskVect::new(config, integers);
                         let scalar = MaskOne::empty(config);
                         MaskObject::new(model, scalar)
                     });
--- a/rust/xaynet-core/src/mask/masking.rs
+++ b/rust/xaynet-core/src/mask/masking.rs
@@ -497,7 +497,7 @@ mod tests {
                     // c. unmask the model and check it against the original one.
                     let (mask_seed, masked_model) =
                         Masker::new(config.clone(), config.clone()).mask(1_f64, model.clone());
-                    assert_eq!(masked_model.vector.data.len(), model.len());
+                    assert_eq!(masked_model.vect.data.len(), model.len());
                     assert!(masked_model.is_valid());
 
                     let mask = mask_seed.derive_mask(model.len(), config.clone(), config.clone());
@@ -626,7 +626,7 @@ mod tests {
                             .take($len as usize)
                             .collect::<Vec<_>>();
                         let model = MaskVect::new(config, integers);
-                        let scalar = MaskOne::empty(config);
+                        let scalar = MaskUnit::default(config);
                         MaskObject::new(model, scalar)
                     });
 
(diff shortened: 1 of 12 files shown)


rust/xaynet-core/src/mask/object/mod.rs
```diff
@@ -103,66 +100,55 @@ impl MaskOne {
...
-    ) -> Self {
-        Self {
-            vector: MaskMany::new(config_v, data_v),
-            scalar: MaskOne::new(config_s, data_s),
-        }
-    }
-
-    // TODO doc
+    /// Creates a new mask object from the given vector, unit and masking configurations.
     pub fn new_checked(
```
little-dude, line 131: I'm not sure we need this method. Could we have only `MaskObject::new` which takes a `MaskVect` and `MaskUnit` that are assumed to be valid, and `Mask{Vect,Unit}::new` that always perform the validation? That way we'd also remove `Mask{Vect,Unit}::new_checked`.

Also, if we really need both the checked and unchecked variant, my personal perference would be to have `new` and `new_unchecked` instead of `new_checked` and `new`. It is more obvious for the caller that the method skips some potentially important checks.

rust/xaynet-core/src/message/payload/sum2.rs
```diff
@@ -102,15 +102,6 @@ impl<T: AsRef<[u8]> + AsMut<[u8]>> Sum2Buffer<T> {
         let offset = self.model_mask_offset();
         &mut self.inner.as_mut()[offset..]
     }
-
```
little-dude, line 105: There are nice simplifications for the `sum2` and `update` module :+1:

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -190,18 +188,26 @@ impl Aggregation {
...
-    pub fn unmask(mut self, mask_obj: MaskObject) -> Model {
-        // unmask (over-scaled) global model
-        let scaled_add_shift_n =
-            self.object.vector.config.add_shift() * BigInt::from(self.nb_models);
-        let exp_shift_n = self.object.vector.config.exp_shift();
-        let order_n = self.object.vector.config.order();
-        let model: Model = self
-            .object
-            .vector
+    pub fn unmask(self, mask_obj: MaskObject) -> Model {
```
janpetschexain, line 191: we might destructure `self` and `mask_obj` before using it below, but that's mostly a matter of taste and up to your decision (i also removed the notion of `weight` and updated the `panic_safe` comment):
```rust
pub fn unmask(self, mask_obj: MaskObject) -> Model {
    let Aggregation {
        nb_models,
        object:
            MaskObject {
                vect:
                    MaskVect {
                        data: masked_n,
                        config: config_n,
                    },
                unit:
                    MaskUnit {
                        data: masked_1,
                        config: config_1,
                    },
            },
        ..
    } = self;
    let MaskObject {
        vect: MaskVect { data: mask_n, .. },
        unit: MaskUnit { data: mask_1, .. },
    } = mask_obj;

    // unmask scalar sum
    let scaled_add_shift_1 = config_1.add_shift() * BigInt::from(nb_models);
    let exp_shift_1 = config_1.exp_shift();
    let order_1 = config_1.order();
    let n = (masked_1 + &order_1 - mask_1) % &order_1;
    let ratio = Ratio::<BigInt>::from(n.to_bigint().unwrap());
    let scalar_sum = ratio / &exp_shift_1 - &scaled_add_shift_1;

    // unmask global model
    let scaled_add_shift_n = config_n.add_shift() * BigInt::from(nb_models);
    let exp_shift_n = config_n.exp_shift();
    let order_n = config_n.order();
    masked_n
        .into_iter()
        .zip(mask_n)
        .map(|(masked, mask)| {
            // PANIC_SAFE: The substraction panics if it
            // underflows, which can only happen if:
            //
            //     mask > order_n
            //
            // If the mask is valid, we are guaranteed that this
            // cannot happen. Thus this method may panic only if
            // given an invalid mask.
            let n = (masked + &order_n - mask) % &order_n;

            // UNWRAP_SAFE: to_bigint never fails for BigUint
            let ratio = Ratio::<BigInt>::from(n.to_bigint().unwrap());
            let unmasked = ratio / &exp_shift_n - &scaled_add_shift_n;

            // scaling correction
            unmasked / &scalar_sum
        })
        .collect()
}
```

janpetschexain (commented):
looks good, this improves readability quite a bit.

rust/xaynet-core/src/mask/object/mod.rs
janpetschexain: agree, the `..._unchecked` notation is also consistent with the std library.

janpetschexain (approved):

rust/xaynet-core/src/mask/object/mod.rs
finiteprods: thanks both. Renamed `new`/`new_checked` to `new_unchecked`/`new` accordingly. Currently there's use of both so I've left both variants in, but I agree it can probably be consolidated, at a later date.

rust/xaynet-core/src/mask/masking.rs
finiteprods: nice! I've destructured like so
```rust
let (masked_n, config_n) = (self.object.vect.data, self.object.vect.config);
let (masked_1, config_1) = (self.object.unit.data, self.object.unit.config);
let mask_n = mask_obj.vect.data;
let mask_1 = mask_obj.unit.data;
```
to save a few lines, otherwise taken your suggestion.

PR #560, opened 2020-10-13
Bind error codes to error variants

Robert-Steiner:
small follow up PR based on https://github.com/xaynetwork/xaynet/pull/538#discussion_r500269333

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -497,6 +497,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derivative"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb582b60359da160a9477ee80f15c8d784c477e69c217ef2cdd4169c24ea380f"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "derive_more"
 version = "0.99.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


finiteprods (commented):
this looks nice, thanks for trying it out. It's good to have the error codes defined together with the error variants, and there's even a modest saving in LoC (more than I was expecting with this `num_enum` crate, nice find!). Looks very promising to me. :1st_place_medal:

little-dude (approved):
This would be nice to have yes. I initially thought we didn't need `num_enum` since we can already [explicitely choose an enum discriminant](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations). However this only gives us the enum -> integer conversion, while we also need the integer -> enum one. `num_enum` provides that.

janpetschexain (approved):

finiteprods (commented):
looks great. one minor comment: I notice sometimes `0` represents OK, negative numbers are errors. other times, `1` is OK and `0` is error. should we settle on a convention?

Robert-Steiner:
I agree, that would make it more consistent but then I have to add code that converts the numbers because I in this case simply use the return values ​​of the redis commands.

e.g. in the `add_sum_participant` the return values comes from `hset_nx`

https://github.com/xaynetwork/xaynet/blob/6eec7ec2f4b9a9520bf59257c0e4176442c7d895/rust/xaynet-server/src/storage/redis.rs#L160-L179

finiteprods:
> I agree, that would make it more consistent but then I have to add code that converts the numbers because I in this case simply use the return values ​​of the redis commands.

Ah, ok. Yep, makes sense to keep it as it is then.

PR #561, opened 2020-10-14
gather some test utilities under `xaynet_core::testutils`

little-dude:
This makes some helpers for crafting messages re-usable from other
crates. `xaynet_core::testutils` requires the "testutils" feature.

This commit also moves the benchmarks to a separate crate.


rust/benches/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
Robert-Steiner, line 1:
```suggestion

```

rust/xaynet-client/Cargo.toml
```diff
@@ -19,13 +19,14 @@ bincode = "1.3.1"
 thiserror = "1.0.20"
 tracing = "0.1.19"
 async-trait = "0.1.40"
-xaynet-core = { path = "../xaynet-core", version = "0.1.0" }
+xaynet-core = { path = "../xaynet-core", version = "0.1.0", features = [] }
```
Robert-Steiner, line 22: Do we need to add `features` here?

rust/benches/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "benches"
+version = "0.0.0"
+authors = ["Xayn Engineering <engineering@xaynet.dev>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+criterion = "0.3.3"
+xaynet-core = { version = "0.1.0", path = "../xaynet-core", features = ["testutils"] }
```
Robert-Steiner, line 10: Not sure but do we need to add 
```toml
[[bench]]
name = "messages"
```
here?

Robert-Steiner (approved):
The feature flag is a bit hacky, but I don't have a better solution 😅 However, we only add the flag in one cfg annotation, so I'm totally fine with that. Like the idea of the benches crate 👍

little-dude:
@Robert-Steiner thanks for the review (and for spotting that the benchmarks were actually broken :confounded: ). I think I addressed all of your comments. Let's see if the CI is happy.

janpetschexain (approved):
looks good. the tuple outputs of the helper functions are sometimes a bit hard to read when used, one needs to jump back and forth between definition and call. sth like a struct output with named fields would be nice to read, but this would also introduce a lot of additional code. those helpers are anyways only meant for internal usage so i guess that's ok to have tuples.

finiteprods (commented):
if I'm understanding the compromise here: although an update message helper feels like it ought to live with the update message module (say), the fact is it's used for tests in multiple places. so it's justifiable for such helpers to live in a shared crate (and necessary even, to be accessible from a different crate like benchmarks). And we put the helpers behind a feature flag because we consider them "internal".

little-dude:
> if I'm understanding the compromise here: although an update message helper feels like it ought to live with the update message module (say), the fact is it's used for tests in multiple places. so it's justifiable for such helpers to live in a shared crate (and necessary even, to be accessible from a different crate like benchmarks). And we put the helpers behind a feature flag because we consider them "internal".

Exactly

PR #564, opened 2020-10-15
make the max message size configurable on the client

little-dude:


rust/xaynet-client/src/mobile_client/participant/mod.rs
```diff
@@ -32,19 +34,93 @@ pub struct ParticipantState {
...
 
 #[derive(Serialize, Deserialize)]
 pub struct ParticipantSettings {
     pub secret_key: ParticipantSecretKey,
     pub aggregation_config: AggregationConfig,
+    pub max_message_size: MaxMessageSize,
+}
+
+const MESSAGE_HEADER_LEN: usize = 136;
+const ENCRYPTION_HEADER_LEN: usize = 48;
```
janpetschexain, line 48: those constants are duplicates of `xaynet_core::message::message::HEADER_LENGTH` and `xaynet_core::crypto::encrpyt::SEALBYTES`, we should use the original ones instead of redefining them here.

rust/xaynet-client/src/mobile_client/participant/mod.rs
```diff
@@ -32,19 +34,93 @@ pub struct ParticipantState {
...
+pub const MIN_MESSAGE_SIZE: usize = MESSAGE_HEADER_LEN + ENCRYPTION_HEADER_LEN + 1;
```
janpetschexain, line 49: would be good to have some documentation here, especially regarding the `+1`, i guess it means that the payload must be at least one byte?

rust/xaynet-client/src/mobile_client/participant/mod.rs
```diff
@@ -32,19 +34,93 @@ pub struct ParticipantState {
...
+pub struct InvalidMaxMessageSize;
+
+/// Represent the maximum size messages sent by a participant can
+/// have. If a larger message needs to be sent, it will be chunked and
+/// sent in several parts. Note that messages have a minimal size of
+/// [`MIN_MESSAGE_SIZE`].
+#[derive(Serialize, Deserialize, Clone, Copy, Debug)]
+pub struct MaxMessageSize(
+    #[serde(deserialize_with = "deserialize_max_message_size")] Option<usize>,
+);
```
janpetschexain, line 61: this represents no splitting with `None` whereas the multipart encoding uses `0`. it would be nice to unify this to use only one of the two approaches, maybe in a later refactoring pr, i guess the `Option<usize>` is preferable over the `usize`.

little-dude: Yeah I agree. I hesitated but wanted to keep the changes small for this PR. The encoder also uses the _payload_ size while the settings use the _encrypted message_ size. We should also consolidate that as well.

rust/xaynet-client/src/mobile_client/participant/mod.rs
little-dude: Good poitn

Robert-Steiner (approved):
Looks good to me 👍

janpetschexain (approved):
lgtm, ready once rustfmt is happy.

PR #567, opened 2020-10-15
Integrate S3 into state machine

Robert-Steiner:
**Summary**
- integrated s3 client into state machine 
- moved s3 settings into a separate file to reduce the number of feature flag annotations
- adjusted tests

The changes are similar to the redis integration pr. Many of the changes result from separating the s3 settings from the rest of the settings. It took me some time to figure out how the validation of the s3 settings doesn't lead to an error if we don't compile the coordinator with the `model-persistence` feature.

## Error handling
If the upload of the global model fails, we restart the round.

## S3 client initialisation
When the coordinator starts, the s3 client establishes a connection to s3 / minio and tries to create the bucket for storing the global models. This operation is successful if:
- the bucket doesn't exist yet
- the bucket has already been created
- the bucket is already owned by us.

Any other error leads to a crash of the coordinator. The same applies if the s3 client cannot establish a connection.

## Further details
- the key/id of the global model is the round id and round seed of the current round.
  - format: `roundid_roundseed`
- we only store global model, we never delete them (not even when re/starting the coordinator)


Robert-Steiner:
#build

rust/xaynet-server/src/state_machine/phases/error.rs
```diff
@@ -25,6 +25,9 @@ pub enum PhaseStateError {
     Timeout(#[from] tokio::time::Elapsed),
     #[error("redis request failed: {0}")]
     Redis(#[from] RedisError),
+    #[cfg(feature = "model-persistence")]
+    #[error("s3 request failed: {0}")]
+    S3(#[from] crate::storage::s3::S3Error),
```
little-dude, line 30: Do we actually use the inner error? If not I think we don't have to store the inner error. We could have `SaveGlobalModel` and `LoadGlobalModel` variants instead, which seem more specific to the phases logic.

little-dude (approved):
I haven't reviewed the settings changes, because it seems that you just split the module.

The model persistence changes look great :+1:

rust/xaynet-server/src/settings/mod.rs
```diff
@@ -0,0 +1,623 @@
...
+use validator::{Validate, ValidationError, ValidationErrors};
+
+use xaynet_core::mask::{BoundType, DataType, GroupType, MaskConfig, ModelType};
+
+#[cfg(test)]
+mod tests;
+
+#[cfg(not(feature = "model-persistence"))]
+#[derive(Debug, Validate, Deserialize, Default)]
+pub struct S3Settings {}
```
janpetschexain, line 24: with the suggestion mentioned in `xaynet-server/src/bin/main.rs` this is not needed anymore, which means we can eliminate the notion of a s3 setting entirely if the corresponding feature flag is not set. see also the required addtion in lines 53-55.
```suggestion
```

rust/xaynet-server/src/settings/mod.rs
```diff
@@ -0,0 +1,623 @@
...
+    pub pet: PetSettings,
+    pub mask: MaskSettings,
+    pub log: LoggingSettings,
+    pub model: ModelSettings,
+    #[validate]
+    pub metrics: MetricsSettings,
+    pub redis: RedisSettings,
+    #[validate]
+    #[cfg_attr(not(feature = "model-persistence"), serde(default))]
+    pub s3: S3Settings,
```
janpetschexain, line 55: as suggested above, we could gate this entirely behind the corresponding feature flag as follows:
```suggestion
    #[cfg(feature = "model-persistence")]
    #[validate]
    pub s3: S3Settings,
```

janpetschexain (approved):
looks good, i left a suggestion to eliminate the placeholder `S3Settings` entirely if the feature flag is not set.

rust/xaynet-server/src/state_machine/phases/error.rs
Robert-Steiner: Yes that is a good idea👍

rust/xaynet-server/Cargo.toml
```diff
@@ -82,6 +84,7 @@ path = "src/bin/main.rs"
 
 [features]
 default = []
-full = ["tls", "metrics"]
 tls = ["warp/tls"]
 metrics = ["influxdb", "chrono"]
+model-persistence = ["rusoto_core", "rusoto_s3", "hex", "fancy-regex"]
+full = ["tls", "metrics", "model-persistence"]
```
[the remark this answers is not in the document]
Robert-Steiner: Thats right, it only makes sense if xaynet-server is used as a lib then it is easier to add all features via `full` in the cargo.toml
```
xaynet-server = { version = "0.1.0", features = ["full"] }
```
However, I created it to make it easier for building a docker container with all features

https://github.com/xaynetwork/xaynet/blob/98153d6640e6939fefb3ad2da15f9c8d043cfbbf/docker/Dockerfile#L25-L29

rust/xaynet-server/src/bin/main.rs
```diff
@@ -37,7 +40,7 @@ async fn main() {
         model: model_settings,
         metrics: metrics_settings,
         redis: redis_settings,
-        ..
+        s3: s3_settings,
```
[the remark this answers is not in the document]
Robert-Steiner: it works, a bit surprised why🤔

rust/xaynet-server/src/state_machine/phases/error.rs
Robert-Steiner: I tried to implement it but not sure how to implement a quick good solution here. I think it is still good to keep the inner error so that we can log the s3 error here

https://github.com/xaynetwork/xaynet/blob/7f5dc2eb65787654a9dfa0be5f734702b0684dd1/rust/xaynet-server/src/state_machine/phases/error.rs#L45

Otherwise I would need to log the error here:
https://github.com/xaynetwork/xaynet/blob/2d6cededc16a95be503cec4a8a4f7f7db53fea30/rust/xaynet-server/src/state_machine/phases/unmask.rs#L77
or in the s3 client. 

But I like the idea to hide the s3 and redis errors behind more higher level error variants.

Is it fine if we pospone it and maybe combine it with the ideas that I wrote in #544 (defining traits and using `tower`)?

janpetschexain (approved):

little-dude (approved):

PR #570, opened 2020-10-20
XN-1178 doc forbid warnings

janpetschexain:
**References**
- [XN-1178](https://xainag.atlassian.net/browse/XN-1178)

**Summary**
The `cargo doc` respectively `rustdoc` seems not to provide an option like `--deny warnings` as for example in `clippy`. Therefore we can't simply attach this to the CI command, but have to specify it in code. `forbid` was chosen over `deny` to avoid `allow`ing locally by accident. The dead documentations links have been fixed as well.


Robert-Steiner (approved):
Nice! 👍

PR #572, opened 2020-10-20
Add --all-targets / --all-features flag

Robert-Steiner:
- Adds `--all-targets` flag to `cargo check` and `cargo test`
- Adds a separate command that runs `cargo check` and `cargo test` with `--all-features` flag


janpetschexain (approved):

janpetschexain:
btw, this pr has the changes of #571 as well, which may lead to some confusion.

Robert-Steiner:
> btw, this pr has the changes of #571 as well, which may lead to some confusion.
I removed the changes of #571

PR #573, opened 2020-10-20
XN-1132 enable clippy everywhere

janpetschexain:
**References**
- [XN-1132](https://xainag.atlassian.net/browse/XN-1132)

**Summary**
Enables `cargo clippy` for all targets (including tests) on the CI and fixes the newly raised issues. Also enables `clippy::cargo` lints, those are off by default and check some best practices for the `Cargo.toml` files. Fixed the issues within the `package` section and reordered them in the same way as the rust book. Also took the opportunity to sort the dependencies alphanumerically to improve findability (if the sorting is deemed too much, i can also revert this).

**Todo**
- [x] fix remaining clippy issues
- [x] check if enabling clippy for manifests is a viable option
- [x] wait for #567 to be merged to rebase

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -95,7 +95,7 @@ jobs:
 
       - name: cargo clippy
         working-directory: ./rust
-        run: cargo clippy -- -D warnings
+        run: cargo clippy --all-targets -- -D warnings
 
   test:
     name: cargo-test
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -95,7 +95,9 @@ jobs:
 
       - name: cargo clippy
         working-directory: ./rust
-        run: cargo clippy --all-targets -- -D warnings
+        run: |
+          cargo clippy --all-targets -- --deny warnings
+          cargo clippy --all-targets --all-features -- --deny warnings
 
   test:
     name: cargo-test
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -96,8 +96,8 @@ jobs:
       - name: cargo clippy
         working-directory: ./rust
         run: |
-          cargo clippy --all-targets -- --deny warnings
-          cargo clippy --all-targets --all-features -- --deny warnings
+          cargo clippy --all-targets -- --deny warnings --deny clippy::cargo
+          cargo clippy --all-targets --all-features -- --deny warnings --deny clippy::cargo
 
   test:
     name: cargo-test
--- a/rust/examples/mobile-client.rs
+++ b/rust/examples/mobile-client.rs
@@ -181,12 +181,11 @@ fn perform_task(
         Err((client, _)) => client,
     };
 
-    match client.get_global_model().unwrap() {
-        Some(model) => println!(
+    if let Some(model) = client.get_global_model().unwrap() {
+        println!(
             "global model: {:?}",
             model.into_primitives_unchecked().collect::<Vec<f32>>()
-        ),
-        _ => (),
+        );
     };
 
     let new_bytes = client.serialize();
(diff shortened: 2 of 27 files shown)


.github/workflows/rust.yml
```diff
@@ -93,7 +93,7 @@ jobs:
 
       - name: cargo clippy
         working-directory: ./rust
-        run: cargo clippy -- -D warnings
+        run: cargo clippy --all-targets -- -D warnings
```
Robert-Steiner, line 96: here we can also add the command to check the code that is hidden behind the feature flags
```suggestion
        run: |
          cargo clippy --all-targets -- -D warnings
          cargo clippy --all-targets --all-features -- -D warnings
```

rust/xaynet-client/src/mobile_client/participant/awaiting.rs
```diff
@@ -115,7 +115,7 @@ mod tests {
 
         match part.determine_role(eligible_sum_seed, 0.5_f64, 0.5_f64) {
             Role::Summer(_) => (),
-            _ => assert!(false),
+            _ => panic!(),
```
Robert-Steiner, line 118: maybe we can  use `assert!(matches!(...))` instead of `panic!()` here?

Robert-Steiner (commented):
wow so many improvements nice!

rust/xaynet-client/src/mobile_client/participant/awaiting.rs
janpetschexain: good idea, i'll change that for all tests in that module.

it would be nice to have a negation of `if let`, that would make things even more clear.

rust/examples/mobile-client.rs
```diff
@@ -181,12 +181,11 @@ fn perform_task(
         Err((client, _)) => client,
     };
 
-    match client.get_global_model().unwrap() {
-        Some(model) => println!(
+    if let Some(model) = client.get_global_model().unwrap() {
```
little-dude, line 184: we can have `if let` with an `else` clause now? :exploding_head: This is cool

janpetschexain: i think omitting the `else` clause of the `if let` only works because the `if` clause has the `!` return type. i guess it won't work in a variable assignment.

little-dude (approved):
:+1: nice cleanups in the code. Regarding sorting the deps, I mean... who am I to judge your fetishes :laughing:

rust/xaynet-macros/Cargo.toml
```diff
@@ -3,14 +3,17 @@ name = "xaynet-macros"
 version = "0.1.0"
 authors = ["Xayn Engineering <engineering@xaynet.dev>"]
 edition = "2018"
-license-file = "../../LICENSE"
-repository = "https://github.com/xaynetwork/xaynet/"
+description = "The Xayn Network project is building a privacy layer for machine learning so that AI projects can meet compliance such as GDPR and CCPA. The approach relies on Federated Learning as enabling technology that allows production AI applications to be fully privacy compliant."
+readme = "../../README.md"
```
Robert-Steiner, line 7: Nice didn't know that this is possible 👍

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -74,7 +73,7 @@ impl Phase for PhaseState<Unmask> {
                 .s3
                 .upload_global_model(&key, &global_model)
                 .await
-                .map_err(|err| PhaseStateError::SaveGlobalModel(err))?;
+                .map_err(PhaseStateError::SaveGlobalModel)?;
```
Robert-Steiner, line 76: ok wow also didn't know that this is possible 😅

Robert-Steiner (approved):
LGTM👍

rust/xaynet-server/src/state_machine/phases/unmask.rs
janpetschexain: that was one of the suggestions by clippy

PR #577, opened 2020-10-26
fix dropping message in select branch

Robert-Steiner:
Addresses an issue where a message may be dropped after the delay has elapsed.

Found the solution in the [tokio::select! documentation](https://docs.rs/tokio/0.2.22/tokio/macro.select.html#avoid-racy-if-preconditions). 

We should check later if we have the same error in phase timeout.


janpetschexain (approved):

little-dude (approved):
Good catch, and very nice fix :+1:

PR #579, opened 2020-10-27
XN-1209 allow the deployment of a pr on dev k8s using /build_pr args

abargiela:
Add the option of when you add a comment using: 
```/build_pr args```
It will trigger the build of the container and deploy it to k8s

### DIFF
--- a/.github/workflows/dockerhub-pr-with-parameters.yml
+++ b/.github/workflows/dockerhub-pr-with-parameters.yml
@@ -1,6 +1,4 @@
 on:
-  pull_request:
-    types: [opened]
   issue_comment:
     types: [created]
     
@@ -20,7 +18,7 @@ jobs:
           reaction: "true"
           reaction-type: "eyes"
           allow-edits: "false"
-          permission-level: write
+          permission-level: maintain
       - name: Act on the command
         run: echo "The command was '${{ steps.command.outputs.command-name }}' with arguments '${{ steps.command.outputs.command-arguments }}'"
       # Build container
@@ -39,18 +37,6 @@ jobs:
           find: "newTag: development"
           replace: "newTag: ${{ steps.comment-branch.outputs.head_ref }}"
           include: "kustomization.yaml"
-
-      - name: build-tag-push-pr
-        uses: docker/build-push-action@v1
-        with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
-          dockerfile: docker/Dockerfile
-          cache_froms: buildpack-deps:stable-curl
-          repository: xaynetwork/xaynet
-          tags: ${{ steps.pr_data.outputs.branch }}
-          push: true
-          build_args: '${{ steps.command.outputs.command-arguments }}'
       # k8s deploy
       - name: build-tag-push
         uses: docker/build-push-action@v1
--- a/.github/workflows/dockerhub-pr-with-parameters.yml
+++ b/.github/workflows/dockerhub-pr-with-parameters.yml
@@ -1,4 +1,9 @@
-on: issue_comment
+on:
+  pull_request:
+    types: [opened]
+  issue_comment:
+    types: [created]
+    
 name: DockerHub (PR) with parameters
 
 jobs:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/dockerhub-pr-with-parameters.yml [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+      # Build container
+      - uses: xt0rted/pull-request-comment-branch@v1
+        id: comment-branch
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v2
+        if: success()
+        with:
+          ref: ${{ steps.comment-branch.outputs.head_ref }}
+      - run: git rev-parse --abbrev-ref HEAD
```
Robert-Steiner, line 30: we can remove these lines here

```
      - run: git rev-parse --abbrev-ref HEAD
      - run: git rev-parse --verify HEAD
      - name: Clone branch of the PR
        uses: actions/checkout@v2
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          ref: ${{ steps.pr_data.outputs.branch }}
```
because we already cloned the branch in line 26

.github/workflows/dockerhub-pr-with-parameters.yml
```diff
@@ -0,0 +1,101 @@
...
+          ref: ${{ steps.pr_data.outputs.branch }}
+      - name: build-tag-push-pr
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          dockerfile: docker/Dockerfile
+          cache_froms: buildpack-deps:stable-curl
+          repository: xaynetwork/xaynet
+          tags: pr-${{ steps.pr_data.outputs.branch }}
```
Robert-Steiner, line 45:
```suggestion
          tags: pr-${{ steps.comment-branch.outputs.head_ref }}
```

.github/workflows/dockerhub-pr-with-parameters.yml
```diff
@@ -0,0 +1,101 @@
...
+          push: true
+          build_args: '${{ steps.command.outputs.command-arguments }}'
+      # k8s deploy
+      - name: build-tag-push
```
Robert-Steiner, line 49: ahh just discovered a problem
once we restart the coordinator pod, k8s tries to pull the docker image with the tag `development` but in our case it is the name of the branch. Not sure how we can make this configurable. I think we need to think about it.

Robert-Steiner (commented):
Looks good 👍
just some lines that we can remove 
Can you please remove the deployment part. it won't work the way i thought it would

.github/workflows/dockerhub-pr-with-parameters.yml
abargiela: This action seems to solve this necessity 
```
      - name: Find and Replace
        uses: jacobtomlinson/gha-find-replace@master
```

.github/workflows/dockerhub-pr-with-parameters.yml
abargiela: I also remove the prefix `pr-` seems not be necessary anymore

.github/workflows/dockerhub-pr-with-parameters.yml
```diff
@@ -0,0 +1,106 @@
+on:
+  pull_request:
```
Robert-Steiner, line 2: these two line can be removed
```
pull_request:
types: [opened]
```
it should work with only 
```
  issue_comment:
    types: [created]
```

.github/workflows/dockerhub-pr-with-parameters.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Check for Command
+        id: command
+        uses: xt0rted/slash-command-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          command: build_pr
+          reaction: "true"
+          reaction-type: "eyes"
+          allow-edits: "false"
+          permission-level: write
```
Robert-Steiner, line 23:
```suggestion
          permission-level: maintain
```

.github/workflows/dockerhub-pr-with-parameters.yml
Robert-Steiner: Nice!
can you please remove the `build-tag-push-pr` (43-L53) step because we have this step twice

abargiela: Done

Robert-Steiner (approved):
LGTM🚀

PR #582, opened 2020-10-28
XN-1460 add model conversion benchmarks

wilk10:
Adds benchmarks for parsing a vector `i32` into a Model, and serialising a Model into a vector `i32`.

Performance:
| test | time |
| ---    | ---    |
| parse model from 4 bytes vector   |   351.78 ns |
| parse model (bounded) from 100kB vector   |   3.4069 ms |
| parse model from 1MB vector   |   41.986 ms |
| serialize 4 bytes model to primitives   |   829.05 ns |
| serialize 100kB model into primitives   |   9.5551 ms |
| serialize 1MB model to primitives   |   103.23 ms |

I tried with 10MB as well but preparing the 1000 iterations was too painfully slow.

Not sure this is the best way, so please don't hold back on comments.

I also get a weird warning that the `make_model_XXX` functions in the utils mod are unused, but they clearly are, and they work. [SO answer](https://stackoverflow.com/questions/42869134/compiler-warns-about-unused-code-for-private-functions-that-are-actually-used) was a bit overwhelming.. any word of advice? Thanks!

EDIT: warning disappeared, so that's good.


rust/benches/benches/models/serialize_models.rs
```diff
@@ -0,0 +1,48 @@
...
+mod utils;
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use std::time::Duration;
+
+use xaynet_core::mask::{IntoPrimitives, Model};
+
+fn serialize(model: &Model) -> Vec<i32> {
+    model
+        .clone()
```
little-dude, line 12: We don't want to clone were. Instead we can call `to_primitives()` which doesn't consume the model.

little-dude (changes requested):
This is nice :+1:  However the "serialize"/"de-serialize" terminology from the ticket is misleading. Serialization/de-serialization would be converting the model to a byte representation for storing it (which we'll also need benchmarks for). I think we can keep the "to/from primitives" terminology.

rust/benches/benches/models/serialize_models.rs
wilk10: ok thanks, will do

wilk10:
> This is nice 👍 However the "serialize"/"de-serialize" terminology from the ticket is misleading. Serialization/de-serialization would be converting the model to a byte representation for storing it (which we'll also need benchmarks for). I think we can keep the "to/from primitives" terminology.

ahah are you saying i implemented benchmarks for something else..?
so should the whole terminology be parse model to/from primitives? (in the ticket and in the code?)

rust/benches/benches/models/parse_models.rs
```diff
@@ -0,0 +1,44 @@
+#![allow(non_snake_case)]
+
+mod utils;
```
janpetschexain, line 3: it's idiomatic to not declare shared dependencies in every module of a directory, but instead declare them in a `mod.rs` and then `use` them in the modules of the directory. this has the benefit of a central place to look for shared dependencies. the directory structure could for example look like:
```
- models
    - mod.rs
    - parse_models.rs
    - serialize_models.rs
    - utils.rs
```
where the `mod.rs` then contains the following lines:
```rust
pub mod parse_models;
pub mod serialize_models;
mod utils;
```

rust/benches/benches/models/parse_models.rs
```diff
@@ -0,0 +1,44 @@
...
+pub use utils::*;
```
janpetschexain, line 4: `pub use` is a public re-export of the named modules, which is usually not what's needed in internal modules (but it has of course useful applications). i'd suggest to just `use utils;`.

as a side note, in the current way of declaring it this `pub use` is completely unused, because the methods are called as `utils::...` which is possible due to line 3 above: `mod utils;`. clippy just doesn't complain about unused functions because being unused is expected for public re-exports.

this is probably also the reason why you have seen the unused-functions-warning.

rust/benches/benches/models/utils.rs
```diff
@@ -0,0 +1,39 @@
+pub mod vector_or_model {
+    use xaynet_core::mask::{FromPrimitives, Model};
+
+    fn make_vector(bytes_size: usize) -> Vec<i32> {
+        // 1 i32 -> 4 bytes
+        let n_elements = bytes_size / 4;
+        let vector = vec![0_i32; n_elements];
+        vector
+    }
```
janpetschexain, line 9: it's idiomatic rust to just return the result without assigning it to a variable first, not sure why clippy didn't complain here because it usually does. for example like:
```rust
    fn make_vector(bytes_size: usize) -> Vec<i32> {
        // 1 i32 -> 4 bytes
        let n_elements = bytes_size / 4;
        vec![0_i32; n_elements]
    }
```

i would even go one step further and write this as the following, but this is of course personal taste:
```rust
    fn make_vector(bytes_size: usize) -> Vec<i32> {
        // 1 i32 -> 4 bytes
        vec![0_i32; bytes_size / 4]
    }
```

rust/benches/benches/models/parse_models.rs
```diff
@@ -0,0 +1,44 @@
...
+    c.bench_function("parse model from 1MB vector", |bench| {
+        bench.iter(|| Model::from_primitives(black_box(&mut vector_1MB.clone().into_iter())))
+    });
+}
+
+criterion_group!(
+    name = benches;
+    config = Criterion::default()
+        .sample_size(1000)
+        .measurement_time(Duration::new(10, 0));
```
janpetschexain, line 38: we might adjust the duration to sth like 60 or even 100 seconds due to how long proper benchmarking takes for those methods. criterion will complain about the time being to short anyways. same for the other benchmark.

janpetschexain (changes requested):
the benchmarks look good 👍 

>> This is nice 👍 However the "serialize"/"de-serialize" terminology from the ticket is misleading. Serialization/de-serialization would be converting the model to a byte representation for storing it (which we'll also need benchmarks for). I think we can keep the "to/from primitives" terminology.
>
> ahah are you saying i implemented benchmarks for something else..?
> so should the whole terminology be parse model to/from primitives? (in the ticket and in the code?)

i agree that we should use the `from/to_primitives` terminology here. to give you an idea why, the mental model for the conversions look roughly like the following:
```
              from/to_primitives
Vec<f32>   <---------------------->   Model
   ^                                    ^
   |                                    |
   | de/serialize                       | de/serialize
   |                                    |
   v                                    v
Vec<u8>                              Vec<u8>
```
this is because a `Model` holds a representation of the original primitive values, but as rational numbers of arbitrary precision (which is somewhat required for the masking). converting between a `Model` and a `Vec<f32>` is referred to as `from/to_primitives`. `de/serialization` on the other hand could be done for both representations to get a bunch of bytes to ship, but we usually only serialize the `Model`, but not the `Vec<f32>`. i hope this helps a bit to clear things up.

only partly related to your mr, i would suggest to give the benches a proper structure since it is going to grow further in the future. since this might be asking too much, i created a pr for your pr for a cleaned up directory structure, please have a look and feel free to merge this or make further adjustments: #591 (this also partly addresses some of my review comments)

rust/benches/benches/models/parse_models.rs
wilk10: sounds good and makes sense!

rust/benches/benches/models/parse_models.rs
wilk10: i took it from [here](https://github.com/rust-lang/rust/issues/46379#issuecomment-487556641). i stopped getting unused warnings when i added this line, but it's just a hack, i guess

i'll address the project structure in another reply

rust/benches/benches/models/utils.rs
wilk10: yeh i need to get used to this, i usually prefer 1 operation per line. totally happy to go the rust way in the first suggestion, for the second i guess it's case by case

rust/benches/benches/models/parse_models.rs
wilk10: sounds good! i just copied what Corentin did in his PR

wilk10:
> the benchmarks look good 👍
> 
> > > This is nice 👍 However the "serialize"/"de-serialize" terminology from the ticket is misleading. Serialization/de-serialization would be converting the model to a byte representation for storing it (which we'll also need benchmarks for). I think we can keep the "to/from primitives" terminology.
> > 
> > 
> > ahah are you saying i implemented benchmarks for something else..?
> > so should the whole terminology be parse model to/from primitives? (in the ticket and in the code?)
> 
> i agree that we should use the `from/to_primitives` terminology here. to give you an idea why, the mental model for the conversions look roughly like the following:
> 
> ```
>               from/to_primitives
> Vec<f32>   <---------------------->   Model
>    ^                                    ^
>    |                                    |
>    | de/serialize                       | de/serialize
>    |                                    |
>    v                                    v
> Vec<u8>                              Vec<u8>
> ```
> 
> this is because a `Model` holds a representation of the original primitive values, but as rational numbers of arbitrary precision (which is somewhat required for the masking). converting between a `Model` and a `Vec<f32>` is referred to as `from/to_primitives`. `de/serialization` on the other hand could be done for both representations to get a bunch of bytes to ship, but we usually only serialize the `Model`, but not the `Vec<f32>`. i hope this helps a bit to clear things up.
> 
> only partly related to your mr, i would suggest to give the benches a proper structure since it is going to grow further in the future. since this might be asking too much, i created a pr for your pr for a cleaned up directory structure, please have a look and feel free to merge this or make further adjustments: #591 (this also partly addresses some of my review comments)

Thanks a lot for the explanation! Much clearer now, maybe i should have asked for it before venturing into this PR. Agree with the naming, just need to find in the code these actual (de)serialisation ops, so that i can build the benchmarks for them.

There is a bit going on now with the change requests from you and Corentin, as well as your other extremely useful PR though.
With Corentin on a call last week we actually re-structured the crate in a different way than your own restructuring in [#591 ](https://github.com/xaynetwork/xaynet/pull/591) . so maybe we should coordinate a bit? i haven't committed those changes yet of course.
Shall we maybe pause these benchmarks until the second part of this sprint, and perhaps just briefly discuss a unified project structure and then going ahead with the implementation?
Not sure whether we should then close this PR or not, i'm open to anything, not sure if you peeps are fine with PRs open for a long time and what's the process there.

janpetschexain:
> Shall we maybe pause these benchmarks until the second part of this sprint, and perhaps just briefly discuss a unified project structure and then going ahead with the implementation?
> Not sure whether we should then close this PR or not, i'm open to anything, not sure if you peeps are fine with PRs open for a long time and what's the process there.

makes sense to align on a common structure, just leave the pr open no worries.

janpetschexain:
#566 has to be merged first, then this must be rebased onto master to avoid having code from the other pr.

rust/benches/benches/models/serialize_models.rs
janpetschexain: changed.

rust/benches/benches/models/utils.rs
```diff
@@ -0,0 +1,39 @@
+pub mod vector_or_model {
```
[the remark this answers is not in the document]
janpetschexain: cleaned up in a more isolated way.

rust/benches/benches/models/parse_models.rs
janpetschexain: cleaned up in a more isolated way.

rust/benches/benches/models/parse_models.rs
janpetschexain: changed.

wilk10 (commented):
@janpetschexain Thanks, LGTM, though i hope git doesn't make a mess of the history, with multiple authors and such. it's also fine if you want to close this and cherry-pick the stuff that should be merged on top of #566 in a clean PR authored only by you

janpetschexain:
> @janpetschexain Thanks, LGTM, though i hope git doesn't make a mess of the history, with multiple authors and such. it's also fine if you want to close this and cherry-pick the stuff that should be merged on top of #566 in a clean PR authored only by you

multiple authors shouldn't be an issue. regarding the commit history, i prefer to squash merge anyways.

janpetschexain (approved):

PR #583, opened 2020-10-28
XN-1209 A few improvements

abargiela:
- Added 
    In the action: `Notify on Slack` 
    added `if: ${{ success() }}` 
    So if everything runs with success, slack will be notified.

    Action: jungwinter/split@v1

- Changed
    Changed the the trigger command:
        From: /build_pr 
        To: /deploy

- Removed
    Removed `dockerhub-pr.yml` 
    Replaced by: `dockerhub-pr-with-parameters.yml`


Robert-Steiner (approved):
Looks good👍

PR #584, opened 2020-10-28
XN-1209 Add action docker/setup-buildx-action@v1 to fix: buildx call

abargiela:
Add action: `docker/setup-buildx-action@v1` 
 
To fix: 
```
auto-push is currently not implemented for docker driver
Error: buildx call failed with: auto-push is currently not implemented for docker driver
```

### DIFF
--- a/.github/workflows/dockerhub-pr-with-parameters.yml
+++ b/.github/workflows/dockerhub-pr-with-parameters.yml
@@ -36,9 +36,6 @@ jobs:
         with:
           ref: ${{ steps.comment-branch.outputs.head_ref }}
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-pr-with-parameters.yml
```diff
@@ -36,6 +36,9 @@ jobs:
         with:
           ref: ${{ steps.comment-branch.outputs.head_ref }}
 
+      - name: Set up Docker Buildx
```
Robert-Steiner, line 39: can you please move it between the steps:
`- name: Login to DockerHub` and 
`- name: build-tag-push` 
so that they are  grouped together

abargiela: done

Robert-Steiner (approved):

PR #588, opened 2020-10-29
call the multipart handler in xaynet-server

little-dude:
We had implemented the service for handling multipart messages, but never actually called it.

### DIFF
--- a/rust/xaynet-server/src/services/messages/mod.rs
+++ b/rust/xaynet-server/src/services/messages/mod.rs
@@ -12,6 +12,7 @@ pub use self::error::ServiceError;
 use self::{
     decryptor::Decryptor,
     message_parser::MessageParser,
+    multipart::MultipartHandler,
     state_machine::StateMachine,
     task_validator::TaskValidator,
 };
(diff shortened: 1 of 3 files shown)


rust/xaynet-server/src/services/messages/mod.rs
```diff
@@ -55,6 +58,17 @@ impl PetMessageHandler {
         self.message_parser.call(data).await
     }
 
+    async fn handle_multipart(
+        &mut self,
+        message: Message,
+    ) -> Result<Option<Message>, ServiceError> {
+        poll_fn(|cx| {
```
Robert-Steiner, line 65: Is it possible to call this function as with the other functions?

like:
```rust
poll_fn(|cx| self.multipart_handler.poll_ready(cx)).await?;
```

Robert-Steiner (approved):
Looks good to me 👍
added only one question

rust/xaynet-server/src/services/messages/mod.rs
little-dude: Oh actually yes, thank you!

PR #589, opened 2020-10-29
randomized scalar tests

finiteprods:
**Summary.**

This PR adds to the existing suite of masking / aggregation tests, to more thoroughly test the generalised scalar (e.g. there had been no test for scalars taking arbitrary values > 1).

More specifically,

- a new suite of tests `test_masking_scalar!` (to complement `test_masking!`) which samples a random scalar. 
- a new suite of tests `test_masking_and_aggregation_scalar!` (to complement `test_masking_and_aggregation!`) which samples random scalars.
- `test_aggregation!` is extended slightly, to additionally sample a random mask unit. `test_masking!` and `test_masking_and_aggregation!` are largely unchanged.

**Explanation of the sense in which the new tests _complement_ the existing.** 

The structure of `test_masking_scalar!` largely follows the existing `test_masking!`. The main difference is that while the latter generates a random model with a fixed scalar, `test_masking_scalar!` generates a _random scalar_ with a _fixed model_.

Similar story with `test_masking_and_aggregation_scalar!`. Rather than generating `n` random models each with a fixed scalar `1/n` (a canonical choice under the older, more restrictive requirements on scalars), instead it generates `n` random scalars (of arbitrary positive value, under the more relaxed requirements of generalised scalars) each with the same fixed model.

While it would be possible to randomize both scalars and models within a single combined test, this would complicate the test, in part due to current limitations with our masking configurations:

- assuming scalars and scaled models follow the _same_ masking config, we need to be careful that the sampled weights, after scaling, don't blow the bounds of the masking config.
- the same applies when the masking configs for both are distinct, except more complicated: some pairs of masking configs likely won't make sense.

For these reasons, it is simpler to randomize _either_ one of scalars or models, so that we are free to fix a safe choice of value for the other.

### DIFF
--- a/rust/xaynet-core/src/mask/masking.rs
+++ b/rust/xaynet-core/src/mask/masking.rs
@@ -588,6 +588,20 @@ mod tests {
     test_masking!(pow_i64_b6, Power2, i64, 1_000_000, 10);
     test_masking!(pow_i64_bmax, Power2, i64, 10);
 
+    /// Generate tests for masking and unmasking of a single model:
+    /// - generate random scalar from a uniform distribution with a seeded PRNG
+    /// - scale a model of unit weights and mask it
+    /// - check that all masked weights belong to the chosen finite group
+    /// - unmask the masked model
+    /// - check that all unmasked weights are equal to the original weights (up to a tolerance
+    ///   determined by the masking configuration)
+    ///
+    /// The arguments to the macro are:
+    /// - a suffix for the test name
+    /// - the group type of the model (variants of `GroupType`)
+    /// - the data type of the model (either primitives or variants of `DataType`)
+    /// - an absolute bound for the weights (optional, choices: 1, 100, 10_000, 1_000_000)
+    /// - the number of weights
     macro_rules! test_masking_scalar {
         ($suffix:ident, $group:ty, $data:ty, $bound:expr, $len:expr $(,)?) => {
             paste::item! {
@@ -1036,6 +1050,21 @@ mod tests {
     test_masking_and_aggregation!(pow_i64_b6, Power2, i64, 1_000_000, 10, 5);
     test_masking_and_aggregation!(pow_i64_bmax, Power2, i64, 10, 5);
 
+    /// Generate tests for masking, aggregation and unmasking of multiple models:
+    /// - generate random scalars from a uniform distribution with a seeded PRNG
+    /// - scale a model of unit weights, mask and aggregate it to the aggregated masked models
+    /// - derive a mask from the mask seed and aggregate it to the aggregated masks
+    /// - unmask the aggregated masked model
+    /// - check that all aggregated unmasked weights are equal to the original unit weights (up
+    ///   to a tolerance determined by the masking configuration)
+    ///
+    /// The arguments to the macro are:
+    /// - a suffix for the test name
+    /// - the group type of the model (variants of `GroupType`)
+    /// - the data type of the model (either primitives or variants of `DataType`)
+    /// - an absolute bound for the weights (optional, choices: 1, 100, 10_000, 1_000_000)
+    /// - the number of weights per model
+    /// - the number of models
     macro_rules! test_masking_and_aggregation_scalar {
         ($suffix:ident, $group:ty, $data:ty, $bound:expr, $len:expr, $count:expr $(,)?) => {
             paste::item! {
--- a/rust/xaynet-core/src/mask/masking.rs
+++ b/rust/xaynet-core/src/mask/masking.rs
@@ -1030,4 +1030,156 @@ mod tests {
     test_masking_and_aggregation!(pow_i64_b4, Power2, i64, 10_000, 10, 5);
     test_masking_and_aggregation!(pow_i64_b6, Power2, i64, 1_000_000, 10, 5);
     test_masking_and_aggregation!(pow_i64_bmax, Power2, i64, 10, 5);
+
+    macro_rules! test_masking_and_aggregation_scalar {
+        ($suffix:ident, $group:ty, $data:ty, $bound:expr, $len:expr, $count:expr $(,)?) => {
+            paste::item! {
+                #[test]
+                fn [<test_masking_and_aggregation_scalar $suffix>]() {
+                    // Step 1: Build the masking config
+                    let config = MaskConfig {
+                        group_type: $group,
+                        data_type: paste::expr! { [<$data:upper>] },
+                        bound_type: match $bound {
+                            1 => B0,
+                            100 => B2,
+                            10_000 => B4,
+                            1_000_000 => B6,
+                            _ => Bmax,
+                        },
+                        model_type: M3,
+                    };
+                    let model_size = $len as usize;
+
+                    // Step 2: Generate random scalars
+                    let bound = if $bound == 0 {
+                        paste::expr! { [<$data:lower>]::MAX / (2 as [<$data:lower>]) }
+                    } else {
+                        paste::expr! { $bound as [<$data:lower>] }
+                    };
+                    let zero = 0 as [<$data:lower>];
+                    let mut prng = ChaCha20Rng::from_seed(MaskSeed::generate().as_array());
+                    let mut scalars = iter::repeat_with(move || {
+                        let rand_neg = Uniform::new(-bound, zero).sample(&mut prng);
+                        rand_neg.abs() as f64
+                    });
+                    let mut models =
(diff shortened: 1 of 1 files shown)


rust/xaynet-core/src/mask/masking.rs
```diff
@@ -585,6 +587,155 @@ mod tests {
...
+                    // take vector [1, ..., 1] as the model to scale
+                    let bound = if $bound == 0 {
+                        paste::expr! { [<$data:lower>]::MAX / (2 as [<$data:lower>]) }
+                    } else {
+                        paste::expr! { $bound as [<$data:lower>] }
+                    };
+                    let zero = 0 as [<$data:lower>];
+                    let mut prng = ChaCha20Rng::from_seed(MaskSeed::generate().as_array());
+                    let rand_neg = Uniform::new(-bound, zero).sample(&mut prng);
+                    let scalar = rand_neg.abs() as f64;
```
janpetschexain, line 634: to avoid the double negation of the scalar and to still sample it from the correct interval, this could also be written as
```suggestion
                    let eps = [<$data:lower>]::EPSILON;
                    let mut prng = ChaCha20Rng::from_seed(MaskSeed::generate().as_array());
                    let scalar = Uniform::new_inclusive(eps, bound).sample(&mut prng) as f64;
```

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -585,6 +587,155 @@ mod tests {
...
+    /// - check that all masked weights belong to the chosen finite group
+    /// - unmask the masked model
+    /// - check that all unmasked weights are equal to the original weights (up to a tolerance
+    ///   determined by the masking configuration)
+    ///
+    /// The arguments to the macro are:
+    /// - a suffix for the test name
+    /// - the group type of the model (variants of `GroupType`)
+    /// - the data type of the model (either primitives or variants of `DataType`)
+    /// - an absolute bound for the weights (optional, choices: 1, 100, 10_000, 1_000_000)
```
janpetschexain, line 602: we actually need two data types here, one for the model as usual and one for the scalar, because in our masking setup the scalar must always be a floating point value (which would not hold for integer models). since the model is set to constant 1 for this test, we down gain anything from having int and float data types for the model and we might simplify a bit and only call this macro with float data types, eliminating the need to have two different data types.

since we currently go for the same mask configuration for both model and scalar, we probably should take advantage of this simplification and extend the test later on.

```suggestion
    /// The arguments to the macro are:
    /// - a suffix for the test name
    /// - the group type of the model and scalar (variants of `GroupType`)
    /// - the data type of the model and sclar (either float primitives or float variants of `DataType`)
    /// - an absolute bound for the scalar (optional, choices: 1, 100, 10_000, 1_000_000)
```

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -896,4 +1048,171 @@ mod tests {
...
+                        paste::expr! { [<$data:lower>]::MAX / (2 as [<$data:lower>]) }
+                    } else {
+                        paste::expr! { $bound as [<$data:lower>] }
+                    };
+                    let zero = 0 as [<$data:lower>];
+                    let mut prng = ChaCha20Rng::from_seed(MaskSeed::generate().as_array());
+                    let mut scalars = iter::repeat_with(move || {
+                        let rand_neg = Uniform::new(-bound, zero).sample(&mut prng);
+                        rand_neg.abs() as f64
+                    });
```
janpetschexain, line 1099: similarly as above:
```suggestion
                    let eps = [<$data:lower>]::EPSILON;
                    let mut prng = ChaCha20Rng::from_seed(MaskSeed::generate().as_array());
                    let mut scalars = iter::repeat_with(move || Uniform::new_inclusive(eps, bound).sample(&mut prng) as f64);
```

janpetschexain (changes requested):
looks good, only some docs need to be updated.

i'd suggest to remove the `..._scalar!` tests for `i32` and `i64` as explained in the comments and add them again once we fully support different mask configurations for the model and scalar.

rust/xaynet-core/src/mask/masking.rs
finiteprods: not sure I follow, are you saying that the `i32`/`i64` cases are redundant here?

janpetschexain: what i mean is that in the current implementation the values for the scalar are drawn from the same data type as the model. in case of `i32`/`i64` the random scalars would only take integer values, which doesn't make much sense regarding the definition and purpose of the scalar.

finiteprods: I see, so I think you're saying that in practice, one wouldn't actually take scalars to be int values. that may well be the case, but there is no harm as such in allowing for this, is there?

janpetschexain: there is probably no harm, it just doesn't make much sense to test for sth that isn't intended in the first place.

rust/xaynet-core/src/mask/masking.rs
finiteprods: nice suggestion :+1:

rust/xaynet-core/src/mask/masking.rs
finiteprods: fair enough. we can consider bringing them back then once we can properly mix & match mask configurations.

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -482,19 +483,20 @@ mod tests {
                     let mut prng = ChaCha20Rng::from_seed(MaskSeed::generate().as_array());
```
little-dude, line 483: I think it would be nice for all the tests that use randomness to:

1. print out the seed they are using
2. have a way to manually set the seed for debugging

That would make test failures reproducible, which is always a challenge when tests use random inputs.

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -897,4 +1013,135 @@ mod tests {
...
+                            1_000_000 => B6,
+                            _ => Bmax,
+                        },
+                        model_type: M3,
+                    };
+                    let vect_len = $len as usize;
+
+                    // Step 2: Generate random scalars
+                    // take vectors [1, ..., 1] as models to scale
+                    let bound = if $bound == 0 {
```
little-dude, line 1055: It seems that `$bound` is never 0, so is this really needed?

rust/xaynet-core/src/mask/masking.rs
```diff
@@ -897,4 +1013,135 @@ mod tests {
...
+    ///
+    /// The arguments to the macro are:
+    /// - a suffix for the test name
+    /// - the group type of the model and scalar (variants of `GroupType`)
+    /// - the data type of the model and scalar (either float primitives or float variants of
+    ///   `DataType`)
+    /// - an absolute bound for the scalar (optional, choices: 1, 100, 10_000, 1_000_000)
+    /// - the number of weights per model
+    /// - the number of models
+    macro_rules! test_masking_and_aggregation_scalar {
```
little-dude, line 1033: The macro is quite long, and even though the comments are very helpful, perhaps it would be worth breaking it down into smaller pieces, either with other macros, or with functions? Each step could be its own macro/function for example.

rust/xaynet-core/src/mask/masking.rs
janpetschexain: it is used for the `Bmax` cases in the 2nd branch of the macro.

rust/xaynet-core/src/mask/masking.rs
little-dude: Although these are tests, so maybe it's not worth the effort.

little-dude (commented):
As far as I can tell with my limited understanding of the maths behind all this, this is good code-wise. I'll trust Jan for reviewing the nitty gritty mathematical subtleties. The only thing I think would be nice to add is a way to control the input by specifying a seed. Otherwise, it's going to be very difficult to debug spurious failures.

janpetschexain (approved):

rust/xaynet-core/src/mask/masking.rs
finiteprods: nice suggestions, I agree it would be nice to have this capability for reproducibility reasons. we can think about tackling this in a separate issue.

rust/xaynet-core/src/mask/masking.rs
finiteprods: they're tests after all, but there is enough repetition in the masking tests to justify some refactor-tidying, I agree.

PR #590, opened 2020-10-29
XN-1272 tls client authentication

janpetschexain:
**References**
- [XN-1272](https://xainag.atlassian.net/browse/XN-1272)

**Summary**
this builds on the additions for tls client authentication with warp, which is [reviewable here](https://github.com/xaynetwork/warp/pull/1). ~i assume that this will cause quite some merge conflicts with #578 and i guess it makes sense that the client refactoring is merged first and the the client authentication is rebased afterwards~ (edit: rebase done). also this might seem quite large initially, and it must somewhat be because it touches server and client, but without the autogenerated parts it's only roughly half the size.

on the server side:
- add settings to read trusted anchors to verify client identities (i.e. certificates) and update the configs (yet commented out because the respective roots are not yet on k8)
- configure and run the warp server with tls client authentication
- update the binaries

on the client side:
- add settings to read client identities
- update the http ~and mobile~ (edit: cut out due to client refactoring) client to run with tls client authentication
- update the examples
- ~update the ffi to run with tls client authentication and regenerate the C files (but no need to review the `.h` file since it's auto generated)~ (edit: still in there, but outdated and disabled due to client refactoring)


rust/xaynet-server/src/rest.rs
```diff
@@ -22,10 +23,17 @@ use xaynet_core::{crypto::ByteObject, ParticipantPublicKey};
...
 /// * `pet_message_handler`: handler for responding to PET messages.
-pub async fn serve<F>(api_settings: ApiSettings, fetcher: F, pet_message_handler: PetMessageHandler)
+///
+/// # Errors
+/// Fails if invalid TLS settings are provided when the `tls` feature is enabled.
+pub async fn serve<F>(
+    api_settings: ApiSettings,
+    fetcher: F,
+    pet_message_handler: PetMessageHandler,
+) -> Option<()>
```
finiteprods, line 36: was going to say, wouldn't a `bool` return type do just as well? but perhaps `Option` is more convenient with the `?` syntax ...

rust/xaynet-server/src/settings/mod.rs
```diff
@@ -384,9 +388,9 @@ pub struct ApiSettings {
 #[cfg(feature = "tls")]
 /// Checks API settings.
 fn validate_api(s: &ApiSettings) -> Result<(), ValidationError> {
-    match (&s.tls_certificate, &s.tls_key) {
-        (Some(_), Some(_)) | (None, None) => Ok(()),
-        _ => Err(ValidationError::new("invalid server authentication paths")),
+    match (&s.tls_certificate, &s.tls_key, &s.tls_root) {
+        (Some(_), Some(_), Some(_)) | (Some(_), Some(_), None) | (None, None, Some(_)) => Ok(()),
```
finiteprods, line 392: possible simplification:
```rust
(Some(_), Some(_), _) | (None, None, Some(_)) => Ok(()),
```

finiteprods (commented):
looks very good, a couple of minor comments.

rust/xaynet-server/src/rest.rs
janpetschexain: that's indeed just for `?` error handling from within the `run_...` methods. later on we should probably replace the `Option` with a `Result` and add some proper error handling to the `serve` method.

janpetschexain: i introduced an error type right away during the rebase.

janpetschexain:
i rebased on the client refactorings. now we just need to merge the [warp fork pr](https://github.com/xaynetwork/warp/pull/1) first and then adjust the import here before it can get merged.

finiteprods (approved):

PR #591, opened 2020-10-30
clean up directory structure

janpetschexain:
directory clean up for the benches related to #582


wilk10 (commented):
Thanks a lot! much cleaner now. my only issue is that this project structure is different than the one i have discussed with Corentin on a call last week (see [here](https://github.com/xaynetwork/xaynet/pull/582#issuecomment-720283769) for the same info).
My suggestion is to pause with the benchmarks until the second part of this sprint, get on the same page about the project structure and then I / we can go on with the implementation. Does it make sense?

rust/benches/Cargo.toml
```diff
@@ -21,15 +21,6 @@ criterion = "0.3.3"
...
-[[bench]]
-name = "parse_models"
-path = "benches/models/parse_models.rs"
-harness = false
-
-[[bench]]
-name = "serialize_models"
-path = "benches/models/serialize_models.rs"
+name = "benches"
+path = "main.rs"
```
little-dude, line 25: If we do this we can we still build/run a specific benchmark? Because currently we can do `cargo bench --bench xxx` and `cargo bench --build xxx`

little-dude:
Our idea for the benches was to create an actual crate for the utilities:

```
.
├── benches
│  ├── messages.rs
│  ├── models.rs
│  └── masking.rs
├── Cargo.toml
└── src // Not sure exactly how we'd structure this, this is just one idea where we mimic the structure of the benchmarks
   ├── messages.rs // utilities for the messages benchmarks
   ├── models.rs   // utilities for the models bencmarks
   ├── masking.rs
   ├── common.rs   // stuff that is common to most benchmarks
   └── lib.rs
```

The advantage of using a crate:

- we can just run `cargo check` when writing the helpers, which is faster than compiling benchmarks, because they are compiled in release mode
- makes it easy to share helpers (helpers from `crate::messages` can `use crate::models` for example). This is possible in benchmarks directly but:
	- using other modules increases compile time
  	- we can have only one entry point (the `main.rs` in this PR), which is less flexible

rust/benches/Cargo.toml
janpetschexain: this way they can only be run all together. if we want to run them individually, we have to keep the criterion main and the toml tables separated.

janpetschexain:
> Our idea for the benches was to create an actual crate for the utilities:
> 
> ```
> .
> ├── benches
> │  ├── messages.rs
> │  ├── models.rs
> │  └── masking.rs
> ├── Cargo.toml
> └── src // Not sure exactly how we'd structure this, this is just one idea where we mimic the structure of the benchmarks
>    ├── messages.rs // utilities for the messages benchmarks
>    ├── models.rs   // utilities for the models bencmarks
>    ├── masking.rs
>    ├── common.rs   // stuff that is common to most benchmarks
>    └── lib.rs
> ```
> 
> The advantage of using a crate:
> 
> * we can just run `cargo check` when writing the helpers, which is faster than compiling benchmarks, because they are compiled in release mode
> * makes it easy to share helpers (helpers from `crate::messages` can `use crate::models` for example). This is possible in benchmarks directly but:
>   
>   * using other modules increases compile time
>   * we can have only one entry point (the `main.rs` in this PR), which is less flexible

makes also sense, i don't have a strong opinion about the final structure of the benches as long as they are more structured than now :)

rust/benches/Cargo.toml
little-dude: I'd prefer to be able to run benchmarks individually because they are quite long. Plus when making changes to messages for instance, it's not necessary to run masking or models benchmarks

janpetschexain: sounds reasonable. how should we proceed with the whole topic? it probably doesn't make much sense to merge this as it is. should the structuring be part of #582 and who takes that? @wilk10 do you feel comfortable to attempt the restructuring yourself with input and support from @little-dude and myself?

wilk10: yes! @little-dude already helped me with making into a crate at the end of last sprint, so i can use that as a base and adapt it to the proposed structure, now that i have some time after the analytics proof of concept

PR #592, opened 2020-10-30
XN-1163 replace macro use

janpetschexain:
**References**
- [XN-1163](https://xainag.atlassian.net/browse/XN-1163)

**Summary**
- replaces the 2015 edition's `#[macro_use]` imports with the 2018 edition's `use` import, see the [rust edition guide](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html)
- commits are grouped by workspace to help readability
- cleans up some imports

one thing i noted while working on this is that the `metrics!` proc macro has the `#[cfg(feature = "metrics")]` flag hidden inside of its definition in the `xaynet_macros` crate. this has some implications:
- `metrics!` will do nothing when the flag is not enabled, which is not clear from the calling site
- `metrics!` must be imported when called whether or not the flag is enabled

i'd suggest to remove the flag from within the proc macro and put it at the caller site like usual.
```rust
#[cfg(feature = "metrics")]
use xaynet_macros::metrics;

// function body
    #[cfg(feature = "metrics")]
    metrics!(...);
```
should we do this in a separate pr or should we include it in this one since the changes are quickly done, wdyt?

**Todo**
- [x] rebase on client refactoring
- [x] decided on metrics proc macro: no change for now

### DIFF
--- a/rust/examples/Cargo.toml
+++ b/rust/examples/Cargo.toml
@@ -25,7 +25,7 @@ tracing = "0.1.19"
 tracing-futures = "0.2.4"
 tracing-subscriber = "0.2.12"
 xaynet-core = { path = "../xaynet-core" }
-xaynet-sdk = { path = "../xaynet-sdk", features = ["reqwest_client"] }
+xaynet-sdk = { path = "../xaynet-sdk", features = ["reqwest-client"] }
 
 [[example]]
 name = "test-drive"
--- a/rust/examples/test-drive/main.rs
+++ b/rust/examples/test-drive/main.rs
@@ -1,15 +1,10 @@
-#[macro_use]
-extern crate tracing;
-
-#[macro_use]
-extern crate async_trait;
-
 use std::{path::PathBuf, sync::Arc, time::Duration};
 
 use reqwest::Certificate;
 use structopt::StructOpt;
+use tracing::error_span;
 use tracing_futures::Instrument;
-use tracing_subscriber::*;
+use tracing_subscriber::{EnvFilter, FmtSubscriber};
 
 use xaynet_core::{
     crypto::SigningKeyPair,
--- a/rust/examples/test-drive/participant.rs
+++ b/rust/examples/test-drive/participant.rs
@@ -1,6 +1,9 @@
 use std::{sync::Arc, time::Duration};
 
+use async_trait::async_trait;
 use tokio::{sync::mpsc, time::delay_for};
+use tracing::{info, warn};
+
 use xaynet_core::mask::Model;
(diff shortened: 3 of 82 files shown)


little-dude:
I don't really understand the motivation for switching to explicit imports here. I like that we can just slap a `#[macro_use] extern crate tracing;` and have the macros everywhere. Is there a downside to just doing this?

I like the re-ordering of the use statements though :+1:

janpetschexain:
> I don't really understand the motivation for switching to explicit imports here. I like that we can just slap a `#[macro_use] extern crate tracing;` and have the macros everywhere. Is there a downside to just doing this?
> 
> I like the re-ordering of the use statements though 👍

there are three reasons for switching.

one is that the `#[macro_use]` is already discouraged from the 2018 version onwards (https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html), but it is not yet deprecated because it is still needed for crate-internal macros. the progress on declarative macros is already quite far (https://github.com/rust-lang/rust/issues/39412) and it is quite likely that the `#[macro_use]` will become deprecated then and we would have to switch anyways, so we can already set it up on the new way.

the other is that it caused some issues with rust-analyzer (at least on my machine^TM) which should be gone with the usual `use ...` declaration. it seems to me that rust-analyzer is not good with macros in general.

the last reason is quite subjective tbh, i really like rust's explicit style, not much happens magically under the hood if you don't call for it. using `#[macro_use]` for example somewhere in `lib.rs` to provide access to foreign macros in some submodule of the crate is quite counter-intuitive to that paradigm. with `use ...` at the appropriate places we have full knowledge what is used where.

little-dude:
Fair enough, these are good reasons. :+1: from me then.

> i'd suggest to remove the flag from within the proc macro and put it at the caller site like usual.

That sounds good to me but I'll let @Robert-Steiner comment because I think he did that on purpose.

Robert-Steiner:
> Fair enough, these are good reasons. 👍 from me then.
> 
> > i'd suggest to remove the flag from within the proc macro and put it at the caller site like usual.
> 
> That sounds good to me but I'll let @Robert-Steiner comment because I think he did that on purpose.

Right, the idea was to hide the feature flag so that the code stays easy to read. Not sure what the best solution is. 
In the long run, I would like to replace the macro with [metrics](https://github.com/metrics-rs/metrics). It's basically like `log` just for metrics. With this we can remove the sender part from the state machine and easliy collect metrics on any part of the code we want. But would like to wait for the fist non-alpha release

Robert-Steiner (approved):
Looks good to me👍

PR #594, opened 2020-11-05
Impl storage traits

Robert-Steiner:
**Summary**
- added a `CoordinatorStorage`and a `ModelStorage` trait 
The reason why I chose these names instead of `VolatileStorage` and `PersistentStorage` is that the coordinator state also has values ​​that have to be persistence like the `round_id`. 
- refactored pet wrappers
- added a generic store

### DIFF
--- a/rust/xaynet-server/src/state_machine/mod.rs
+++ b/rust/xaynet-server/src/state_machine/mod.rs
@@ -126,7 +126,7 @@ use crate::metrics::MetricsSender;
 use crate::{settings::RestoreSettings, storage::s3};
 use crate::{
     settings::{MaskSettings, ModelSettings, PetSettings},
-    storage::{redis, MaskDictIncrError, RedisError, SeedDictUpdateError, SumDictAddError},
+    storage::{redis, LocalSeedDictAddError, MaskScoreIncrError, RedisError, SumPartAddError},
 };
 #[cfg(feature = "model-persistence")]
 use xaynet_core::mask::Model;
@@ -135,26 +135,33 @@ use xaynet_core::mask::UnmaskingError;
 /// Error returned when the state machine fails to handle a request
 #[derive(Debug, Error)]
 pub enum RequestError {
+    /// the message was rejected
     #[error("the message was rejected")]
     MessageRejected,
 
+    /// the model or scalar sent by the participant could not be aggregated
     #[error("invalid update: the model or scalar sent by the participant could not be aggregated")]
     AggregationFailed,
 
+    /// the request could not be processed due to an internal error
     #[error("the request could not be processed due to an internal error: {0}")]
     InternalError(&'static str),
 
+    /// a redis request failed
     #[error("redis request failed: {0}")]
     Redis(#[from] RedisError),
 
+    /// adding a local seed dict to the seed dictionary failed
     #[error(transparent)]
-    SeedDictUpdate(#[from] SeedDictUpdateError),
+    LocalSeedDictAdd(#[from] LocalSeedDictAddError),
 
+    /// adding a sum participant to the sum dictionary failed
     #[error(transparent)]
-    SumDictAdd(#[from] SumDictAddError),
+    SumPartAdd(#[from] SumPartAddError),
 
+    /// incrementing a mask score failed
     #[error(transparent)]
-    MaskDictIncr(#[from] MaskDictIncrError),
+    MaskScoreIncr(#[from] MaskScoreIncrError),
 }
 
 pub type StateMachineResult = Result<(), RequestError>;
--- a/rust/xaynet-server/src/storage/api.rs
+++ b/rust/xaynet-server/src/storage/api.rs
@@ -189,3 +189,67 @@ where
     }
 }
 
+/// A wrapper that contains the result of the "add sum participant" operation.
+#[derive(Deref)]
+pub struct SumPartAdd(pub(crate) Result<(), SumPartAddError>);
+
+impl SumPartAdd {
+    /// Unwraps this wrapper, returning the underlying result.
+    pub fn into_inner(self) -> Result<(), SumPartAddError> {
+        self.0
+    }
+}
+
+/// Error that can occur when adding a sum participant to the [`SumDict`].
+#[derive(Display, Error, Debug, TryFromPrimitive)]
+#[repr(i64)]
+pub enum SumPartAddError {
+    /// sum participant already exists
+    AlreadyExists = 0,
+}
+
+/// A wrapper that contains the result of the "add local seed dict" operation.
+#[derive(Deref)]
+pub struct LocalSeedDictAdd(pub(crate) Result<(), LocalSeedDictAddError>);
+
+impl LocalSeedDictAdd {
+    /// Unwraps this wrapper, returning the underlying result.
+    pub fn into_inner(self) -> Result<(), LocalSeedDictAddError> {
+        self.0
+    }
+}
+
+/// Error that can occur when adding a local seed dict to the [`SeedDict`].
+#[derive(Display, Error, Debug, TryFromPrimitive)]
+#[repr(i64)]
+pub enum LocalSeedDictAddError {
+    /// the length of the local seed dict and the length of sum dict are not equal
+    LengthMisMatch = -1,
+    /// local dict contains an unknown sum participant
+    UnknownSumParticipant = -2,
+    /// update participant already submitted an update
+    UpdatePkAlreadySubmitted = -3,
+    /// update participant already exists in the inner update seed dict
+    UpdatePkAlreadyExistsInUpdateSeedDict = -4,
+}
+
+/// A wrapper that contains the result of the "increment mask score" operation.
+#[derive(Deref)]
+pub struct MaskScoreIncr(pub(crate) Result<(), MaskScoreIncrError>);
+
+impl MaskScoreIncr {
+    /// Unwraps this wrapper, returning the underlying result.
+    pub fn into_inner(self) -> Result<(), MaskScoreIncrError> {
+        self.0
+    }
+}
+
+/// Error that can occur when incrementing a mask score.
+#[derive(Display, Error, Debug, TryFromPrimitive)]
+#[repr(i64)]
+pub enum MaskScoreIncrError {
+    /// unknown sum participant
+    UnknownSumPk = -1,
+    /// sum participant submitted a mask already
+    MaskAlreadySubmitted = -2,
+}
--- a/rust/xaynet-server/src/storage/api.rs
+++ b/rust/xaynet-server/src/storage/api.rs
@@ -1,0 +1,191 @@
+//! Storage API.
+
+use async_trait::async_trait;
+use derive_more::Deref;
+use displaydoc::Display;
+use num_enum::TryFromPrimitive;
+use thiserror::Error;
+
+use crate::state_machine::coordinator::CoordinatorState;
+use xaynet_core::{
+    common::RoundSeed,
+    crypto::ByteObject,
+    mask::{MaskObject, Model},
+    LocalSeedDict,
+    SeedDict,
+    SumDict,
+    SumParticipantEphemeralPublicKey,
+    SumParticipantPublicKey,
+    UpdateParticipantPublicKey,
+};
+
+/// The error type for storage operations that are not directly related to application domain.
+/// These include, for example IO errors like broken pipe, file not found, out-of-memory, etc.
+pub type StorageError = anyhow::Error;
+
+/// The result of the storage operation.
+pub type StorageResult<T> = Result<T, StorageError>;
+
+#[async_trait]
+/// An abstract coordinator storage.
+pub trait CoordinatorStorage
+where
+    Self: Clone + Send + Sync + 'static,
+{
+    /// Sets a [`CoordinatorState`].
+    ///
+    /// # Behavior
+    ///
+    /// - If no state has been set yet, set the state and return `StorageResult::Ok(())`.
+    /// - If a state already exists, override the state and return `StorageResult::Ok(())`.
+    async fn set_coordinator_state(&mut self, state: &CoordinatorState) -> StorageResult<()>;
+
+    /// Returns a [`CoordinatorState`].
+    ///
+    /// # Behavior
+    ///
+    /// - If no state has been set yet, return `StorageResult::Ok(Option::None)`.
+    /// - If a state exists, return `StorageResult::Ok(Some(CoordinatorState))`.
+    async fn coordinator_state(&mut self) -> StorageResult<Option<CoordinatorState>>;
+
+    /// Adds a sum participant entry to the [`SumDict`].
+    ///
+    /// # Behavior
+    ///
+    /// - If a sum participant has been successfully added, return `StorageResult::Ok(SumPartAdd)`
+    ///   containing a `Result::Ok(())`.
+    /// - If the participant could not be added due to a PET protocol error, return
+    ///   the corresponding `StorageResult::Ok(SumPartAdd)` containing a
+    ///   `Result::Err(SumPartAddError)`.
+    async fn add_sum_participant(
+        &mut self,
+        pk: &SumParticipantPublicKey,
+        ephm_pk: &SumParticipantEphemeralPublicKey,
+    ) -> StorageResult<SumPartAdd>;
+
+    /// Returns the [`SumDict`].
+    ///
+    /// # Behavior
+    ///
+    /// - If the sum dict does not exist, return `StorageResult::Ok(Option::None)`.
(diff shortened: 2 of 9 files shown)


rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Gets the [`SeedDict`].
+    ///
+    /// # Behavior
+    ///
+    /// - If the seed dict does not exist, return `Result::Ok(Option::None)`.
+    /// - If the seed dict exists, return `Result::Ok(Option::Some(SeedDict))`.
+    async fn get_seed_dict(&mut self) -> StorageResult<Option<SeedDict>>;
+
+    /// Increments the mask score with the given [`MaskObject`].
+    /// The score of the given mask is incremented by `1`.
```
janpetschexain, line 91:
```suggestion
    /// Increments the mask score of the given [`MaskObject`] by one.
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    /// # Behavior
+    ///
+    /// - If the mask score has been successfully incremented, return `Result::Ok(Vec<(MaskObject, u64)>)`.
+    /// - If the mask score could not be incremented due to a PET protocol error, return
+    ///   the corresponding `Result::Ok(MaskScoreIncr(Result::Err(MaskScoreIncrError)))`.
```
janpetschexain, line 97: the described `Vec<_>` return type doesn't fit.

```suggestion
    /// - If the mask score has been successfully incremented, return `StorageResult::Ok(MaskScoreIncr(Result::Ok(())))`.
    /// - If the mask score could not be incremented due to a PET protocol error, return
    ///   the corresponding `StorageResult::Ok(MaskScoreIncr(Result::Err(MaskScoreIncrError)))`.
```

similarly as before, maybe without exposing internals:

```suggestion
    /// - If the mask score has been successfully incremented, return `StorageResult::Ok(MaskScoreIncr(Result::Ok(())))`.
    /// - If the mask score could not be incremented due to a PET protocol error, return
    ///   the corresponding `StorageResult::Ok(MaskScoreIncr(Result::Err(MaskScoreIncrError)))`.
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    /// # Behavior
+    ///
+    /// - If no masks exist, return `Result::Ok(Option::None)`.
+    /// - If only one mask exists, return this mask
+    ///   `Result::Ok(Option::Some(Vec<(MaskObject, u64)>))`.
+    /// - If two masks exist with the same score, return both
+    ///   `Result::Ok(Option::Some(Vec<(MaskObject, u64)>))`.
+    /// - If two masks exist with the different score, return
+    ///   both in descending order `Result::Ok(Option::Some(Vec<(MaskObject, u64)>))`.
```
janpetschexain, line 114:
```suggestion
    /// - If no masks exist, return `StorageResult::Ok(Option::None)`.
    /// - If only one mask exists, return this mask
    ///   `StorageResult::Ok(Option::Some(Vec<(MaskObject, u64)>))`.
    /// - If two masks exist with the same score, return both
    ///   `StorageResult::Ok(Option::Some(Vec<(MaskObject, u64)>))`.
    /// - If two masks exist with the different score, return
    ///   both in descending order `StorageResult::Ok(Option::Some(Vec<(MaskObject, u64)>))`.
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Deletes the [`SumDict`], [`SeedDict`] and `mask` dictionary.
+    async fn delete_dicts(&mut self) -> StorageResult<()>;
+
+    /// Sets the latest global model id.
+    ///
+    /// # Behavior
+    ///
+    /// - If no global model id has been set yet, set and return the new id `Result::Ok(String))`.
+    /// - If the global model id already exists, override and return the new id `Result::Ok(String))`.
+    async fn set_latest_global_model_id(&mut self, id: &str) -> StorageResult<()>;
```
janpetschexain, line 133: description doesn't match the return type, either

```suggestion
    /// - If no global model id has been set yet, set and return the new id `StorageResult::Ok(String)`.
    /// - If the global model id already exists, override and return the new id `StorageResult::Ok(String)`.
    async fn set_latest_global_model_id(&mut self, id: &str) -> StorageResult<String>;
```

or

```suggestion
    /// - If no global model id has been set yet, set the new id `StorageResult::Ok(())`.
    /// - If the global model id already exists, override with the new id `StorageResult::Ok(())`.
    async fn set_latest_global_model_id(&mut self, id: &str) -> StorageResult<()>;
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Gets the latest global model id.
+    ///
+    /// # Behavior
+    ///
+    /// - If the global model id does not exist, return `Result::Ok(None)`.
+    /// - If the global model id exists, return `Result::Ok(String))`.
```
janpetschexain, line 140:
```suggestion
    /// - If the global model id does not exist, return `StorageResult::Ok(None)`.
    /// - If the global model id exists, return `StorageResult::Ok(String)`.
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+where
+    Self: Clone + Send + Sync + 'static,
+{
+    /// Sets a global model.
+    ///
+    /// # Behavior
+    ///
+    /// - If the global model already exists (has the same model id), return `Result::Err(StorageError))`.
+    /// - If the global model does not exist, set the model and return the corresponding id.
+    async fn set_global_model(&mut self, id: &str, global_model: &Model) -> StorageResult<()>;
```
janpetschexain, line 156: description doesn't match the return type, either

```suggestion
    /// - If the global model already exists (has the same model id), return `StorageResult::Err(StorageError)`.
    /// - If the global model does not exist, set the model and return `StorageResult::Ok(())`.
    async fn set_global_model(&mut self, id: &str, global_model: &Model) -> StorageResult<()>;
```

or

```suggestion
    /// - If the global model already exists (has the same model id), return `StorageResult::Err(StorageError)`.
    /// - If the global model does not exist, set the model and return the corresponding id `StorageResult::Ok(String)`.
    async fn set_global_model(&mut self, id: &str, global_model: &Model) -> StorageResult<String>;
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Gets a global model.
+    ///
+    /// # Behavior
+    ///
+    /// - If the global model does not exist, return `Result::Ok(Option::None)`.
+    /// - If the global model exists, return `Result::Ok(Option::Some(Model))`.
```
janpetschexain, line 163:
```suggestion
    /// - If the global model does not exist, return `StorageResult::Ok(Option::None)`.
    /// - If the global model exists, return `StorageResult::Ok(Option::Some(Model))`.
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+    async fn get_global_model(&mut self, id: &str) -> StorageResult<Option<Model>>;
+
+    /// Creates a unique global model id by using the round id and the round seed in which
+    /// the global model was created.
+    ///
+    /// The format of the default implementation is as follows `roundid_roundseed`,
```
janpetschexain, line 169:
```suggestion
    /// The format of the default implementation is `roundid_roundseed`,
```

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+        format!("{}_{}", round_id, round_seed)
+    }
+}
+
+/// A wrapper that contains the result of the "add sum participant" operation.
+#[derive(Deref)]
+pub struct SumPartAdd(pub(crate) Result<(), SumPartAddError>);
+
+impl SumPartAdd {
+    /// Unwraps the result.
```
janpetschexain, line 182: actually this unwraps the wrapper, but not the result. not sure what would be a good terminology here though.

```suggestion
    /// Unwraps the wrapper.
```

same for the other unwraps below.

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub fn into_inner(self) -> Result<(), SumPartAddError> {
+        self.0
+    }
+}
+
+/// Error that can occur when adding a sum participant to the [`SumDict`].
+#[derive(Error, Debug, TryFromPrimitive)]
+#[repr(i64)]
+pub enum SumPartAddError {
```
janpetschexain, line 191: it would be nice to have a docstring for all variants of this enum, because its pub visible. similarly for the other pub enums in this module.

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
+//! Storage API.
```
janpetschexain, line 1: all the getters of the traits in this module take the implementer by `&mut self`. is that required by the types that usually would implement the traits? if so we should probably document this somewhere, otherwise i would expect the getters to just take `&self`.

rust/xaynet-server/src/storage/store.rs
```diff
@@ -0,0 +1,143 @@
...
+    C: CoordinatorStorage,
+    M: ModelStorage,
+{
+    /// A coordinator store.
+    coordinator: C,
+    /// A model store.
+    model: M,
+}
+
+impl<C, M> Store<C, M>
```
janpetschexain, line 38: reading this impl block i was wondering why you didn't impl the individual traits to expose their methods?

like having a `impl<C, M> Store<C, M> {}` for the `new()` method and then
```rust
impl<C, M> CoordinatorStorage for Store<C, M> {}
impl<C, M> ModelStorage for Store<C, M> {}
```

this way no method of the traits might be accidentally forgotten and changes to the trait will immediately be noticed in the `Store` impl. this is not to be confused with an overarching storage trait like in option 2 and 3 of #593.

janpetschexain (commented):
traits are looking good, i left some comments on documentation and the `Store` impl.

rust/xaynet-server/src/storage/api.rs
```diff
@@ -0,0 +1,240 @@
...
+};
+
+/// The error type for storage operations that are not directly related to application domain.
+/// These include, for example IOErrors like broken pipe, file not found, out-of-memory, etc.
+pub type StorageError = anyhow::Error;
+
+/// The result of the storage operation.
+pub type StorageResult<T> = Result<T, StorageError>;
+
+#[async_trait]
```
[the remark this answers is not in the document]
Robert-Steiner: fixed

rust/xaynet-server/src/state_machine/mod.rs
```diff
@@ -148,13 +148,13 @@ pub enum RequestError {
     Redis(#[from] RedisError),
 
     #[error(transparent)]
-    SeedDictUpdate(#[from] SeedDictUpdateError),
+    LocalSeedDictAdd(#[from] LocalSeedDictAddError),
```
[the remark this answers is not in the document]
Robert-Steiner: Good idea👍 I decided to use [`displaydoc`](https://github.com/yaahc/displaydoc) to avoid code duplications.

rust/xaynet-server/src/storage/api.rs
Robert-Steiner: Hm in [std](https://doc.rust-lang.org/std/io/struct.IntoInnerError.html?search=intoinner) they also have both versions. What about `Unwraps this wrapper, returning the underlying result.`?

rust/xaynet-server/src/storage/api.rs
Robert-Steiner: Good idea👍 I decided to use [`displaydoc`](https://github.com/yaahc/displaydoc) to avoid code duplications.

rust/xaynet-server/src/storage/api.rs
Robert-Steiner: I've made them `mut` just in case we want to use `tower` later, as the `tower` `call` method requires `mut`.

rust/xaynet-server/src/storage/store.rs
Robert-Steiner: I didn't do it because otherwise the method `create_global_model_id` would have to be called in every concrete implementation that implements the` ModelStorage` trait.

https://github.com/xaynetwork/xaynet/blob/d1cfc811917ac21d196a7e4dbf35030234c09213/rust/xaynet-server/src/storage/store.rs#L127-L137

I am not sure what is the best solution here. what do you prefer?

Robert-Steiner:
Any opinions on removing the `get_` prefix on the `get` methods?

rust/xaynet-server/src/storage/api.rs
janpetschexain: sounds good 👍

rust/xaynet-server/src/storage/api.rs
janpetschexain: that's a pretty nice feature 👍

rust/xaynet-server/src/storage/store.rs
janpetschexain: mhh, i see, so it seems to me that `create_global_model_id` is a helper method that we use for the `ModelStorage` trait impl and we don't expect every implementer of the trait to actually implement or use this, right? then it probably shouldn't be part of the trait itself, but instead it could be a helper function either living in the `api` module or the `store` module or some utils module.

janpetschexain (approved):

janpetschexain:
> Any opinions on removing the `get_` prefix on the `get` methods?

the [lang rfc](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis) suggests to not use the `get_` prefix for getters, but only `set_` prefix for setters. i guess we should stick with the rfc.

rust/xaynet-server/src/storage/store.rs
Robert-Steiner: I implemented the `CoordinatorStorage` and the `ModelStorage` trait for `Store`. Currently I prefer to keep the `create_global_model_id` method in the `ModelStorage` trait. Maybe I will make it a helper function later. Let's see how it will look like once the store is integrated.

janpetschexain (approved):

rust/xaynet-server/src/storage/store.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+#[async_trait]
+impl<C, M> CoordinatorStorage for Store<C, M>
+where
+    C: CoordinatorStorage,
+    M: ModelStorage,
+{
+    /// See [`CoordinatorStorage::set_coordinator_state`].
```
little-dude, line 57: If I'm not mistaken the doc is inherited from the trait definition, so this should not be needed.

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -169,14 +171,14 @@ impl Connection {
...
-    /// [`SumDictAddError::AlreadyExists`] if field already exists.
+    /// [`SumPartAddError::AlreadyExists`] if field already exists.
     ///
-    /// [`SumDictAddError::AlreadyExists`]: [crate::storage]
+    /// [`SumPartAddError::AlreadyExists`]: [crate::storage]
     pub async fn add_sum_participant(
         mut self,
         pk: &SumParticipantPublicKey,
         ephm_pk: &SumParticipantEphemeralPublicKey,
-    ) -> RedisResult<SumDictAdd> {
```
little-dude, line 179: The new name feels more correct to me. But at first it just confused me because I didn't directly understand that "part" was short for "participant". But I understand that `SumParticipantAddError` is annoyingly long...

rust/xaynet-server/src/storage/redis.rs
```diff
@@ -656,7 +661,7 @@ pub(in crate) mod tests {
 
         assert!(matches!(
             unknown_sum_pk.into_inner().unwrap_err(),
-            MaskDictIncrError::UnknownSumPk
+            MaskScoreIncrError::UnknownSumPk
```
little-dude, line 664: Really like the dict -> score renaming :+1:

little-dude (approved):
Very nice :+1: 

I think we should remove the docstrings from the trait impls, but otherwise let get this in :rocket:

rust/xaynet-server/src/storage/store.rs
Robert-Steiner: good point 👍 I removed it

PR #597, opened 2020-11-10
Impl ModelStorage for s3 client

Robert-Steiner:
**Summary**
- impls `ModelStorage` for s3 client
- refactored s3 client errors

### DIFF
--- a/rust/xaynet-server/src/state_machine/mod.rs
+++ b/rust/xaynet-server/src/state_machine/mod.rs
@@ -236,7 +236,7 @@ type StateMachineInitializationResult<T> = Result<T, StateMachineInitializationE
 pub enum StateMachineInitializationError {
     #[error("redis request failed: {0}")]
     Redis(#[from] RedisError),
-    #[error(transparent)]
+    #[error("failed to fetch global model: {0}")]
     Store(#[from] StorageError),
     #[error("failed to initialize crypto library")]
     CryptoInit,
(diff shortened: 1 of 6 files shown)


rust/xaynet-server/src/state_machine/mod.rs
```diff
@@ -229,9 +236,11 @@ type StateMachineInitializationResult<T> = Result<T, StateMachineInitializationE
 pub enum StateMachineInitializationError {
     #[error("redis request failed: {0}")]
     Redis(#[from] RedisError),
+    #[error(transparent)]
+    Store(#[from] StorageError),
     #[error("failed to initialize crypto library")]
     CryptoInit,
-    #[error("failed to fetch global model: {0}")]
+    #[error("{0}")]
```
little-dude, line 243: why removing the first part of the error description? It seems useful.
If the concern is that it's somewhat redundant with the inner String, I think it's fine. I'd rather have slightly redundant information than lose a bit of it

little-dude (approved):
This all looks good to me :+1:

rust/xaynet-server/src/state_machine/mod.rs
Robert-Steiner: previously there was no distinction between io error and the model does not exist
the io error moved to L240
and this error is just for the case that the model does not exist
I can add `failed to fetch global model:` to the `StorageError` wdyt?

little-dude: ahh ok I understand why you removed it then. +1 for adding this to `StorageError`

finiteprods (approved):
Looks good. So what were previously `download_global_model` / `upload_global_model` now become `global_model` / `set_global_model` (plus extra error handling to catch a model with the same id re-uploaded) of the new `ModelStorage` impl.

PR #598, opened 2020-11-11
Impl CoordinatorStorage for redis client

Robert-Steiner:
**Summary**
- impl `CoordinatorStorage` for redis client
- removed concurrency limit on redis client (we can add this feature back via `tower` later)
- improved error handling in phase states (each of the phase states has it own error enum)
- other small improvements


rust/xaynet-server/src/state_machine/phases/error.rs
```diff
@@ -48,31 +63,16 @@ impl Phase for PhaseState<PhaseStateError> {
...
-                .connection()
-                .await
-                .ping()
-                .await
-                .is_err()
-            {
-                info!("try to reconnect to Redis in 5 sec");
-                delay_for(Duration::from_secs(5)).await;
-            }
+        while self.shared.io.redis.coordinator_state().await.is_err() {
```
Robert-Steiner, line 73: I would like to improve this part in a follow up pr

little-dude: What kind of improvement are you planning on making here?

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -137,24 +113,72 @@ impl PhaseState<Unmask> {
...
+            .io
+            .redis
+            .set_latest_global_model_id(&global_model_id)
+            .await
+            .map_err(|err| warn!("failed to update latest global model id: {}", err));
+        Ok(())
+    }
+}
+
+#[cfg(feature = "metrics")]
```
little-dude, line 161: Nice to have these functions separated.

rust/xaynet-server/src/state_machine/phases/error.rs
Robert-Steiner: I'm not really sure yet

One option I see here is to add something like a `health` or `is_ready` method 
- for external storage solution it would return `Ok` once the client was able to reconnect
- for in-mem / filesystem it would always return `Ok` / or we could also check if enough space is left or so 

Another option could be that the memory itself is blocked until for example a client is reconnect. However, this can mean that the state machine is blocking in unusual places.

e.g.

https://github.com/xaynetwork/xaynet/blob/4fcf402a4f7cf2866d071b51bf572a2d4a1a7ba6/rust/xaynet-server/src/state_machine/phases/idle.rs#L40-L71

lets say the coordinator can set the state but it cannot broadcast the round_params because the redis store blocks in the `flush_dicts` method

rust/xaynet-server/src/storage/redis/mod.rs
```diff
@@ -70,115 +77,100 @@ use xaynet_core::{
...
+
+        // https://redis.io/commands/del
+        // > Return value:
+        //   The number of keys that were removed.
+        //
+        // Returns `0` if the key does not exist.
+        // We ignore the return value because we are not interested in it.
+
+        // delete sum dict
+        pipe.del("sum_dict").ignore();
```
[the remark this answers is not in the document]
Robert-Steiner: I agree, but I would like to do this in a seperate pr 
I will add this to the optimise redis 3 story

rust/xaynet-server/src/storage/redis/mod.rs
```diff
@@ -70,115 +77,100 @@ use xaynet_core::{
...
-            .field("semaphore", &self.semaphore)
-            .finish()
+        f.debug_struct("redis client").finish()
     }
 }
 
-pub struct Connection {
-    connection: ConnectionManager,
-    _permit: OwnedSemaphorePermit,
+fn to_storage_err(e: RedisError) -> StorageError {
```
little-dude, line 93: implementing `From<RedisError>` for `StorageError` feels a bit more natural imo

Robert-Steiner: At first I implemented that way but then I noticed I can't because of the orphan rule

rust/xaynet-server/src/storage/redis/mod.rs
```diff
@@ -193,40 +185,41 @@ impl Connection {
...
-        let seed_dict = result
-            .into_iter()
-            .map(|(pk, seed)| (pk.into(), seed.into()))
-            .collect();
+            .hgetall("sum_dict")
+            .await
+            .map_err(to_storage_err)?;
 
-        Ok(seed_dict)
+        let result = match reply.is_empty() {
```
little-dude, line 203: This is personal taste but I think an early return is more explicit here:

```rust
        if reply.is_empty() {
            return Ok(None);
        }

        let sum_dict = reply
            .into_iter()
            .map(|(pk, ephm_pk)| (pk.into(), ephm_pk.into()))
            .collect();

        Ok(Some(sum_dict))
```

rust/xaynet-server/src/storage/redis/mod.rs
```diff
@@ -280,48 +273,49 @@ impl Connection {
...
-                sum_pk.into(),
-                sum_pk_seed_dict
-                    .into_iter()
-                    .map(|(pk, seed)| (pk.into(), seed.into()))
-                    .collect(),
-            );
-        }
-
-        Ok(seed_dict)
+        let result = if sum_pks.is_empty() {
```
little-dude, line 288: Like above, I think I prefer a version with an early return in case `sum_pks` is empty.

rust/xaynet-server/src/storage/redis/mod.rs
little-dude: ahh because `StorageError` is an alias for `anyhow::Error`. Fair enough then.

little-dude (approved):
This looks good to me. Nice to see the error handling story improve with each PR too!

rust/xaynet-server/src/storage/redis/mod.rs
Robert-Steiner: fixed

rust/xaynet-server/src/storage/redis/mod.rs
Robert-Steiner: fixed

rust/xaynet-server/src/storage/redis/mod.rs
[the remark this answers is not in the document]
Robert-Steiner: I created a ticket for that

PR #602, opened 2020-11-13
add Phase<NewRound> test

little-dude:
This first unit test paves the way for writing unit tests for all the
state machine. The crux of it is the `MockIO` type that can be used to
avoid simulating a coordinator.

### DIFF
--- a/rust/xaynet-sdk/src/state_machine/phases/new_round.rs
+++ b/rust/xaynet-sdk/src/state_machine/phases/new_round.rs
@@ -69,3 +69,73 @@ impl Phase<NewRound> {
         state.into_phase(self.io)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::state_machine::{
+        testutils::{shared_state, SelectFor},
+        MockIO,
+        StateMachine,
+    };
+
+    #[tokio::test]
+    async fn test_selected_for_sum() {
+        let mut io = MockIO::new();
+        io.expect_notify_sum().return_const(());
+        let phase = make_phase(SelectFor::Sum, io);
+
+        let outcome = <Phase<NewRound> as Step>::step(phase).await;
+        matches!(outcome, TransitionOutcome::Complete(StateMachine::Sum(_)));
+    }
+
+    #[tokio::test]
+    async fn test_selected_for_update() {
+        let mut io = MockIO::new();
+        io.expect_notify_update().times(1).return_const(());
+        io.expect_notify_load_model().times(1).return_const(());
+        let phase = make_phase(SelectFor::Update, io);
+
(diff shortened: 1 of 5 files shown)


rust/xaynet-sdk/src/state_machine/phases/new_round.rs
```diff
@@ -69,3 +69,64 @@ impl Phase<NewRound> {
...
+        let shared = shared_state(task);
+
+        // Check IntoPhase<NewRound> implementation
+        let mut mock = MockIO::new();
+        mock.expect_notify_new_round().times(1).return_const(());
+        let mut phase: Phase<NewRound> = State::new(shared, NewRound).into_phase(Box::new(mock));
+
+        // Note that this has the side effect of dropping `mock`
+        // forcing the checks to run
+        let _ = std::mem::replace(&mut phase.io, Box::new(io));
```
Robert-Steiner, line 129: I don't really understand why we have to do this

Robert-Steiner (approved):
Nice that the tests are so small 👍
if I remember correctly, we also used `mockall` in the old coordinator, right?

rust/xaynet-sdk/src/state_machine/phases/new_round.rs
little-dude: It's a bit tricky.

Ideally in the tests we should be able to do something like this:

```rust
    #[tokio::test]
    async fn test_selected_for_sum() {
        let mut phase = make_phase(SelectFor::Sum);
        phase.io.expect_notify_sum().return_const(());
        let outcome = <Phase<NewRound> as Step>::step(phase).await;
        matches!(outcome, TransitionOutcome::Complete(StateMachine::Sum(_)));
    }
```
But `phase.io` is a `Box<dyn IO>`, not a `MockIO`. So it doesn't have `expect_notify_sum()`. Therefore the test need to pass the mock it wants to use to `make_phase`:

```rust
// create a mock with the test's expectations
let mut io = MockIO::new();
io.expect_notify_sum().return_const(());
// pass the mock to `make_phase`. Internally `make_phase` sets `phase.io` to the mock we're passing
let phase = make_phase(SelectFor::Sum, io);

// rest of the test
```

little-dude: Maybe my comment is not very helpful. I'll reword it.

Robert-Steiner: Ahh ok I see 
thanks for the explanation 👍

little-dude: I documented the function and edited the comment, with a shorter version of the explanation above.

PR #603, opened 2020-11-13
alternative aggregation PoC

finiteprods:
**Warning.** The changes here are meant just for illustrative PoC purposes, and should not be merged.

This branch contains some test code to prove that it is possible, in principle, to support an alternative aggregation to "federated averaging" in the PET protocol. The particular example illustrated here is a _histogram_ aggregation. A quick breakdown of the experiment:

- the coordinator and participants are assumed to be in agreement about the ranges of values in the histogram, e.g. `0 - 5`, `5 - 10`, `10 - 15`, `15 - 20`. 
- the test-drive is used to simulate clients providing a "measurement" in one of those ranges, e.g. for a measurement of 7.5, this falls into the `5 - 10` range. To convey this range to the coordinator, it would construct a "model" `[0, 1, 0, 0]`.
- the PET protocol runs as usual (not completely true - explained later), so that the coordinator learns none of the individual models, but still unmasks their overall sum. From this, the coordinator has the histogram.

In a particular test run of this, spinning up a coordinator and running the test-drive with `-n 10`, I observed

- 7 update participants computed a masked model. Of these:
- 2 sent `[1, 0, 0, 0]`
- 3 sent `[0, 1, 0, 0]`
- 2 sent `[0, 0, 0, 1]`

On the coordinator, the unmasked model and histogram is visible in the console output:

```
histogram for [2, 3, 0, 2]

    *        
*   *       *
*   *       *
*************
```
**Further remarks.** 

- with some refinements to the histogram protocol above (not shown here), it is possible to compute other kinds of aggregations, such as a maximum or minimum. This is still done in a privacy-preserving way (whether these cover a reasonable amount of our mobile analytics use cases, is still to be investigated).
- how one _parametrises_ the aggregation, e.g. switch between averaging or histogram or something else, is something to be considered elsewhere.
- the test-drive still uses dummy models (just in this slightly different form). In the real implementation, a client would construct a model based on some value accessible on the device.
- as mentioned above, every party was assumed to know the ranges of the histogram upfront. In practice, this would need to be communicated somehow - clients need this information to compute their models.
- a small modification was made to the unmasking part of the PET protocol - we skip the "correction" step which re-scales the unmasked vector. The reason is because we would like a straightforward _sum_ of the models, rather than a weighted average. 
- (minor point) to print the mini-histogram, I used a tiny crate `hist`. While the above output looks sensible, my mileage varied a lot! it behaves a little peculiarly, sometimes adding / removing a point. It's not documented so perhaps I'm misusing it.

### DIFF
--- a/configs/docker-dev.toml
+++ b/configs/docker-dev.toml
@@ -14,8 +14,8 @@ min_sum_time = 5
 min_update_time = 10
 max_sum_time = 3600
 max_update_time = 3600
-sum = 0.01
-update = 0.1
+sum = 0.5
+update = 0.9
 
 [mask]
 group_type = "Prime"
@@ -31,7 +31,7 @@ url = "http://influxdb:8086"
 db = "metrics"
 
 [redis]
-url = "redis://redis"
+url = "redis://127.00.0.1/"
 
 [s3]
 access_key = "minio"
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -946,6 +946,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "644f9158b2f133fd50f5fb3242878846d9eb792e445c893805ff0e3824006e35"
 
 [[package]]
+name = "hist"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


finiteprods:
closing as this is superseded by #635

PR #604, opened 2020-11-13
Impl store in state machine

Robert-Steiner:
**Summary**
- removed `IO` struct in `Shared` (it has no real benefit)
- made the `StateMachine`, `StateMachineBuilder` and `StateMachineInitializer` generic over `Store`
- refactored `StateMachineBuilder` (used in `StateMachine` tests)
- implemented a `NoOp` model store, since we don't have any alternative implementation yet and `Store` requires something that implements `ModelStorage`
  - however the `NoOp` methods are never called because they are still guarded behind the `model-persistence` feature flag 
- refactored `main.rs`
- renamed `state.inner` to `state.private` to be consitent with the client state machine 
- refactored tests 

TODO
- [x] checking docs
- [x] remove `FIXME`
- [x] update diagram


little-dude:
I guess we can get rid of the `model-persistence` feature once we have a file-backed store? Like by default it would store the models in files, but we can have an option to use S3.

rust/xaynet-server/src/state_machine/mod.rs
```diff
@@ -238,29 +246,34 @@ pub enum StateMachineInitializationError {
...
-    redis_handle: redis::Client,
-    #[cfg(feature = "model-persistence")]
-    s3_handle: s3::Client,
+    store: Store<C, M>,
     #[cfg(feature = "metrics")]
     metrics_handle: MetricsSender,
 }
 
-impl StateMachineInitializer {
+impl<C, M> StateMachineInitializer<C, M>
```
little-dude, line 265: This module is growing. Maybe it would be worth having an `initializer.rs` module. In a follow-up PR though.

rust/xaynet-server/src/state_machine/mod.rs
```diff
@@ -329,20 +341,22 @@ impl StateMachineInitializer {
...
             self.metrics_handle,
         );
 
-        let state_machine = StateMachine::from(PhaseState::<Idle>::new(shared));
+        let state_machine = StateMachine::from(PhaseState::<Idle, _, _>::new(shared));
         (state_machine, request_tx, event_subscriber)
     }
 }
 
 #[cfg(feature = "model-persistence")]
```
little-dude, line 354: Do we still need this `#[cfg]` attribute? The store should work even when `mode-persistence` isn't enabled now, since the store internally uses `NoOp`

little-dude: mmh but may `load_global_model` wouldn't work then

rust/xaynet-server/src/state_machine/phases/mod.rs
```diff
@@ -70,60 +72,52 @@ pub trait Handler {
     async fn handle_request(&mut self, req: StateMachineRequest) -> Result<(), RequestError>;
 }
 
-/// I/O interfaces.
-#[cfg_attr(test, derive(Debug))]
-pub struct IO {
+/// A struct that contains the coordinator state and the I/O interfaces that are shared and
+/// accessible by all `PhaseState`s.
+pub struct Shared<C, M>
```
little-dude, line 77: Maybe we could derive `Debug`, unless some fields aren't `Debug`?

rust/xaynet-server/src/storage/tests/mod.rs
```diff
@@ -1,133 +1,27 @@
...
-use xaynet_core::{
-    crypto::{ByteObject, EncryptKeyPair, SigningKeyPair},
-    mask::{BoundType, DataType, EncryptedMaskSeed, GroupType, MaskConfig, MaskObject, ModelType},
-    LocalSeedDict,
-    SeedDict,
-    SumDict,
-    SumParticipantEphemeralPublicKey,
-    SumParticipantPublicKey,
-    UpdateParticipantPublicKey,
+pub mod utils;
```
little-dude, line 1: :+1:

little-dude (approved):
:+1: This was surprisingly easy to review. Mostly new generic params. I skipped the tests changes though.

Really like that you removed the `Shared.io` attribute, it looks nicer. I can't wait to get rid of the `model-persistence` feature, it should lead to some simplifications.

rust/xaynet-server/src/state_machine/mod.rs
Robert-Steiner: I agree, I will add a ticket in the storage api follow up story

rust/xaynet-server/src/state_machine/mod.rs
Robert-Steiner: Right, the problem is that if redis returns a `latest_global_model_id` and the `NoOp` model store returns `None` or an error, the state initializer would fail. 
Maybe it is okay to initialize the state machine without the latest global model, but I want to test this in a separate ticket. Or we can somehow find out at run time which storage implementations are in use and print a warning if the recovery flag is true but the storage doesn't support it.

rust/xaynet-server/src/state_machine/phases/mod.rs
Robert-Steiner: Storage is not `Debug` because `Redis` and `S3` do not provide `Debug` and I thought there was no point having an empty debug for both of them.
But I have added a custom `Debug` that includes `state`, `events` and `request_rx`.

rust/xaynet-server/src/state_machine/mod.rs
little-dude: Let's address that properly in another PR then. I think once we have a file backend for models, we should make it the default, and add a `s3-model-persistence` feature instead, that enables the s3 backend.

PR #610, opened 2020-11-18
Add bors config

Robert-Steiner:
enables bors

there are some more options we can set 
https://bors.tech/documentation/

Let me know if I should add/remove any of them

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,0 +1,13 @@
+status = [
+    "cargo-fmt",
+    "cargo-check",
+    "cargo-clippy",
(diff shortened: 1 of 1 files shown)


wilk10 (approved):
are you waiting for review/approval here as well?

Robert-Steiner:
> are you waiting for review/approval here as well?

not right now 😊 
but you can review #625 and #640 if you like

Robert-Steiner:
we will tackle this later
first we should check if bors is working together with 
- dependabot 
- protected branches

PR #613, opened 2020-11-19
make the masking configuration part of the round parameters

little-dude:


rust/xaynet-mobile/xaynet_ffi.h
```diff
@@ -7,50 +7,15 @@
...
-  BOUNDTYPE_B6 = 6, /** Numerical values absolutely bounded by 1_000_000 */
-  BOUNDTYPE_BMAX = 255, /** Numerical values absolutely bounded by their original primitive data type's maximum absolute */
-};
-
-/**
- * The original primitive data type of the numerical values to be masked.
+ * The original primitive data type of the model weights.
  */
-enum MASKCONFIG_DATATYPE {
+enum SETMODEL_DATATYPE {
```
Robert-Steiner, line 12: Later we can i think about getting that information from the client and using it directly in the `xaynet_ffi_participant_set_model` method

```rust
pub unsafe extern "C" fn xaynet_ffi_participant_set_model(
    participant: *mut Participant,
    buffer: *const c_void,
    len: c_uint,
) -> c_int {
    let participant = match unsafe { participant.as_mut() } {
        Some(participant) => participant,
        None => return ERR_NULLPTR,
    };

    let data_type = participant.model_settings().data_type;

    let model = match data_type {
    ...
    }
... 
}

```

wdyt?

Robert-Steiner (approved):
Awesome to see that not only make the participant easier to use but also reduces the loc 👍

rust/xaynet-mobile/xaynet_ffi.h
little-dude: Good point. I'm going to touch to the area of the code anyway, because we want to make the scalar part of the model store.

PR #614, opened 2020-11-20
add rustc version to cache key

Robert-Steiner:
Adds the version of rustc to the cache key. As soon as a new rust version is available, a new cache is created instead of using the old one. Without the rustc version, the old cache would be used and the dependencies would be constantly recompiled.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,11 +15,15 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install nightly toolchain
+        id: rust-toolchain
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: nightly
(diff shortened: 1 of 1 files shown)


little-dude:
Wow how did you notice that @Robert-Steiner ?

little-dude (approved):

Robert-Steiner:
@acrrd drew my attention to this

you can see quite well what effect it has if you don't take the version into account in this [ci run](https://github.com/xaynetwork/xaynet/runs/1429485938)

Screen Shot 2020-11-20 at 15 53 29

It uses the old cache, but since a new rust version was released yesterday, some of the dependencies need to be recompiled. However, in the end the new artifacts are not saved because the key is already there

Screen Shot 2020-11-20 at 15 53 36

PR #615, opened 2020-11-20
Storage is ready method

Robert-Steiner:
Adds a `is_ready` method that can be used to check whether the `Store` is able to process a request.
The idea of ​​this method comes from the `ready` / `poll_ready` methods of the [`tower::ServiceExt::ready`](https://docs.rs/tower/0.3.1/tower/trait.ServiceExt.html#method.ready) /  [`tower::Service::poll_ready`](https://docs.rs/tower/0.3.1/tower/trait.Service.html#tymethod.poll_ready)  traits.


little-dude (approved):
:+1:

PR #616, opened 2020-11-23
Update warp dependency

janpetschexain:
our changes have been merged to `warp`, so we can depend on it again (and remove our fork). once they release a new version to crates.io, we can change to the registry as well.

### DIFF
--- a/rust/xaynet-server/Cargo.toml
+++ b/rust/xaynet-server/Cargo.toml
@@ -66,7 +66,7 @@ tracing = "0.1.19"
 tracing-futures = "0.2.4"
 tracing-subscriber = "0.2.12"
 validator = { version = "0.11.0", features = ["derive"] }
-warp = { git = "https://github.com/xaynetwork/warp" }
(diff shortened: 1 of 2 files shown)


rust/xaynet-server/Cargo.toml
```diff
@@ -66,7 +66,7 @@ tracing = "0.1.19"
 tracing-futures = "0.2.4"
 tracing-subscriber = "0.2.12"
 validator = { version = "0.11.0", features = ["derive"] }
-warp = { git = "https://github.com/xaynetwork/warp" }
+warp = { git = "https://github.com/seanmonstar/warp", branch = "master" }
```
little-dude, line 69: Maybe we should pin this to a certain commit. The master branch may contain breaking changes at any point.

janpetschexain: good idea, i'll change that.

little-dude (approved):

PR #617, opened 2020-11-23
xaynet-sdk: allow other HTTP clients

little-dude:
The only client we currently support is `reqwest` but this opens the
door to supporting more clients in the future.

To add a new client `NewClient`, we just need to implement the
`HttpClient` trait.


rust/xaynet-sdk/src/client.rs
```diff
@@ -0,0 +1,194 @@
...
+        Self::Deserialize(format!("{:?}", e))
+    }
+}
+
+/// A trait that Xaynet clients must implement
+#[async_trait]
+pub trait XaynetHttpClient {
+    type Error: std::error::Error;
+
+    async fn get(&mut self, url: &str) -> Result<Rc<[u8]>, ClientError>;
```
little-dude, line 60: The reason for this `Rc` was that I didn't to impose `Vec<u8>`, but I couldn't just do a `&[u8]`. But I think I found a better way to do this by using an associated `GetResponse` type actually.

little-dude:
So one thing I'm not sure about is whether we should re-export `reqwest` in `xaynet_sdk`.

In `xaynet_mobile` and in the test drive example, we now have to depend explicitely _on the same version of `reqwest` than the one used in `xaynet_sdk`_. If we re-export it, we can just use `reqwest from `xaynet_sdk`. _But_, I'm not sure how it work wrt to features. In `xaynet_sdk` we don't enable any specific feature. But in `xaynet_mobile`, we enable `rustls-tls`. I'm not sure if cargo is smart enough to compile only one version of the crate and re-export it with `rustls-tls` enabled in `xaynet-sdk` in that case

Robert-Steiner (commented):
Looks good so far 👍 I will take a closer look tomorrow

Robert-Steiner:
> So one thing I'm not sure about is whether we should re-export `reqwest` in `xaynet_sdk`.
> 
> In `xaynet_mobile` and in the test drive example, we now have to depend explicitely _on the same version of `reqwest` than the one used in `xaynet_sdk`_. If we re-export it, we can just use `reqwest` from `xaynet_sdk`. _But_, I'm not sure how it work wrt to features. In `xaynet_sdk` we don't enable any specific feature. But in `xaynet_mobile`, we enable `rustls-tls`. I'm not sure if cargo is smart enough to compile only one version of the crate and re-export it with `rustls-tls` enabled in `xaynet-sdk` in that case

In the current setup cargo only builds one version of `reqwest` with `rustls-tls` enabled and uses it for both `xaynet-sdk` and `xaynet-mobile`.
you can see it when you run `cargo +nightly build -Z unstable-options --build-plan` in `xaynet-mobile`
However, I'm not sure how re-exporting could help here because we would still need to add `reqwest = { version = "0.10", default-features = false, features = ["rustls-tls"]}` in `rust/xaynet-mobile/Cargo.toml` to enable `rustls-tls` feature right? 
Or do I missing something?

rust/xaynet-sdk/src/client.rs
```diff
@@ -0,0 +1,194 @@
...
+        Ok(self.get(&url).await?)
+    }
+
+    async fn get_model(&mut self) -> Result<Option<Model>, Self::Error> {
+        let url = self.url("model");
+        Ok(self.get(&url).await?)
+    }
+
+    async fn send_message(&mut self, msg: Vec<u8>) -> Result<(), Self::Error> {
+        let url = self.url("model");
```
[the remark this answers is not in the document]
little-dude: Geez

little-dude:
Wow I didn't know about `--build-plan`, so nice! You're right that re-exporting `reqwest` doesn't really help.

little-dude:
I broke something:

```
Nov 24 09:30:17.449  WARN agent{id=4}: xaynet_sdk::state_machine::phases::sum2: failed to fetch seeds: HTTP request failed: HTTP status client error (400 Bad Request) for url (http://127.0.0.1:8081/seeds?pk=mpr5BMwLzdxjRVuLAlNl5aJBNa3UAMh
oVLc0IxhvZds%3D)
```

little-dude:
Oh. Of course: I need to update the `warp` part in xaynet-server :facepalm:

little-dude:
Seems like deserialization is broken on the client side:

```
Nov 24 09:58:52.222  WARN agent{id=26}: xaynet_sdk::state_machine::phases::update: failed to fetch sum dictionary: Deserialize("tag for enum is not valid, found 9")
```

and

```
Nov 24 09:58:50.387  WARN agent{id=14}: xaynet_sdk::state_machine::phases::sum2: failed to fetch seeds: failed to deserialize data: io error: unexpected end of file
```

little-dude:
Found the root cause.

rust/xaynet-sdk/Cargo.toml
```diff
@@ -27,9 +27,15 @@ thiserror = "1.0.20"
...
+url = "2.2.0"
+base64 = "0.13.0"
+bincode = { version = "1.3.1" }
+
 # feature: reqwest client
-bincode = { version = "1.3.1", optional = true }
-reqwest = { version = "0.10.8", default-features = false, optional = true }
+reqwest = { version = "0.10", default-features = false, optional = true }
+# This has to match the version used by reqwest. It would be nice if
+# rewest just re-exported it
```
Robert-Steiner, line 37:
```suggestion
# reqwest just re-exported it
```

rust/xaynet-sdk/src/client.rs
```diff
@@ -0,0 +1,217 @@
...
+    type GetResponse = bytes::Bytes;
+
+    async fn get(&mut self, url: &str) -> Result<Option<Self::GetResponse>, ClientError> {
+        let resp = reqwest::Client::get(self, url)
+            .send()
+            .await
+            .map_err(ClientError::http_error)?
+            .error_for_status()
+            .map_err(ClientError::http_error)?;
+        match resp.status() {
```
Robert-Steiner, line 198: Nice!

rust/xaynet-server/src/rest.rs
```diff
@@ -226,14 +233,17 @@ fn with_fetcher<F: Fetcher + Sync + Send + 'static + Clone>(
 }
 
 /// Extracts a participant public key from a request body
```
Robert-Steiner, line 235:
```suggestion
/// Extracts a participant public key from the url query string
```

Robert-Steiner (approved):
Looks good to me!

PR #624, opened 2020-11-30
XN-1370 stable docs

janpetschexain:
**References**

- [XN-1370](https://xainag.atlassian.net/browse/XN-1370)

**Summary**

- updates msrv and github workflow to use stable cargo doc
- update readme and template
- add a readme check to the workflow: use `cargo readme` to generate a readme from the code docs + template. this will notify us if the docs and the readme diverge. in consequence no changes should be made to the readme itself, but only to the docs and the template, then the readme can be generated via `cargo readme > README.md`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -191,13 +191,12 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      - name: Install nightly toolchain
+      - name: Install stable toolchain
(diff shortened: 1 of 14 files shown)


Robert-Steiner (approved):
👍
we also need to update the version in the README (badge + text)

janpetschexain:
> 👍
> we also need to update the version in the README (badge + text)

good catch, i'll change that.

PR #626, opened 2020-12-01
fix multipart messages

little-dude:
There were multiple combined issues:

- a bug in the `ExactSizeIterator` implementation for `MultipartMessageBuffer`
- in `xaynet_sdk`, the chunk IDs start at 0, while the server expected
  them to start at 1. The server now expects the chunk IDs to start a 0
- under certain conditions, `MessageBuilder::has_all_chunk()` would
  never return `true`
- in `xaynet_sdk`, we were generating new random message IDs for each
  chunks

### DIFF
--- a/rust/xaynet-server/src/services/messages/multipart/service.rs
+++ b/rust/xaynet-server/src/services/messages/multipart/service.rs
@@ -8,6 +8,7 @@ use futures::{
     task::Context,
 };
 use tower::Service;
+use tracing::{debug, trace, warn};
 
 use crate::services::messages::{multipart::buffer::MultipartMessageBuffer, ServiceError};
 use xaynet_core::{
@@ -50,7 +51,10 @@ impl MessageBuilder {
     /// builder holds all the chunks.
     fn has_all_chunks(&self) -> bool {
         self.last_chunk_id
-            .map(|expected_number_of_chunks| self.data.len() == expected_number_of_chunks as usize)
+            .map(|last_chunk_id| {
+                // The IDs start at 0, hence the + 1
+                self.data.len() >= (last_chunk_id as usize + 1)
+            })
             .unwrap_or(false)
     }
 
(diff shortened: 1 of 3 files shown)


little-dude:
Probably a bug in the iterator `size_hint`:

```rust
impl Iterator for MultipartMessageBuffer {
    type Item = u8;

    fn next(&mut self) -> Option<Self::Item> {
        if self.current_chunk.is_none() {
            let (_, chunk) = self.remaining_chunks.next()?;
            self.current_chunk = Some(chunk.into_iter());
            return self.next();
        }

        // Per `if` above, `self.current_chunk` is not None
        match self.current_chunk.as_mut().unwrap().next() {
            Some(b) => {
                self.consumed += 1;
                Some(b)
            }
            None => {
                self.current_chunk = None;
                self.next()
            }
        }
    }

    fn size_hint(&self) -> (usize, Option<usize>) {
        let lower_bound = self.initial_length - self.consumed;
        let upper_bound = self.initial_length;
        (lower_bound, Some(upper_bound))
    }
}
```

rust/xaynet-server/src/services/messages/multipart/service.rs
```diff
@@ -138,21 +144,28 @@ impl Service<Message> for MultipartHandler {
...
             // and return it
             if mp_message.has_all_chunks() {
+                debug!("received the final message chunk, now parsing the full message");
                 // This entry exists, because `mp_message` above
                 // refers to it, so it's ok to unwrap.
                 match self.message_builders.remove(&id).unwrap().into_message() {
-                    Ok(message) => ready_ok(Some(message)),
-                    Err(e) => ready_err(ServiceError::Parsing(e)),
+                    Ok(message) => {
+                        debug!("multipart message succesfully parsed ");
```
Robert-Steiner, line 162:
```suggestion
                        debug!("multipart message succesfully parsed");
```

Robert-Steiner (approved):
🚀

PR #628, opened 2020-12-03
xaynet-mobile as rlib

janpetschexain:
**Summary**

- set the `xaynet-mobile` as `rlib` next to the other `crate_type`s to make it usable downstream as a rust dependency
- re-export the `ffi_support` types which are used in the public api to avoid downstream version collisions by separate imports

### DIFF
--- a/rust/xaynet-mobile/src/ffi/mod.rs
+++ b/rust/xaynet-mobile/src/ffi/mod.rs
@@ -6,6 +6,7 @@ pub use participant::*;
 mod settings;
 pub use settings::*;
 
+pub use ffi_support::{ByteBuffer, FfiStr};
 use std::os::raw::c_int;
 
 /// Destroy the given `ByteBuffer` and free its memory. This function must only be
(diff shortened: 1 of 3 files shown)


Robert-Steiner (approved):

rust/xaynet-mobile/src/ffi/mod.rs
```diff
@@ -6,6 +6,7 @@ pub use participant::*;
 mod settings;
 pub use settings::*;
 
+pub use ffi_support::{ByteBuffer, FfiStr};
```
little-dude, line 9: Why do we need `FfiStr`? I don't see it being used in the diff

little-dude (approved):

rust/xaynet-mobile/src/ffi/mod.rs
janpetschexain: it is used in the `participant` and `settings` mods, which are exposed as part of the publich api.

PR #629, opened 2020-12-04
impl get global model

Robert-Steiner:
**Summary**
- adds the `get_global_model` method / function to the `xaynet-mobile::Participant` / C-ffi
- adds a method to retrieve the `ModelConfig` of the model that is expected in the update phase

TODO
- [x] update C tests (I will add a get global model test after the first review)


rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -345,3 +355,93 @@ pub unsafe extern "C" fn xaynet_ffi_participant_set_model(
         ERR_SETMODEL_MODEL
     }
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn xaynet_ffi_participant_global_model(
+    participant: *mut Participant,
+    buffer: *mut c_void,
+    // len: c_uint,
```
Robert-Steiner, line 363: Not sure if we should keep the `len` and the `data_type` here.
both can be accessed through `participant.model_config().model_length` and `participant.model_config().mask_config.vect.data_type`.

rust/xaynet-mobile/src/participant.rs
```diff
@@ -133,6 +142,8 @@ pub struct Participant {
     store: Store,
     /// Async runtime to execute the state machine
     runtime: Runtime,
+    /// Xaynet client
+    client: Client<reqwest::Client>,
```
Robert-Steiner, line 146: Should that be better an `Arc`?

rust/xaynet-mobile/src/participant.rs
```diff
@@ -293,4 +310,21 @@ impl Participant {
         });
         self.should_set_model = false;
     }
+
+    /// Retrieve the current global model, if available.
+    pub fn global_model(&mut self) -> Option<Model> {
```
Robert-Steiner, line 315: Here I'm also not sure if we should return `Result<Option<Model>>` or just `Option<Model>`

rust/xaynet-mobile/src/participant.rs
little-dude: `reqwest::Client` is already an `Arc`, internally, so I think it's fine.

```rust
#[derive(Clone)]
pub struct Client {
    inner: Arc<ClientRef>,
}
```

rust/xaynet-mobile/src/participant.rs
Robert-Steiner: i changed it to `Result<Option<Model>>` so that the caller can retry if the call failed due to a IO error

rust/xaynet-mobile/src/ffi/config.rs
```diff
@@ -0,0 +1,56 @@
...
+    if model_config.is_null() {
+        return ERR_NULLPTR;
+    }
+    pv::_xaynet_ffi_model_config_destroy(model_config);
+    OK
+}
+
+#[repr(C)]
+/// The model configuration of the model that is expected in [`xaynet_ffi_participant_set_model`].
+pub struct ModelConfig {
```
Robert-Steiner, line 42: maybe redundant? we could also just annotate `ModelConfig` with `#[repr(C)]` in https://github.com/xaynetwork/xaynet/pull/629/files#diff-ae8e63e3ded5ff755ba7e68cae29f44ed121b618634584aed4d1ac4b2a86a188R290

rust/xaynet-mobile/xaynet_ffi.h
```diff
@@ -117,6 +132,34 @@ enum SETMODEL_DATATYPE {
...
+#define PARTICIPANT_NEW_GLOBAL_MODEL (1 << 5)
+
+/**
+ * The original primitive data type of the numerical values to be masked.
+ */
+enum DataType {
+  /**
+   * Numbers of type f32.
+   */
+  F32 = 0,
```
little-dude, line 147: These are constants in C, so I think their names should be a little more specific, like `MODEL_DATATYPE_F32`, etc.

rust/xaynet-mobile/src/ffi/config.rs
little-dude: Yeah I'm not sure to be honest. On the one hand, I think it's nice to have separate types for FFI, but on the other hand, it's a bit artificial sometimes. So I don't really have a preference

rust/xaynet-mobile/src/ffi/mod.rs
```diff
@@ -97,11 +100,21 @@ pub const ERR_SETTINGS_MASKCONFIG: c_int = 8;
...
-/// Failed to set the local model: data type
+/// Failed to set the local model: invalid data type
 pub const ERR_SETMODEL_DATATYPE: c_int = 11;
 /// Failed to initialized the crypto library
 pub const ERR_CRYPTO_INIT: c_int = 12;
 /// Invalid secret signing key
 pub const ERR_CRYPTO_SECRET_KEY: c_int = 13;
 /// Invalid public signing key
 pub const ERR_CRYPTO_PUBLIC_KEY: c_int = 14;
+/// Global model does not exist
```
little-dude, line 111:
```suggestion
/// No global model is currently available
```

rust/xaynet-mobile/tests/ffi_test.c
```diff
@@ -74,6 +74,34 @@ static char *test_settings() {
...
+  with_keys(settings);
+  with_url(settings);
+  xaynet_ffi_settings_set_url(settings, "http://localhost:8081");
+
+  Participant *participant = xaynet_ffi_participant_new(settings);
+  mu_assert("failed to create participant", participant != NULL);
+  ModelConfig *model_config = xaynet_ffi_participant_model_config(participant);
+  float* buffer = (float *)malloc(sizeof(float) * model_config->len);
+
+  int err = xaynet_ffi_participant_global_model(NULL, buffer, model_config->data_type, model_config->len);
```
little-dude, line 88: Why not have `xaynet_ffi_participant_global_model` take `*ModelConfig` as argument instead of passing the data type and length separately?

rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -345,3 +362,143 @@ pub unsafe extern "C" fn xaynet_ffi_participant_set_model(
...
+/// - [`ERR_NULLPTR`] if `participant` or the `buffer` is NULL
+/// - [`GLOBAL_MODEL_NONE`] if no model exists
+/// - [`ERR_GLOBALMODEL_IO`] if the communication with the coordinator
+/// - [`ERR_GLOBALMODEL_DATATYPE`] if the datatype is invalid
+/// - [`ERR_GLOBALMODEL_LEN`] if the length of the buffer does not match the length of the model
+/// - [`ERR_GLOBALMODEL_CONVERT`] if the conversion of the model failed
+///
+/// # Note
+///
+///    It is **not** guaranteed, that the model configuration returned by
```
little-dude, line 385: I think there's an extra indentation level on this paragraph. Not sure if that affects how it is rendered though.

rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -345,3 +362,143 @@ pub unsafe extern "C" fn xaynet_ffi_participant_set_model(
...
+        Err(_) => return ERR_GLOBALMODEL_DATATYPE,
+    };
+
+    let len = len as usize;
+    if len != global_model.len() {
+        return ERR_GLOBALMODEL_LEN;
+    }
+
+    match data_type {
+        DataType::F32 => {
```
little-dude, line 433: Not sure, bit maybe a macro would help making this a bit shorter.

rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -345,3 +362,143 @@ pub unsafe extern "C" fn xaynet_ffi_participant_set_model(
...
+///
+/// 1. When calling this method, you have to ensure that *either* the pointer is NULL
+///    *or* all of the following is true:
+///    - The pointer must be properly [aligned].
+///    - It must be "dereferencable" in the sense defined in the [`::std::ptr`] module
+///      documentation.
+/// 2. If `len` or `data_type` do not match the model in `buffer`, this method will
+///    result in a buffer over-read.
+#[no_mangle]
+pub unsafe extern "C" fn xaynet_ffi_participant_global_model(
```
little-dude, line 401: So when the coordinator returns the model config along with the model, the idea is that this function will look like this, correct?

```rust
pub unsafe extern "C" fn xaynet_ffi_participant_global_model(participant: *mut Participant, *mut ModelConfig) -> *const ByteBuffer
{
  // ...
}
```
Where `*ModelConfig` is populated based on the model config sent by the coordinator, and `ByteBuffer` contains the raw model.

little-dude (commented):
I left a few comments but this looks great already.

rust/xaynet-mobile/src/ffi/participant.rs
Robert-Steiner: My idea was that we have something like this 
```rust
pub unsafe extern "C" fn xaynet_ffi_participant_global_model(participant: *mut Participant) -> *const ByteBuffer
{
  // p.global_model() 
  // would return a struct which contains the model as well as the data type.
  // Note that we have not currently linked the data type to the model, we just send the model

  GlobalModel {
  	model: Model,
  	type: DataType
  }

  // then we can convert the model in this function and write it in a ByteBuffer 
  let model = match global_model.type { ... }

  return GlobalModelAsCRep {
  	model: ByteBuffer,
  	type: DataType
  }
}
```

but I have already noticed a flaw in my idea
we can't use `ByteBuffer` here because it can only holds an `u8` array. The data type of a model however can be `f32`, `f64`, `i32` or `i64`. So I don't know what the reutrn type should be. 
what we maybe can do is a create a buffer that holds a c_void pointer instead of `u8` but I'm not sure if this is possible  if we can implement this in a safe way.

rust/xaynet-mobile/src/ffi/participant.rs
Robert-Steiner: That is a good idea👍

rust/xaynet-mobile/tests/ffi_test.c
Robert-Steiner: The `ModelConfig` only describes the model that is expected in the `set_model` function. I didn't want to create the impression that `ModelConfig` belongs to the global model. That was the only reason. I mean the user can use the config as long as the config is never changed. However, it can change if a user restarts the coordinator with a different model len/ mask configuration.
But yeah maybe it is better to use the `ModelConfig`. I will change that

finiteprods: if you're worried that `ModelConfig` might cause confusion with that of the global model (and I think that is a valid point), maybe it can be called something else, e.g. `LocalModelConfig`. Then we're sort of saying we expect (or, _hope_) the global model is compatible with our "local model config", although it is not guaranteed as you mentioned in docstrings.

rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -120,6 +133,8 @@ pub unsafe extern "C" fn xaynet_ffi_participant_new(settings: *const Settings) -
 ///     model once the [`PARTICIPANT_SHOULD_SET_MODEL`] flag is set.
 ///   - [`PARTICIPANT_SHOULD_SET_MODEL`]: if set, then the participant should set its
 ///     model, by calling [`xaynet_ffi_participant_set_model()`]
+///   - [`PARTICIPANT_NEW_GLOBAL_MODEL`]: if set, the participant can fetch the new global
+///     model, by calling [`xaynet_ffi_participant_global_model()`]
```
finiteprods, line 137: this is fine, but noticed in line 109/110 can remove "tries" / "attempts" from the doc string. also maybe an unintended line break in line 145.

rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -273,14 +291,13 @@ pub unsafe extern "C" fn xaynet_ffi_participant_restore(
 ///
 /// - `buffer` should be a pointer to a buffer that contains the model
 /// - `data_type` specify the type of the model weights (see [`DataType`]). The C header
```
finiteprods, line 293:
```suggestion
/// - `data_type` specifies the type of the model weights (see [`DataType`]). The C header
```

rust/xaynet-mobile/src/ffi/participant.rs
```diff
@@ -345,3 +362,143 @@ pub unsafe extern "C" fn xaynet_ffi_participant_set_model(
...
+/// - `data_type` specifies the type of the model weights (see [`DataType`]). The C header
+///   file generated by this crate provides an enum corresponding to the parameters: `DataType`.
+/// - `len` is the number of weights the model has
+///
+/// # Return Value
+///
+/// - [`OK`] if the model is set successfully
+/// - [`ERR_NULLPTR`] if `participant` or the `buffer` is NULL
+/// - [`GLOBAL_MODEL_NONE`] if no model exists
+/// - [`ERR_GLOBALMODEL_IO`] if the communication with the coordinator
```
finiteprods, line 378:
```suggestion
/// - [`ERR_GLOBALMODEL_IO`] if the communication with the coordinator failed
```

rust/xaynet-mobile/tests/ffi_test.c
```diff
@@ -74,6 +74,34 @@ static char *test_settings() {
...
+  mu_assert("expected participant is null error", err == ERR_NULLPTR);
+
+  err = xaynet_ffi_participant_global_model(participant, NULL, model_config->data_type, model_config->len);
+  mu_assert("expected buffer is null error", err == ERR_NULLPTR);
+
+  err = xaynet_ffi_participant_global_model(participant, buffer, model_config->data_type, model_config->len);
+  mu_assert("expected io error (cannot connect to coordinator)", err == ERR_GLOBALMODEL_IO);
+
```
finiteprods, line 96: Q. beyond this, i guess testing the other error codes would be trickier?

finiteprods (commented):
looks nice, just some minor things I spotted!

rust/xaynet-mobile/xaynet_ffi.h
Robert-Steiner: Good point I changed it to `MODEL_DATATYPE_F32`

rust/xaynet-sdk/src/state_machine/state_machine.rs
```diff
@@ -65,6 +66,17 @@ impl StateMachine {
             StateMachine::Sum2(phase) => phase.state.into(),
         }
     }
+
+    /// Return the model configuration of the model that is expected in the update phase.
+    pub fn model_config(&self) -> ModelConfig {
```
Robert-Steiner, line 71: @little-dude @finiteprods 
What do you think about the idea that we pass the local model config to the user via the notifier (maybe in `IO::notify_load_model`) instead of a state machine method?    

The problem I see here with the state machine method is that the state machine can no longer simply be executed in a loop like:

```rust
        loop {
            self = match self.0.transition().await {
                TransitionOutcome::Pending(state_machine) => {
                    delay_for(tick).await;
                    Self(state_machine)
                }
                TransitionOutcome::Complete(state_machine) => Self(state_machine),
            };
        }
```

If the user wants to receive the model configuration, he has to interrupt the loop in order to call the method

But I would like to do this once we included the data type to the global model so that I don't depend on the local model for converting the global model

rust/xaynet-mobile/tests/ffi_test.c
Robert-Steiner: Like the idea of `LocalModelConfig` 👍 But if we use this name, I would not change the current signature of the function because it would be confusing if the function uses the local model configuration right?

rust/xaynet-mobile/tests/ffi_test.c
Robert-Steiner: yeah we need to have a running coordinator instance for this
I can add a task to the "write integration test" story

rust/xaynet-mobile/src/ffi/participant.rs
Robert-Steiner: > this is fine, but noticed in line 109/110 can remove "tries" / "attempts" from the doc string.
sorry I don't understand what you mean here

rust/xaynet-sdk/src/state_machine/phase.rs
```diff
@@ -260,6 +260,14 @@ impl<P> Phase<P> {
         .unwrap()
     }
 
+    /// Return the model configuration of the model that is expected in the update phase.
+    pub fn model_config(&self) -> ModelConfig {
+        ModelConfig {
+            data_type: self.state.shared.round_params.mask_config.vect.data_type,
```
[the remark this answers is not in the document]
Robert-Steiner: I have added a comment
I'll also create a ticket so we don't forget this

rust/xaynet-mobile/tests/ffi_test.c
Robert-Steiner: I just saw that there is already a task `Write tests for the C bindings of xaynet_mobile`

rust/xaynet-mobile/src/ffi/participant.rs
finiteprods:
```rust
/// Drive the participant internal state machine. Every tick, the state machine tries
/// attempts to perform a small work unit.
```
can either chop `tries` or `attempts`, that's all :)

rust/xaynet-mobile/tests/ffi_test.c
finiteprods: up to you. Passing a `LocalModelConfig` is a bit less brittle, but I understand that this name could be confusing too (but confusing in a _different_ way to if it was just called `ModelConfig`!). It's probably fine if it's documented well. Maybe naming the variable `expected_config` might also help express intent. just a couple of suggestions but happy either way.

rust/xaynet-mobile/src/ffi/participant.rs
Robert-Steiner: Ahh I see, I was looking at the wrong line. Thanks 👍

rust/xaynet-mobile/tests/ffi_test.c
Robert-Steiner: I changed the name of `ModelConfig` to `LocalModelConfig` but I havn't changed the the function signature. Is it fine if we keep it as it is and change it later?

rust/xaynet-mobile/src/ffi/participant.rs
Robert-Steiner: fixed

rust/xaynet-mobile/tests/ffi_test.c
finiteprods: of course, or you can leave as is if that is better for you.

finiteprods (approved):

PR #630, opened 2020-12-07
refactor metrics

Robert-Steiner:
**Summary**
- upgraded `influxdb` to `0.3`
	- faster dispatching of metrics
	- fixes escape issues of fields and tags
	- the timestamp is created by the coordinator and not when the metric has arrived on the InfluxDB instance
- up to 50 metrics can be sent concurrently (before they were sent sequentially)
- fixed the issue where the state machine emits a "handle message successfully" metric even though the message was rejected
- removed `xaynet-macros`
- remove metrics sender in state machine (the new metric recorder is now a global instance that can be accessed from any part of the server)
- influx dispatcher is implemented as a `tower` service

Notes:
The recorder API could be further abstracted (it exposes some InfluxDB types). However, I only see the use of InfluxDB as an intermediate solution. Even with batching, InfluxDB doesn't seem like a good candidate for recording metrics. It is more suitable for recording events. It would be better if the metrics were already aggregated locally and pulled regularly (like it is done in Prometheus) instead of sending each metric individually. But this is something we can decide later (there is already a story for that topic). Therefore, the metrics recorder is not suitable for recording high-frequency metrics (like counting http response codes in the http layer). It will in the end flood the InfluxDb Service.

The `metrics` feature flag still exists (it controls if metrics are recorded or not). I want to remove it in a separate PR because this requires some changes in the CI, config files, README, dockerfiles and coordinator settings.


rust/xaynet-server/src/metrics/recorders/influxdb/recorder.rs
```diff
@@ -0,0 +1,66 @@
...
+        let mut metric = Metric::new(measurement, value.into());
+        if let Some(tags) = tags {
+            metric.with_tags(tags);
+        }
+
+        self.call(Request::from(metric))
+    }
+
+    /// Records a new event and dispatches it to an InfluxDB instance.
+    pub fn event<'a, T>(&self, title: T, description: Option<&str>, tags: Option<&'a [&'a str]>)
```
little-dude, line 37: Why do we need to specify the lifetime here?

rust/xaynet-server/src/metrics/recorders/influxdb/recorder.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        self.call(Request::from(event))
+    }
+
+    fn call(&self, req: Request) {
+        let mut handle = self.service.0.clone();
+        tokio::spawn(async move {
+            if poll_fn(|cx| handle.poll_ready(cx)).await.is_err() {
+                error!("influx service failed")
```
little-dude, line 58: Maybe a more specific error like `error!("influx service temporarily unavailable {}", e)`?

rust/xaynet-server/src/metrics/recorders/influxdb/dispatcher.rs
```diff
@@ -0,0 +1,117 @@
...
+        Self { client }
+    }
+}
+
+impl Service<Request> for Dispatcher {
+    type Response = ();
+    type Error = anyhow::Error;
+    #[allow(clippy::type_complexity)]
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
```
little-dude, line 44: Not sure whether we want to use it but I discovered recently that `futures` provides an alias for this: https://docs.rs/futures/0.3.8/futures/future/type.BoxFuture.html

This could be `BoxFuture<'static, Result<Self::Response, Self::Error>>`

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -171,19 +169,21 @@ where
...
-            match store.number_of_unique_masks().await {
-                Ok(number_of_masks) => metrics_tx.send(metrics::masks::total_number::update(
-                    number_of_masks,
-                    round_id,
-                    phase_name,
-                )),
-                Err(err) => error!("failed to fetch total number of masks: {}", err),
-            };
-        });
+        if GlobalRecorder::global().is_some() {
```
little-dude, line 172: nit: to avoid one level of indentation we could do

```rust
if GlobalRecorder::global().is_none() {
    return;
}
```

rust/xaynet-server/src/metrics/recorders/influxdb/models.rs
```diff
@@ -0,0 +1,259 @@
...
+        Self {
+            name: measurement,
+            time: Utc::now(),
+            value,
+            tags: None,
+        }
+    }
+
+    pub(in crate::metrics) fn with_tags(&mut self, tags: Tags) {
+        // we don't want to extend tags. The user can't call that method
```
little-dude, line 89: What do you mean by "extending" the tags?

little-dude (approved):
Very clean implementation imo :+1:

rust/xaynet-server/src/metrics/recorders/influxdb/recorder.rs
Robert-Steiner: Ah you are right, we don't need that. It was a leftover from my previous attempts

rust/xaynet-server/src/metrics/recorders/influxdb/recorder.rs
Robert-Steiner: 👍 fixed

rust/xaynet-server/src/metrics/recorders/influxdb/dispatcher.rs
Robert-Steiner: > This could be BoxFuture<'static, Result<Self::Response, Self::Error>>

I like it, it's shorter and cleaner 👍

rust/xaynet-server/src/metrics/recorders/influxdb/models.rs
Robert-Steiner: With `metric!` It is not possible to call the function several times. Therefore, I did't cover the case where we would extend `self.tags` when `self.tags` already contains tags.
I updated the comment

PR #640, opened 2021-01-04
Refactor Rust CI

Robert-Steiner:
**Summary**
- split cargo registry from specific job build artifacts (check, test, clippy, docs)
- registry is fetched only once and is shared by all other cargo jobs
- use of `restore-keys`
	- if a cache was not found reuse an older cache
	- all caches, however, are invalidated after a week (to keep the size of a cache small)
- only runs  `cargo test`, `clippy` and `docs` if `cargo check` has been passed beforehand
- cache `cargo-readme` binary instead of compiling it in every CI run (last release was 9 month ago so i don't think that we need to invalidate the cache)
- deleted docker readme sync job 
	- does not work because the docker hub api that is used for that does not support personal access tokens (we would need to turn off 2fa to make it work👎)


wilk10 (approved):
LGTM! You're on fire lately. Thanks a lot

PR #641, opened 2021-01-04
XN-1478 max participant settings

janpetschexain:
**References**

- [XN-1478](https://xainag.atlassian.net/browse/XN-1478)

**Summary**

settings:
- add max participant settings to `xaynet-server`
- update the config files with default values
- update settings validation

state machine:
- check the settings constraints in the `sum`,  `update` and `sum2` phases
- update state logging
- cleanup code: ~i tried to remove some indirections in the code by more prominently utilizing the `handle_request()` impl and to call metric updates for the messages in one place instead of scattered places.~ unified common `handle_request()` code of the sum/update/sum2 phases in the `process_during()` and `process_until_enough()` impls


rust/xaynet-server/src/state_machine/phases/sum.rs
```diff
@@ -40,18 +45,42 @@ where
...
+            ok @ Ok(_) => {
+                self.private.accepted += 1;
+                self.increment_message_metric(Measurement::MessageSum);
+                ok
+            }
+            error @ Err(_) => {
+                self.private.rejected += 1;
+                self.increment_message_metric(Measurement::MessageRejected);
+                error
+            }
```
Robert-Steiner, line 83: I think we can move most of the code into `process_single`.

```rust
async fn handle_request(&mut self, req: StateMachineRequest) -> Result<(), RequestError> {
        // reject if not a `sum` message
        let SumRequest {
            participant_pk,
            ephm_pk,
        } = if let StateMachineRequest::Sum(req) = req {
            req
        } else {
            return Err(RequestError::MessageRejected);
        };

        // accept if processed successfully, otherwise reject
        let res = self.update_sum_dict(participant_pk, ephm_pk).await;
        if res.is_ok() {
            self.private.accepted += 1;
            self.increment_message_metric();
        }
        res
}
```
```rust
#[async_trait]
pub trait Handler {
    /// Handles a request.
    async fn handle_request(&mut self, req: StateMachineRequest) -> Result<(), RequestError>;
    async fn has_enough_messages(&self) -> bool;
    async fn increment_discarded(&mut self);
    async fn increment_rejected(&mut self);
}
```
```rust
#[async_trait]
impl<C, M> Handler for PhaseState<Sum, C, M>
where
    C: CoordinatorStorage,
    M: ModelStorage,
{
    async fn handle_request(&mut self, req: StateMachineRequest) -> Result<(), RequestError> {
    // ...
    }

    fn has_enough_messages(&self) -> bool {
        self.private.accepted >= self.shared.state.max_sum_count
    }
    
    fn increment_discarded(&mut self) {
         self.private.discarded += 1;
    }

    fn increment_rejected(&mut self) -> bool {
        self.private.rejected += 1;
    }
}
```
```rust
async fn process_single(
        &mut self,
        req: StateMachineRequest,
        span: Span,
        resp_tx: ResponseSender,
    ) {
        let _span_guard = span.enter();
        let res = if self.has_enough_messages() {
            Err(RequestError::TooMany)
        } else {
           self.handle_request(req).await
        }

        if let Err(ref err) = res {
            warn!("failed to handle message: {}", err);
            if matches!(RequestError:: TooMany, _err) {
                self.increment_discarded();
                metric!(
                    Measurement::MessageDiscarded,
                    1,
                    ("round_id", self.shared.state.round_id),
                    ("phase", Self::NAME as u8)
                );
            } else {
                self.increment_rejected();
                metric!(
                    Measurement::MessageRejected,
                    1,
                    ("round_id", self.shared.state.round_id),
                    ("phase", Self::NAME as u8)
                );
            }
        }

        // This may error out if the receiver has already been dropped but
        // it doesn't matter for us.
        let _ = resp_tx.send(res);
    }
```
haven't tested the code though, it may not work that way

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -187,3 +180,14 @@ where
         });
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
Robert-Steiner, line 186: 👍

rust/xaynet-server/src/state_machine/phases/sum.rs
```diff
@@ -27,11 +28,15 @@ pub enum SumStateError {
...
 }
 
-/// Sum state
+/// The sum state.
 #[derive(Debug)]
 pub struct Sum {
-    /// The number of Sum messages successfully processed.
-    sum_count: u64,
+    /// The number of sum messages successfully processed.
+    accepted: u64,
```
Robert-Steiner, line 35: Like the renaming of `sum_count` 👍

Robert-Steiner (commented):
Looks great!👍
Good to see that no major changes were needed to implement the feature. I would have thought the changes would be bigger. 😁 I've added some suggestions to reduce code duplication a bit

configs/docker-dev.toml
```diff
@@ -10,6 +10,8 @@ tls_key = "/app/ssl/tls.key"
 [pet]
 min_sum_count = 1
 min_update_count = 3
+max_sum_count = 100
+max_update_count = 10000
```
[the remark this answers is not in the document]
janpetschexain: good catch 👍

rust/xaynet-server/src/settings/mod.rs
```diff
@@ -84,7 +86,7 @@ pub struct PetSettings {
     /// be greater or equal to `1` (i.e. `min_sum_count >= 1`), otherwise the PET protocol will be
     /// broken.
     ///
-    /// This parameter should only be used to enforce security constraints.
+    /// This parameter should only be used to enforce security constraints. Defaults to 1.
```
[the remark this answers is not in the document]
janpetschexain: mhh, i see what you mean. that was more related to the reference config file, but that's not really helpful in the end. i will remove the "defaults to" comments, since it effectively has no defaults.

we can think about adding defaults for everything later on in a separate story if needed.

rust/xaynet-server/src/settings/mod.rs
```diff
@@ -267,36 +314,55 @@ impl Default for PetSettings {
     }
 }
 
-/// Checks PET settings.
-fn validate_pet(s: &PetSettings) -> Result<(), ValidationError> {
-    validate_phase_times(s)?;
-    validate_fractions(s)
-}
+impl PetSettings {
```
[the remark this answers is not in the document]
janpetschexain: well, technically the `#[derive(Validate)]` does exactly that, it implements validation functions on the struct. so imo it doesn't matter if we implement additional validation methods on the struct next to the derived ones.

i added them as methods to make it more natural to call them on the structs, for example in the tests. unfortunately, the `validator_derive` crate can't deal with methods in the proc macro, but must get a function for custom validation, hence we need the wrapper anyways.

tbh i'm undecided, regarding above's points, should we keep them as methods or revert them to functions?

rust/xaynet-server/src/state_machine/phases/sum.rs
janpetschexain: that's a good idea, i'll give that a try. even though it scatters the metrics calls, it reduced code duplication and codifies things in a common interface, which outweighs the former issue.

rust/xaynet-server/src/settings/mod.rs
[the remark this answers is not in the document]
janpetschexain: we have a `Default` impl for the struct, but it is only used in tests currently. i will move that behind a `#[cfg(test)]` for now. if we want to provide a proper impl of default values we should mark the respective fields as `Option<T>` and then replace `None` values with `Some(default)` values after validation.

rust/xaynet-server/src/settings/mod.rs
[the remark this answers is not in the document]
Robert-Steiner: > well, technically the #[derive(Validate)] does exactly that

that's true
then let's leave the methods as they are

rust/xaynet-server/src/state_machine/phases/sum.rs
janpetschexain: i agree that we have to change the grafana stuff first, i left a todo in the code at the appropriate place. for now we can just keep the additional methods on the trait and try to simplify it later on.

i also took the opportunity to remove more code duplication wrt to `process_until_enough()`, this is now similar to `process_during()`. as a nice side effect of unifying that we now also have more debug level logging.

regarding your last point, that probably makes sense. we should keep that in mind when we work on the parameter adjustment stategy.

overall it would be good if you have another look at my latest changes, since they are not minor.

Robert-Steiner: on it👀

rust/Cargo.lock
```diff
@@ -484,22 +491,14 @@ dependencies = [
...
 [[package]]
 name = "combine"
-version = "4.4.0"
+version = "4.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9417a0c314565e2abffaece67e95a8cb51f9238cd39f3764d9dfdf09e72b20c"
+checksum = "01199925a18b00193570e3d70cfe57dcb647eb167c29851983e76fc39e2fee39"
 dependencies = [
- "bytes",
+ "bytes 0.5.6",
```
Robert-Steiner, line 500: I'm currently testing the changes on k8s and get some really weird redis errors (NoSeedDict, LengthMisMatch). I wonder if it comes from the fact that we have two versions of `bytes` in `combine` which is used by `redis`

janpetschexain: we should definitely rule out version conflicts first.

Robert-Steiner: I think it is the same issue https://github.com/mitsuhiko/redis-rs/issues/429
although in our case it compiles and only fails at runtime, which is scary

Robert-Steiner: Hm I might have been wrong with `bytes`. It seems that there is an issue in k8s that only occurs once we create (not restart) a coordinator pod. The bug itself is really strange. redis does return `None` when fetching the dictionaries after a sum or update phase, although the data exists in redis. That error has never happened before. I need to take a look on this later. The error also occurs with the code from master so I think it is safe to merge your pr. 

I think we can also revert the last commit if you like.

Robert-Steiner (approved):
Good job! 🚀
the refactoring is really nice 👍

rust/Cargo.lock
Robert-Steiner: Ok the errors i got are actually related to k8s. When re-creating a pod, it can happen that two coordinators exist at the same time and both then access redis

janpetschexain: mhh ok, strange error, but good catch! i'll revert the last commit.

PR #642, opened 2021-01-05
Clean up c ffi

Robert-Steiner:
just noticed that we have some doc warnings in `xaynet-mobile/ffi`

**Summary**
- fixed doc warnings
- added `#![cfg_attr(doc, forbid(warnings))]` to `lib.rs` so that  the ci fails next time
- removed some unsused error codes
- rebuilt header file with cbindgen 0.16.0


wilk10 (approved):
Thanks a lot! LGTM

PR #643, opened 2021-01-07
XN-1318 sdk clippy warnings

janpetschexain:
**References**

- [XN-1318](https://xainag.atlassian.net/browse/XN-1318) (addresses [#578](https://github.com/xaynetwork/xaynet/pull/578#discussion_r516009877))

**Summary**

*documentation regression:*

while working on the actual issue i noticed regression in the documentation which is not yet caught by the ci. hence:
- add doc check for all docs including private docs
- fix broken doc links

*clippy warnings for large enum variants:*

some of the state machine enum variants are quite large (up to 500-800 bytes) and some of the enums differ between the variants (200 vs 800 bytes), which caused the clippy lint. just boxing the variants leads to some negative side effects:
- boxing of subsequent fields of the variant to avoid unboxing/reboxing for our various from/into impls, which also spirals down to all fields of the hierarchically constructed structs
- adjusting many function signatures from `self` to `self: Box<Self>` which makes it less nice to read
- just boxing the variants related to `Update` would leave us with the same problems for `Sum`/`Sum2`

but working from the root cause upwards, it only requires us to box the fields of `State<P>`, then all enums/structs building on top of this are effectively box-sized as well without extra box annotations. for `Sum`/`Update`/`Sum` the `State<P>` is reasonably large to justify boxing and for the zero-sized `Awaiting`/`NewRound` boxing is a no-op.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -278,7 +278,7 @@ jobs:
 
       - name: Check the building of docs
         working-directory: ./rust
-        run: cargo doc --all-features --no-deps --color always
+        run: cargo doc --all-features --document-private-items --no-deps --color always
 
   coverage:
     name: cargo-tarpaulin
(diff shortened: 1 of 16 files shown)


Robert-Steiner (approved):
Looks great
I like the elegant solution 👍

How did you actually determine the size of the variants?

janpetschexain:
> How did you actually determine the size of the variants?

i wrote some temporary unit tests to print `std::mem::size_of::<T>()` for the whole enum as well as each inner part of the variants.

PR #646, opened 2021-01-11
Move docker latest tag to release job

Robert-Steiner:
- moved `latest` tag to release job
- upgraded `build-push-action` to v2
- added `bindings` to `.dockerignore`


.github/workflows/dockerhub-master.yml
```diff
@@ -12,19 +12,24 @@ jobs:
...
+        id: docker
+        with:
+          context: .
+          file: docker/Dockerfile
+          tags: xaynetwork/xaynet:development
           push: true
-          build_args: COORDINATOR_FEATURES=metrics
+          build-args: COORDINATOR_FEATURES=metrics
 
       - name: Install doctl
```
rsaffi, line 34: This will soon change as we'll move away from DO in favor of IONOS, but we change this when we get to that, no worries. Just mentioning.

rsaffi (approved):

PR #648, opened 2021-01-11
XN-1493 temp change warp to released version

janpetschexain:
**References**

- [XN-1493](https://xainag.atlassian.net/browse/XN-1493)

**Summary**

temporarily change to a released version of `warp` and therefore disable server side client authentication.


Robert-Steiner (approved):
LGTM👍
maybe we should hide the docs of the `tls_client_auth` setting

janpetschexain:
> LGTM👍
> maybe we should hide the docs of the `tls_client_auth` setting

good point, done.

PR #662, opened 2021-01-12
add python migration guide

Robert-Steiner:
- added python migration guide ([rendered version](https://github.com/xaynetwork/xaynet/blob/python-migration-guide/bindings/python/migration_guide.md))
- fixed type issues in the example
- added a note that Python 3.7.1 or higher (because of pandas) is required to run the example

### DIFF
--- a/bindings/python/migration_guide.md
+++ b/bindings/python/migration_guide.md
@@ -1,0 +1,114 @@
+# Migration from `v0.8.0` to `v.0.11.0`
+
+To demonstrate the API changes from `v0.8.0` to `v.0.11.0`, we will use the keras example
+which is available in both versions. For reasons of clarity, some parts of the code have
+been removed.
+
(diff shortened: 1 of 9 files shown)


bindings/python/migration_guide.md
```diff
@@ -0,0 +1,114 @@
...
+
+def main() -> None:
+    participant = Participant(args.data_directory)
+
+    run_participant(
+        participant, args.coordinator_url, heartbeat_period=args.heartbeat_period
+    )
+```
+
+## `v0.11.0`
```
janpetschexain, line 51: maybe a link similar to the v0.8 header

```suggestion
## [`v0.11.0`](https://github.com/xaynetwork/xaynet/blob/v0.10.0/bindings/python/xaynet_sdk/participant.py)
```

the link will be broken until we release v0.11, but that shouldn't matter much.

janpetschexain (approved):
nice guide 👍

PR #669, opened 2021-01-13
set max message size to unlimited

Robert-Steiner:
Small pr that sets the `MaxMessageSize` to `unlimited` in the python participant.
The python bindings are primarily there to experiment with fed ml/secure aggragtion. 
So there is not need to split these messages. 
Does not fix `XN-1503` because we still need to expose this setting to the C API 
and it would be nice to make it configurable in the python participant

### DIFF
--- a/bindings/python/Cargo.lock
+++ b/bindings/python/Cargo.lock
@@ -1594,6 +1594,7 @@ dependencies = [
  "tracing-subscriber",
  "xaynet-core",
  "xaynet-mobile",
+ "xaynet-sdk",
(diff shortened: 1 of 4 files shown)


janpetschexain (approved):

little-dude:
Is there an issue with multipart messages? :S

Robert-Steiner:
> Is there an issue with multipart messages? :S

No works perfectly fine 😊
But in the python client we set this to `unlimited` so that the message can be sent faster

PR #672, opened 2021-01-14
update docs

Robert-Steiner:
- replaced xayn logo with xaynet logo
- removed xayn logo
- replaced internal doc html links with crate links
- enabled docs for all features
- added feature flag hints via 

```rust
#[cfg_attr(docsrs, doc(cfg(feature = "tls")))]
```

Screenshot_20210114_115348
Screenshot_20210114_115326


rust/xaynet-core/Cargo.toml
```diff
@@ -11,6 +11,10 @@ license-file = "../../LICENSE"
 keywords = ["federated-learning", "fl", "ai", "machine-learning"]
 categories = ["science", "cryptography"]
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
janpetschexain, line 16: do you know if we can have a single `[workspace.metadata.docs.rs]` in the workspace's `Cargo.toml` instead of in each package? i think `rustdoc` will behave similarly as `cargo`, so this might work.

Robert-Steiner: not sure 
but I think this is not possible because `docsrs` creates the documentation for each `crate` separately

Robert-Steiner: i check if I can build the workspace with `dorcsrs`

Robert-Steiner: > i check if I can build the workspace with dorcsrs

unfortunately doesn't seem to work

janpetschexain: i meant this differently: let's say we delete the `[package.metadata.docs.rs]` tables from the packages' `Cargo.toml`s and add a `[workspace.metadata.docs.rs]` table in the workspace's `Cargo.toml`. if we then call `cargo run ...` for `docs.rs` on a package, for example `xaynet-core` it should try to look up the metadata in the package's toml and since it doesn't find it, it should then look in the workspace's toml.

Robert-Steiner: ahh I see
I will check

Robert-Steiner: I checked but it does not seem to work (at least when building the docs locally)
I think `docsrs` downloads the crate from crates.io and therefore it does not have access to the workspace file 
but I could be wrong

janpetschexain: ok, then we have to keep the metadata in every package.

janpetschexain (approved):

PR #673, opened 2021-01-14
add data combiner unit tests

wilk10:
Refers to: [XN-1433](https://xainag.atlassian.net/browse/XN-1433)

---

It changes a bit `DataCombiner` so that it's easier to test, and adds relevant unit tests.

Since the initialisation of `DataPoints` has a common interfaces and is quite repetitive, but the implementation differs, maybe it would make sense in the future to split up this component and have one `DataCombiner` for each type of `DataPoint`, with its own configuration (metadata instances above all).

Anyway, for now i guess this is good enough, and now it has its unit tests. 

---

Follow up: [XN-1542](https://xainag.atlassian.net/browse/XN-1542) -> unit tests for each `DataPoint` type.

### DIFF
--- a/rust/xaynet-analytics/src/data_combination/data_combiner.rs
+++ b/rust/xaynet-analytics/src/data_combination/data_combiner.rs
@@ -173,10 +173,10 @@ where
         metadata: DataPointMetadata,
         n_periods_override: Option<u32>,
     ) -> DateTime<Utc> {
-        let n_periods = if n_periods_override == None {
-            metadata.period.n
+        let n_periods = if let Some(n_periods) = n_periods_override {
+            n_periods
         } else {
-            n_periods_override.unwrap()
+            metadata.period.n
         };
         let avg_days_per_month = 365.0 / 12.0;
         let midnight_end_of_period = get_midnight(metadata.end);
@@ -262,15 +262,15 @@ mod tests {
         let metadata = DataPointMetadata::new(Period::new(PeriodUnit::Days, 1), end_period);
         let screen_route = "home_screen".to_string();
         let first_event = AnalyticsEvent::new(
-            "test1".to_string(),
+            "test1",
             AnalyticsEventType::ScreenEnter,
             end_period - Duration::hours(12),
-            Some(screen_route.clone()),
+            screen_route.clone(),
         );
         let all_events = vec![
             first_event.clone(),
             AnalyticsEvent::new(
-                "test1".to_string(),
+                "test1",
                 AnalyticsEventType::AppEvent,
                 end_period - Duration::hours(13),
                 None,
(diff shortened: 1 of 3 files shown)


rust/xaynet-analytics/src/data_provision/analytics_event.rs
```diff
@@ -20,13 +20,14 @@ impl AnalyticsEvent {
     pub fn new(
```
Robert-Steiner, line 20: Two little tricks that make using the API a bit easier `<N: Into<String>, R: Into<Option<String>>>`

```rust
impl AnalyticsEvent {
    pub fn new<N: Into<String>, R: Into<Option<String>>>(
        name: N,
        event_type: AnalyticsEventType,
        timestamp: DateTime<Utc>,
        screen_route: R,
    ) -> AnalyticsEvent {
        AnalyticsEvent {
            name: name.into(),
            event_type,
            timestamp,
            screen_route: screen_route.into(),
        }
    }
}
```

This way we can pass both `&str` as well as `String` in the case of `name` .
For `screen_route` we can omit the keyword `Some`.

```rust
        AnalyticsEvent::new(
            "test2".to_string(),
            AnalyticsEventType::ScreenEnter,
            timestamp,
            "other_screen".to_string(),
        );

        AnalyticsEvent::new(
            "test2",
            AnalyticsEventType::ScreenEnter,
            timestamp,
            None,
        );
```

Just a suggestion you don't have to implement this

rust/xaynet-analytics/src/data_combination/data_combiner.rs
```diff
@@ -128,14 +161,23 @@ where
...
+    fn get_start_of_period(
+        &self,
+        metadata: DataPointMetadata,
+        n_periods_override: Option<u32>,
+    ) -> DateTime<Utc> {
+        let n_periods = if n_periods_override == None {
+            metadata.period.n
+        } else {
+            n_periods_override.unwrap()
+        };
```
Robert-Steiner, line 180: this can be rewritten a bit and thus avoid the `unwrap`

```suggestion
        let n_periods = if let Some(n_periods) = n_periods_override {
            n_periods
        } else {
            metadata.period.n
        };
```

Robert-Steiner (approved):
Looks good

rust/xaynet-analytics/src/data_provision/analytics_event.rs
wilk10: will do! looks good. it's the kind of stuff i still need to learn. thanks a lot

rust/xaynet-analytics/src/data_combination/data_combiner.rs
wilk10: thanks! will do

PR #636, opened 2020-12-17
port iota integration into xaynet os repo

Robert-Steiner:

PR #675, opened 2021-01-15
fix docs tests

Robert-Steiner:
If `cargo test` is called with the `--all-target` flag, the doc tests will not be executed. (related https://github.com/rust-lang/cargo/issues/6669)

Fixes two docs tests. I've already spent some time fixing them in #636 but I didn't get it to work so I just annotated them with `ignore`.


janpetschexain (approved):

janpetschexain:
that's indeed a very weird behavior that the `--doc` target is not included in `--all-targets`.

PR #677, opened 2021-01-18
add data points unit tests

wilk10:
Refers to: [XN-1542](https://xainag.atlassian.net/browse/XN-1542)

### DIFF
--- a/rust/xaynet-analytics/src/data_combination/data_points/was_active_each_past_period.rs
+++ b/rust/xaynet-analytics/src/data_combination/data_points/was_active_each_past_period.rs
@@ -1,5 +1,5 @@
 use chrono::{DateTime, Utc};
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 use crate::{
     data_combination::data_points::data_point::{
@@ -24,8 +24,8 @@ impl CalcWasActiveEachPastPeriod {
     }
 
     // TODO: this could possibly later be optimised by sorting events by timestamp and caching the last timestamp added to the HashMap
-    fn group_timestamps_by_period_threshold(&self) -> HashMap<DateTime<Utc>, Vec<DateTime<Utc>>> {
-        let mut timestamps_by_period_threshold = HashMap::new();
+    fn group_timestamps_by_period_threshold(&self) -> BTreeMap<DateTime<Utc>, Vec<DateTime<Utc>>> {
+        let mut timestamps_by_period_threshold = BTreeMap::new();
         for these_thresholds in self.period_thresholds.windows(2) {
             // safe unwrap: `windows` guarantees that there are at least two elements.
             // If `period_thresholds` contains less than two elements, this code block is not executed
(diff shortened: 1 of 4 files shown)


rust/xaynet-analytics/src/data_combination/data_points/was_active_each_past_period.rs
```diff
@@ -52,9 +52,153 @@ impl CalculateDataPoints for CalcWasActiveEachPastPeriod {
...
+                None,
+            ),
+            AnalyticsEvent::new(
+                "test2",
+                AnalyticsEventType::Error,
+                end_period - Duration::hours(15),
+                None,
+            ),
+        ];
+        // let period_thresholds = vec![end_period, end_period - Duration::days(1), end_period - Duration::days(2)];
```
Robert-Steiner, line 164: can this be removed ?

rust/xaynet-analytics/src/data_combination/data_points/was_active_each_past_period.rs
```diff
@@ -52,9 +52,153 @@ impl CalculateDataPoints for CalcWasActiveEachPastPeriod {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use chrono::{DateTime, Duration, Utc};
+
+    use crate::{
+        data_combination::data_points::{
+            data_point::{CalculateDataPoints, DataPointMetadata, Period, PeriodUnit},
+            was_active_each_past_period::CalcWasActiveEachPastPeriod,
```
Robert-Steiner, line 71: you can replace some imports with `use super::*` which imports everything that is defined in `mod was_active_each_past_period;`

Robert-Steiner (approved):
Really nice!

rust/xaynet-analytics/src/data_combination/data_points/was_active_each_past_period.rs
wilk10: ah shute, missed it!

rust/xaynet-analytics/src/data_combination/data_points/was_active_each_past_period.rs
wilk10: i wasn't sure what's best practice with Rust, whether to be more explicit or more practical. i'll use more `use super::*` then! thanks

Robert-Steiner: For the `tests` mod, it is fine to use` super :: * `because you usually want to test what is in the parent module

janpetschexain: agree, in general more explicit is always better and i personally tend to only import explicitly, even for things like `prelude`s. of course it is more convenient to import everything implicitly, but it is way easier to read for other if imported explicitly.
that said, the only exception i make is for `super::*` in tests, because the tests refer directly to the code and are usually even written in the same file, which leaves very little room for ambiguity and hidden things.

wilk10: sounds good, totally agree

PR #682, opened 2021-01-20
impl trust anchor trait

Robert-Steiner:
Cherry-pick trust anchor trait + implementation into the state machine from #636
- applied review comments


janpetschexain (approved):
the reduction of generic type parameters to be carried around is pretty nice 👍

PR #683, opened 2021-01-20
XN-1556 split pet settings

janpetschexain:
**References**

- [XN-1556](https://xainag.atlassian.net/browse/XN-1556)

**Summary**

- split the pet settings per phase
- update the coordinator state and the state machine accordingly
- update tests and examples


configs/config.toml
```diff
@@ -7,21 +7,35 @@ tls_certificate = "/app/ssl/tls.pem"
...
+min = 10
+max = 3600
+
+[pet.sum2.count]
+min = 1
+max = 100
+
+[pet.sum2.time]
+min = 5
+max = 3600
```
Robert-Steiner, line 38: I would prefer it if it is not so verbose

```suggestion
[pet.sum]
prob = 0.5
count = { min = 1, max = 100 }
time = { min = 5, max = 3600 }

[pet.update]
prob = 0.9
count = { min = 3, max = 10000 }
time = { min = 10, max = 3600 }

[pet.sum2]
count = { min = 1, max = 100 }
time = { min = 5, max = 3600 }
```

rust/xaynet-server/src/state_machine/coordinator.rs
```diff
@@ -18,30 +98,12 @@ pub struct CoordinatorState {
...
-    pub min_update_time: u64,
-    /// The minimum time (in seconds) reserved for processing sum2 messages.
-    pub min_sum2_time: u64,
-    /// The maximum time (in seconds) permitted for processing sum messages.
-    pub max_sum_time: u64,
-    /// The maximum time (in seconds) permitted for processing update messages.
-    pub max_update_time: u64,
-    /// The maximum time (in seconds) permitted for processing sum2 messages.
-    pub max_sum2_time: u64,
+    /// The sum phase parameters.
```
Robert-Steiner, line 101: 👍

Robert-Steiner (approved):
Nice clean up 👍
really like the simplification of the `CoordinatorState` struct

configs/config.toml
janpetschexain: i didn't know that this syntax is possible, that definitely ways more concise 👍 i'll change that.

PR #700, opened 2021-01-29
XN-1522 enable client auth

janpetschexain:
**References**

- [XN-1522](https://xainag.atlassian.net/browse/XN-1522)

**Summary**

- activates server side client auth again, with stable dependency


Robert-Steiner (approved):
we should add this to the changelog too? 
I think it's a nice feature

janpetschexain:
> we should add this to the changelog too?
> I think it's a nice feature

true, i'll add that.

PR #707, opened 2021-02-02
XN-1582 simplify metrics

janpetschexain:
**References**

- [XN-1582](https://xainag.atlassian.net/browse/XN-1582)

**Summary**

- simplify message accepted metrics to one call differentiated by phase tags
- make most of the metrics function signatures more ergonomic to use via `Into` trait bounds
- replace custom metric/request/query conversions with `From` impls

### DIFF
--- a/rust/xaynet-server/src/metrics/recorders/influxdb/models.rs
+++ b/rust/xaynet-server/src/metrics/recorders/influxdb/models.rs
@@ -1,5 +1,7 @@
+use std::{borrow::Borrow, iter::IntoIterator};
+
 use chrono::{DateTime, Utc};
-use influxdb::InfluxDbWriteable;
+use influxdb::{InfluxDbWriteable, Timestamp, Type, WriteQuery};
 
 /// An enum that contains all supported measurements.
 pub enum Measurement {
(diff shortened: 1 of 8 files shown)


rust/xaynet-server/src/metrics/recorders/influxdb/models.rs
```diff
@@ -128,28 +133,40 @@ impl Event {
...
+        self.description = description.into().map(Into::into);
+        self
     }
 
-    pub(in crate::metrics) fn with_tags(&mut self, tags: &[&str]) {
+    pub(in crate::metrics) fn with_tags<T, A, B>(mut self, tags: T) -> Self
+    where
+        T: Into<Option<A>>,
+        A: AsRef<[B]>,
+        B: Borrow<str>,
```
Robert-Steiner, line 149: why do we need the `Borrow` bound here?

Robert-Steiner (approved):
Looks good 👍
nice work on the metrics module 🚀

rust/xaynet-server/src/metrics/recorders/influxdb/models.rs
janpetschexain: that is required by `slice::join()`

PR #708, opened 2021-02-02
increase test coverage state machine

Robert-Steiner:
**Summary**
- reworked state machine test utils
  - the biggest problem with the `StateMachineBuilder` was that the state machine was created with the `from` instead of the `new` method. Sometimes the `new` method contains logic (like updating the round id) that is then never executed 
- converted the integration tests into unit tests (pet messages and the storage are mocked)
- tests can be run without dependencies like Redis
- one test for testing the `message_purging` is currenlty commented out until we fixed the issues with the `message_purging`
- added tests to test the default settings (PET, mask, model) that are used in the state machine test. 
- i'm sure that some tests could be made smaller with macros, but I don't see the need for it right now


rust/xaynet-server/src/state_machine/phases/idle.rs
```diff
@@ -175,80 +175,279 @@ where
...
+        (state, event_publisher, event_subscriber)
+    }
 
-        let id = keys.get_latest().round_id;
-        assert_eq!(id, 1);
+    fn assert_params(params1: &RoundParameters, params2: &RoundParameters) {
+        assert_ne!(params1.pk, params2.pk);
+        assert_ne!(params1.seed, params2.seed);
+        assert!((params1.sum - params2.sum).abs() < f64::EPSILON);
+        assert!((params1.update - params2.update).abs() < f64::EPSILON);
```
janpetschexain, line 218: checking for `< EPSILON` essentially means `== 0` (not exactly true, but close enough), with `<= EPSILON` we should be on the safe side.
```suggestion
        assert!((params1.sum - params2.sum).abs() <= f64::EPSILON);
        assert!((params1.update - params2.update).abs() <= f64::EPSILON);
```

similarly below.

rust/xaynet-server/src/state_machine/phases/sum.rs
```diff
@@ -127,86 +127,577 @@ where
...
+    }
+
+    #[tokio::test]
+    async fn test_rejected_messages() {
+        // No Storage errors
+        //
+        // What should happen:
+        // 1. broadcast Sum phase
+        // 2. accept 7 sum messages
+        // 3. reject 5 update and 2 sum2 messages
```
janpetschexain, line 319:
```suggestion
        // 3. reject 3 update and 5 sum2 messages
```

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -223,9 +223,570 @@ where
...
+                utils::{create_global_model, create_mask},
+                MockCoordinatorStore,
+                MockModelStore,
+                MockTrustAnchor,
+            },
+            Store,
+        },
+    };
+
+    fn events_from_sum_phase(state: &CoordinatorState) -> (EventPublisher, EventSubscriber) {
```
janpetschexain, line 252:
```suggestion
    fn events_from_sum2_phase(state: &CoordinatorState) -> (EventPublisher, EventSubscriber) {
```

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -223,9 +223,570 @@ where
...
+        EventBusBuilder::new(state)
+            .broadcast_phase(PhaseName::Sum2)
+            .broadcast_sum_dict(DictionaryUpdate::New(Arc::new(SumDict::new())))
+            .broadcast_seed_dict(DictionaryUpdate::New(Arc::new(SeedDict::new())))
```
janpetschexain, line 256: i think this should be `Invalidate` since we moved that to the sum2 phase.
```suggestion
            .broadcast_sum_dict(DictionaryUpdate::Invalidate)
            .broadcast_seed_dict(DictionaryUpdate::Invalidate)
```

rust/xaynet-server/src/state_machine/phases/update.rs
```diff
@@ -186,148 +186,370 @@ where
...
+
+        let (event_publisher, event_subscriber) = events_from_sum_phase(&state);
+        let events_before_update = EventSnapshot::from(&event_subscriber);
+        let state_before_update = state.clone();
+
+        let (shared, request_tx) = init_shared(state, store, event_publisher);
+        let state_machine = StateMachine::from(PhaseState::<Update, _>::new(shared));
+        assert!(state_machine.is_update());
+
+        send_update_messages_with_model(3, request_tx.clone(), create_mask(2, 1));
```
janpetschexain, line 479: even though calling this function with a mask instead of a masked model works because both are of `MaskObject` type, this is just another reason to finally have distinct wrapper types for masks and masked models. we should address this in a future refactoring.

janpetschexain (approved):
very nice and comprehensive test refactoring 👍 

> - i'm sure that some tests could be made smaller with macros, but I don't see the need for it right now

i see what you did there 😁 😜

rust/xaynet-server/src/state_machine/phases/unmask.rs
Robert-Steiner: good catch👍

rust/xaynet-server/src/state_machine/phases/failure.rs
```diff
@@ -114,40 +114,283 @@ where
...
-                error: PhaseError::RequestChannel(""),
-            })
+    fn state_and_events_from_sum2_phase() -> (CoordinatorState, EventPublisher, EventSubscriber) {
+        let state = CoordinatorStateBuilder::new().build();
+
+        let (event_publisher, event_subscriber) = EventBusBuilder::new(&state)
+            .broadcast_phase(PhaseName::Sum2)
+            .broadcast_sum_dict(DictionaryUpdate::New(Arc::new(SumDict::new())))
+            .broadcast_seed_dict(DictionaryUpdate::New(Arc::new(SeedDict::new())))
+            .broadcast_model(ModelUpdate::New(Arc::new(create_global_model(10))))
```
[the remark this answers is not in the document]
Robert-Steiner: It doesn't lead to conflicts in the state machine but it would lead to errors in the client. I changed it anyway to be consistent

rust/xaynet-server/src/state_machine/phases/update.rs
Robert-Steiner: I agree👍 Do we have a ticket/story for that?

janpetschexain: i don't think that we already have a ticket, can you create one please?

Robert-Steiner: I can do that 👍

Robert-Steiner:
@janpetschexain thanks again for the review😊 
I hope I addressed all comments (all changes are in the last commit)

PR #712, opened 2021-02-03
first class scalar

finiteprods:
The purpose of this PR is two-fold:
- Adds a new safe API for working with model _scalars_.
- Fixes the "floating point bug" affecting the `test_masking_and_aggregation!` tests, discovered in #542.

Currently, scalars are represented as `f64` floats. But since scalars are (since #496) an integral part of the (un)masking process, they should be represented similarly to model weights, namely in "exact" form as `Ratio`s of big integers.

Indeed, the bug mentioned above arises from the inexact nature of scalars: in the `test_masking_and_aggregation!` tests, we take scalars to be a fraction (morally `1 / model_count`) which as per the usual limitations of floating point, cannot be represented _exactly_ as `f64` in many cases. Subsequently, the error this introduces propagates through the (un)masking process, and leads to failing closeness checks. #542 introduced a workaround by essentially adjusting `model_count` to a known "good" value.

OTOH by working with `f64` scalars, it is easy for clients to construct imprecise (or worse, invalid e.g. negative, NaN) scalars, whether accidentally or maliciously. This could be prevented by insisting that scalars be constructed (or converted) via a safe API that statically guarantees their validity, similar to that of `Model`.

**Summary of changes**:
-  Introduces an API around a new "first class" `Scalar` type:
```rust
pub struct Scalar(Ratio<BigUint>);
```
- Adjusts `Masker::mask` to take a `scalar: Scalar` (instead of `scalar: f64`).
- Adjusts downstream clients of masking accordingly, using the `Scalar` API:
  - e.g. `Scalar::new(1, model_count)` creates an exact rational scalar `1 / model_count`.
  - e.g. `Scalar::unit()` instead of `1.0`.
- Adjusts masking tests, in particular:
  - Fixes the floating point bug in `test_masking_and_aggregation!` tests by using `Scalar`, and removes workaround.
  - Adjusts `test_xyz_scalar!` families of randomized scalar tests (introduced in #589).


rust/xaynet-core/src/mask/scalar.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Constructs a `Scalar` of unit value.
+    pub fn unit() -> Self {
+        Self(Ratio::one())
+    }
+
+    /// Convenience method for conversion into a non-negative ratio of `BigInt`.
+    pub(crate) fn into_ratio(self) -> Ratio<BigInt> {
+        let (numer, denom) = self.0.into();
+        Ratio::new(numer.into(), denom.into())
+    }
```
janpetschexain, line 48: this looks like a custom impl of `From`
```suggestion
impl From<Scalar> for Ratio<BigInt> {
    fn from(scalar: Scalar) -> Self {
        let (numer, denom) = self.0.into();
        Ratio::new(numer.into(), denom.into())
    }
}
```

rust/xaynet-core/src/mask/scalar.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    /// Convenience method for conversion to a non-negative ratio of `BigInt`.
+    pub(crate) fn to_ratio(&self) -> Ratio<BigInt> {
+        let numer = self.0.numer().clone();
+        let denom = self.0.denom().clone();
+        Ratio::new(numer.into(), denom.into())
```
janpetschexain, line 54: with the `From` impl this could be simplified to
```suggestion
        self.clone().into()
```

rust/xaynet-core/src/mask/scalar.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    fn from_ratio(r: Ratio<BigInt>) -> anyhow::Result<Self> {
+        let (inumer, idenom) = r.into();
+        let unumer = BigUint::try_from(inumer)?;
+        let udenom = BigUint::try_from(idenom)?;
+        Ok(Self(Ratio::new(unumer, udenom)))
+    }
```
janpetschexain, line 62: this looks like a custom impl of `TryFrom`:
```suggestion
impl TryFrom<Ratio<BigInt>> for Scalar {
    type Error = <BigUint as TryFrom<BigInt>::Error;
    
    fn try_from(ratio: Ratio<BigInt>) -> Result<Self, Self::Error> {
        let (numer, denom) = ratio.into();
        Ok(Self::new(numer.try_into()?, denom.try_into()?))
    }
}
```

rust/xaynet-core/src/mask/scalar.rs
```diff
@@ -0,0 +1,250 @@
...
+    weight: Ratio<BigUint>,
+    target: PrimitiveType,
+}
+
+/// An interface for conversion into a primitive value.
+///
+/// This trait is used to convert a [`Scalar`], which has its own internal
+/// representation, into a primitive type ([`f32`], [`f64`], [`i32`], [`i64`]).
+/// The opposite trait is [`FromPrimitive`].
+pub trait IntoPrimitive<P>: Sized {
```
janpetschexain, line 78: as a side note, we should probably later refactor the `IntoPrimitive` and `FromPrimitive` traits, not only for `Scalar` but also for `Model`.

for example the conversion error gets always constructed because of `.ok_or()`, where we should use `.ok_or_else()` for lazy evaluation.

also we have a duplication of trait declarations now, which only differ in the output type. we might want to unify them to something similar to
```rust
pub trait IntoPrimitives<P>: Sized {
    type Output<P>;

    fn into_primitive(self) -> Self::Output<P>;

    fn to_primitive(&self) -> Self::Output<P>;

    fn into_primitive_unchecked(self) -> P;
}
```

rust/xaynet-core/src/mask/scalar.rs
```diff
@@ -0,0 +1,250 @@
...
+            target: PrimitiveType::I32,
+        })
+    }
+
+    fn to_primitive(&self) -> Result<i32, ScalarCastError> {
+        let r = self.0.clone();
+        r.to_integer().to_i32().ok_or(ScalarCastError {
+            weight: r,
+            target: PrimitiveType::I32,
+        })
```
janpetschexain, line 133: code duplication could be reduced:
```suggestion
        self.clone().into_primitive()
```
similarly for the other impls below.

rust/xaynet-core/src/mask/scalar.rs
```diff
@@ -0,0 +1,250 @@
...
+        let r = float_to_ratio_bounded::<f64>(prim);
+        // safe unwrap: bounded non-negative ratio r
+        Self::from_ratio(r).unwrap()
+    }
+}
+
+/// Converts the primitive floating point value into a numerical value.
+///
+/// Maps positive/negative infinity to max/min of the primitive data type and NaN to zero.
+pub(crate) fn float_to_ratio_bounded<F: FloatCore>(f: F) -> Ratio<BigInt> {
```
janpetschexain, line 242: the docs don't reflect the content of the function.

also `Ratio<BigUint>` probably reflects the intention better than the current return type.

janpetschexain (commented):
looks good, i'm glad that this fixes the precision issues 👍 

i left some comments mostly regarding trait streamlining.

janpetschexain:
seems that even leaving just comments (neither approved nor requested changes) already puts the pr from draft to open. sorry, i didn't know that :/

rust/xaynet-mobile/src/settings.rs
```diff
@@ -42,9 +45,12 @@ impl Settings {
...
     }
 
-    /// Set the scalar use for masking
+    /// Set the scalar to use for masking
+    ///
+    /// # Panics
+    /// Panics if a `Scalar` cannot be constructed from the given `scalar`.
     pub fn set_scalar(&mut self, scalar: f64) {
-        self.scalar = scalar;
+        self.scalar = Scalar::from_primitive(scalar).unwrap()
```
Robert-Steiner, line 53: I would prefer it if we could avoid a potential `panic` here.  It's also not clear to me when it can fail. Maybe we can combine it with the `check` method.
for example 

```rust
#[derive(Clone, Debug)]
pub struct Settings {
    /// The Xaynet coordinator URL.
    url: Option<String>,
    /// The participant signing keys.
    keys: Option<SigningKeyPair>,
    /// The scalar used for masking.
    scalar: Result<Scalar>,
    /// The maximum possible size of a message.
    max_message_size: MaxMessageSize,
}

impl Settings {
    /// Create new empty settings.
    pub fn new() -> Self {
        Self {
            url: None,
            keys: None,
            scalar: Ok(Scalar::unit()),
            max_message_size: MaxMessageSize::default(),
        }
    }
    
    pub fn set_scalar(&mut self, scalar: f64) {
        self.scalar = Scalar::from_primitive(scalar)
    }
    
    pub fn check(&self) -> Result<(), SettingsError> {
        if self.url.is_none() {
            Err(SettingsError::MissingUrl)
        } else if self.keys.is_none() {
            Err(SettingsError::MissingKeys)
        else if self.scalar.is_err() {
            Err(SettingsError::Scalar...)
        } else {
            Ok(())
        }
    }
}
```

Robert-Steiner (commented):
A great improvement👍 The code looks much cleaner.
I just added a small note on the `Settings`.

rust/xaynet-mobile/src/settings.rs
janpetschexain: it should only fail if the scalar is `< 0` or `NaN` or `Inf`. i agree that we should add this to the checks.

rust/xaynet-core/src/mask/scalar.rs
finiteprods: >we might want to unify them to something similar to ...

this is a very nice idea. while we're at it, we can also lift some types up from `model` that are also used by `scalar`, like `PrimitiveCastError` and `PrimitiveType`. I will write a ticket.

>the conversion error gets always constructed because of `.ok_or()`, where we should use `.ok_or_else()` for lazy evaluation.

I also had this idea, but Clippy thought otherwise: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations. although strangely, it doesn't complain about my use of `unwrap_or_else` e.g.
```rust
fn from_primitive_bounded(prim: i32) -> Self {
    Self::from_primitive(prim).unwrap_or_else(|_| Self(Ratio::zero()))
}
```

rust/xaynet-core/src/mask/scalar.rs
finiteprods: good catch. updated docstring and refactored this a little so that the `Ratio<BigInt>` -> `Scalar` conversion also happens here, hence returning `Scalar` instead.

rust/xaynet-mobile/src/settings.rs
finiteprods: thanks for pointing out @Robert-Steiner (I'd completely missed the `check` function when I was changing this). agree that we should follow a similar pattern to the other error variants here.

janpetschexain (approved):

rust/xaynet-mobile/src/settings.rs
Robert-Steiner: Looks great 👍
we should check if we need to just the ffi layer that is used in `xayn`. we also need to generate a new c header file with cbindgen `v0.17.0`. it is described in the [readme](https://github.com/xaynetwork/xaynet/tree/master/rust/xaynet-mobile#generate-c-header-file) how to do this

finiteprods: looks like the `xain_ai` FFI layer uses `v0.11.0` of `xaynet` rather than `master` so should be fine (thanks @Robert-Steiner for checking).
C header file regenerated - there is a change in `gcc` that affects the Linux command to compile FFI tests, updated the readme.

Robert-Steiner (approved):
🚀

PR #715, opened 2021-02-04
XN-1580 broadcast after purge

janpetschexain:
**References**

- [XN-1580](https://xainag.atlassian.net/browse/XN-1580)
- ~depends on #711 and must be rebased after that is merged~

**Summary**

- move `Phase` related code into its own module
- add `process()` method to Phase and move the process-related code from `run()` there, this includes store-related operations
- add `broadcast()` method to `Phase` and move the broadcast-related code from `run()` there
- unify `run()` to call `process()`, then `purge()` and then `broadcast()` for all phases, merge `run()` into `run_phase()` (`purge()` is still a method of the `PhaseState`, but it might become a method of `Phase` or even `Handler` once we tackle the purge issue)
- move dict invalidation to the sum2 and error phases and move the store readiness check to the error phase transition (requires `next()` to be async), add test
- improve tracing messages
- clean up unnecessary trait bounds and unify generic parameter naming (i already noticed that i missed some and will address them in #723)
- clean up some unnecessary state from the previous pr


rust/xaynet-server/src/state_machine/phases/sum.rs
```diff
@@ -45,34 +47,19 @@ where
...
-        info!(
-            "in total {} sum messages accepted (min {} and max {} required)",
-            self.private.accepted, self.shared.state.sum.count.min, self.shared.state.sum.count.max,
-        );
-        info!("in total {} sum messages rejected", self.private.rejected);
-        info!("in total {} sum messages discarded", self.private.discarded);
+    impl_phase_purge_for_phasestate! { Sum, S }
 
+    async fn broadcast(&mut self) -> Result<(), PhaseStateError> {
+        info!("broadcasting sum dictionary");
```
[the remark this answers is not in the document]
janpetschexain: i moved the store fetching operations to `process()` in all phases.

i fully agree on the `purge` issue, we should tackle it next.

rust/xaynet-server/src/state_machine/phases/idle.rs
```diff
@@ -39,38 +41,50 @@ where
...
+        async fn process(self_: &mut PhaseState<Idle, S>) -> Result<(), PhaseStateError> {
+            info!("updating the keys");
+            self_.gen_round_keypair();
+
+            info!("updating round probabilities");
+            self_.update_round_probabilities();
+
+            info!("updating round seed");
+            self_.update_round_seed();
+
```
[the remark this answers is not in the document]
janpetschexain: i moved the invalidation broadcasts to the error phase, which is the earliest possible time. to avoid store waiting times in the error phase itself, i also moved the store readiness to the `next()` method (it somewhat belongs there anyways, because waiting for the store is a necessary condition to transition to the next phase. it is also a passive task, there is nothing the state machine can do to make the store ready, so it doesn't really fit into the `process()` of the error phase as well).

rust/xaynet-server/src/state_machine/phases/idle.rs
```diff
@@ -33,47 +31,49 @@ pub enum IdleStateError {
...
+impl<T> Phase<T> for PhaseState<Idle, T>
 where
-    S: Storage,
+    T: Storage,
 {
     const NAME: PhaseName = PhaseName::Idle;
 
-    async fn run(&mut self) -> Result<(), PhaseStateError> {
+    async fn process(&mut self) -> Result<(), PhaseStateError> {
+        info!("removing phase dictionaries from previous round");
```
Robert-Steiner, line 41: the invalidation of the dicts is missing here
```suggestion
        info!("removing phase dictionaries from previous round");
        info!("broadcasting invalidation of sum dictionary from previous round");
        events.broadcast_sum_dict(DictionaryUpdate::Invalidate);

        info!("broadcasting invalidation of seed dictionary from previous round");
        events.broadcast_seed_dict(DictionaryUpdate::Invalidate);
```

rust/xaynet-server/src/state_machine/coordinator.rs
```diff
@@ -104,6 +106,9 @@ pub struct CoordinatorState {
     pub update: PhaseParameters,
     /// The sum2 phase parameters.
     pub sum2: PhaseParameters,
+    /// The global model of the current round.
+    #[serde(skip)]
+    pub global_model: Option<Arc<Model>>,
```
Robert-Steiner, line 111: can we move this into the private state of the `Unmask` phase?

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -42,53 +44,57 @@ pub struct Unmask {
...
+        self.save_global_model().await?;
 
         Ok(())
     }
 
-    fn next(self) -> Option<StateMachine<S>> {
+    async fn broadcast(&mut self) -> Result<(), PhaseStateError> {
+        if let Some(ref global_model) = self.shared.state.global_model {
+            info!("broadcasting proof of the new global model");
+            self.shared
```
Robert-Steiner, line 73: I would prefer to move this into the process method so that the `broadcast` can not fail and and does not have to be `async`
but it is also fine like this

rust/xaynet-server/src/state_machine/phases/unmask.rs
```diff
@@ -130,36 +136,44 @@ where
...
-impl<S> PhaseState<Unmask, S>
+impl<T> PhaseState<Unmask, T>
 where
-    Self: Phase<S>,
-    S: Storage,
+    T: Storage,
 {
+    #[cfg(feature = "model-persistence")]
+    async fn save_global_model(&mut self) -> Result<(), UnmaskStateError> {
+        if let Some(ref global_model) = self.shared.state.global_model {
```
Robert-Steiner, line 151: here I would also use `unwrap` to avoid the `if let`
not sure if this is possible though

Robert-Steiner (commented):
Looks really nice 👍
I would be nice if we can move the global model into the private state of the Unmask phase
everything else are only suggestions

rust/xaynet-server/src/state_machine/phases/sum.rs
```diff
@@ -28,49 +28,57 @@ pub enum SumStateError {
...
-        self.shared
-            .events
-            .broadcast_sum_dict(DictionaryUpdate::New(Arc::new(sum_dict)));
+            .map_err(SumStateError::FetchSumDict)?;
+        if sum_dict.is_some() {
+            self.private.sum_dict = sum_dict;
+            Ok(())
+        } else {
+            Err(SumStateError::NoSumDict.into())
+        }
```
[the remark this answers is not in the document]
janpetschexain: i'll change that. i wrote the separated check with some retry loop in mind, where it probably makes sense to assign the value only in a successful iteration of the retry loop. but we don't have that yet anyways, so we can simplify it for now.

rust/xaynet-server/src/state_machine/phases/sum.rs
```diff
@@ -28,49 +28,57 @@ pub enum SumStateError {
...
+    }
+
+    async fn broadcast(&mut self) -> Result<(), PhaseStateError> {
+        if let Some(sum_dict) = self.private.sum_dict.take() {
+            info!("broadcasting sum dictionary");
+            self.shared
+                .events
+                .broadcast_sum_dict(DictionaryUpdate::New(Arc::new(sum_dict)));
```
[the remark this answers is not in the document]
janpetschexain: i used the `if let` to have a custom `unreachable!` panic message. `unwrap()`ing probably makes the code a bit more concise. i'll change that to `expect("...")`, this way we can get both benefits.

rust/xaynet-server/src/state_machine/phases/idle.rs
janpetschexain: i somewhat forgot about a successful round of the state machine :D in case of an error the invalidation is now handled in the error phase. in case of a successful round, i think the earliest we could invalidate the sum and seed dicts is after the sum2 phase, that means in `broadcast()` of the sum2 phase. we should probably add that there, wdyt?

rust/xaynet-server/src/state_machine/phases/unmask.rs
janpetschexain: i'll change it to an `expect()` as well. it should be possible because the `else` branch has return type `!` anyways.

rust/xaynet-server/src/state_machine/phases/idle.rs
Robert-Steiner: I like the idea to add this to the broadcast method of the sum2 phase 👍

rust/xaynet-server/src/state_machine/phases/sum.rs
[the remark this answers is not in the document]
Robert-Steiner: ahh i see 
my plan is to hide the retry (maybe even timeout) method in the storage layer. Not sure how we can do it but I think tower could help here😁

rust/xaynet-server/src/state_machine/phases/unmask.rs
janpetschexain: i'll move it, it is more of a store operation than a broadcast operation after all.

this also allows us to make all other `broadcast()`s sync and infallible.

Robert-Steiner (approved):
LGTM👍

PR #724, opened 2021-02-09
integration tests flag

janpetschexain:
**Summary**

- disable integration tests by default, enable them on the ci (opt-in instead of opt-out)
- fix doc tests

currently the integration tests fail by default, meaning that they fail if the required resources are not available. i made an attempt to disable them by default such that tests can be run locally without false negatives and changed the ci to run all of them.

unfortunately the `#[ignore]` for tests is binary, having configurable flags would be nice (could be done with feature flags, but this would pollute the features namespace on the other hand). anyways, ignored tests can be run locally or on the ci with `cargo test -- --ignored` or all tests can be run with `cargo test -- -Z unstable-options --include-ignored` (the latter is only marked unstable because it doesn't work with `cargo test --benches` but we have a separate benchmarks job anyways, other that that it is obviously using the stable compiler).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -209,14 +209,14 @@ jobs:
         working-directory: ./docker
         run: docker-compose up -d influxdb minio redis
 
-      - name: Run tests
+      - name: Run tests (unit & integration & doc)
         working-directory: ./rust
         env:
           RUSTFLAGS: "-D warnings"
         run: |
-          cargo test --all-targets
-          cargo test --all-targets --all-features
-          cargo test --all-features --doc
+          cargo test --lib --bins --examples --tests -- -Z unstable-options --include-ignored
+          cargo test --lib --bins --examples --tests --all-features -- -Z unstable-options --include-ignored
+          cargo test --doc --all-features
 
       - name: Stop docker-compose
         working-directory: ./docker
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -337,7 +337,7 @@ jobs:
         uses: actions-rs/tarpaulin@v0.1
         with:
           version: '0.16.0'
-          args: '--manifest-path rust/Cargo.toml --all-features --force-clean --lib --ignore-tests'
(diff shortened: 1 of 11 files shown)


Robert-Steiner (commented):
We have annotated the integration test with `#[ignore]` in the past and have changed it because there were problems with docs tests.
for example if you run `cargo test -- --ignored` in `rust` some doc tests will fail

if we don't want to run the integration tests we can use `cargo test -- --skip integration`

janpetschexain:
> We have annotated the integration test with `#[ignore]` in the past and have changed it because there were problems with docs tests.
> for example if you run `cargo test -- --ignored` in `rust` some doc tests will fail

the issue seems to be that the ` ```ignore ` annotation for doc tests triggers the same mechanisms like the `#[ignore]` annotations for tests. i fixed all doc tests to either work correctly or at least to be more specific than just ` ```ignore `.

while fixing this i saw that the [rustdoc book](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes) is quite right about `This is almost never what you want, ...`, because some of doc tests were outdated and nobody noticed because they were ignored. the only fixed doc test i'm not satisfied with are the `xaynet_server::{event, metric}` ones, it should be possible like [described here](https://doc.rust-lang.org/rustdoc/documentation-tests.html#documenting-macros), but i couldn't make it work, maybe there are too many intertwined dependency issues in our case.

nevertheless, passing the `--ignored` flag to `cargo test` should now run without issues.

> if we don't want to run the integration tests we can use `cargo test -- --skip integration`

i didn't know about the `--skip` flag, that is good to know 👍  but this doesn't change the fact that `cargo test` fails by default, which it shouldn't imo. wdyt about the changes now that the doc tests are fixed?

Robert-Steiner (approved):
>  the only fixed doc test i'm not satisfied with are the xaynet_server::{event, metric} ones, it should be possible like described here, but i couldn't make it work, maybe there are too many intertwined dependency issues in our case.

yeah, I had tried it too. but i think it's fine as it is now👍

PR #781, opened 2021-04-01
Pin bindgen dependencies

janpetschexain:
**Summary**

- pin `cbindgen` version: this requires `cbindgen` to be a dev-dependency instead of a command line tool, which in turn requires us to use a build script. a nice side effect is that we can now run it in stable instead of nightly (the nightly `[parse.expand]` option wasn't actually necessary because we wrap the macros and therefore never expose them as part of the ffi api; the generated header file stays the same).
- pin `pyo3` version
- update dependencies: some of them seem to have slipped through dependabot's watch
- rebuild the lock files: all sub-dependencies are up to date now and `cargo update -w` doesn't make any further updates

### DIFF
--- a/bindings/python/Cargo.lock
+++ b/bindings/python/Cargo.lock
@@ -2,6 +2,15 @@
 # It is not intended for manual editing.
 [[package]]
 name = "ansi_term"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
+dependencies = [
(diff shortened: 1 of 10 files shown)


Robert-Steiner (approved):
thanks for the pr👍
updating the sub-deps seems to solve the problem with influxdb🚀
what is the reason for pining ´pyo3´?

janpetschexain:
> what is the reason for pining ´pyo3´?

same as for `cbindgen`, i want to make sure the generated foreign api is fixed. (i'm not a 100% sure if this is needed here, my exposure to `pyo3` is quite limited, but it can't hurt and we may relax this in the future if needed.)

PR #845, opened 2021-07-14
DEVOPS-22 Auto cleanup stale/inactive Docker Hub tags

rsaffi:
The idea is to have this running every week and getting rid of these inactive tags we have on Docker Hub:

image

I've also made this possible to trigger via a `workflow_dispatch` event, so we can run it whenever we want.
You'll notice that my commit history changes drasticallly out of nowhere, that's because I spent yesterday testing this on a private repo so I wouldn't clutter `xaynet`'s Actions history.

### DIFF
--- a/.github/workflows/dockerhub-cleanup.yml
+++ b/.github/workflows/dockerhub-cleanup.yml
@@ -36,7 +36,7 @@ jobs:
           echo -e "Inactive tags:"
           /tmp/hub-tool tag ls xaynetwork/xaynet | grep -e STATUS -e inactive
           echo -e "\n\n"
-          for tag in $(/tmp/hub-tool tag ls xaynetwork/xaynet | grep inactive | grep -v -e "v[0-9]\.[0-9]\+\.[0-9]\+" | awk '{ print $1 }')
+          for tag in $(/tmp/hub-tool tag ls xaynetwork/xaynet | grep inactive | grep -v -e "v[0-9]\+\.[0-9]\+\.[0-9]\+" | awk '{ print $1 }')
             do
               /tmp/hub-tool tag rm -f ${tag}
           done
--- a/.github/workflows/dockerhub-cleanup.yml
+++ b/.github/workflows/dockerhub-cleanup.yml
@@ -21,7 +21,7 @@ jobs:
 
       - name: Setup hub-tool
         run: |
-					LATEST=$(curl --silent "https://api.github.com/repos/$1/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
+					LATEST=$(curl --silent "https://api.github.com/repos/docker/hub-tool/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
           wget https://github.com/docker/hub-tool/releases/download/${LATEST}/hub-tool-linux-amd64.tar.gz -O /tmp/hub-tool-linux-amd64.tar.gz
           tar xzvf /tmp/hub-tool-linux-amd64.tar.gz --strip-components 1 -C /usr/local/bin/ hub-tool/hub-tool
           /usr/bin/expect <(cat << EOF
--- a/.github/workflows/dockerhub-cleanup.yml
+++ b/.github/workflows/dockerhub-cleanup.yml
@@ -38,5 +38,5 @@ jobs:
           echo -e "\n\n"
           for tag in $(/tmp/hub-tool tag ls xaynetwork/xaynet | grep inactive | awk '{ print $1 }')
             do
-              echo /tmp/hub-tool tag rm -f ${tag}
+              /tmp/hub-tool tag rm -f ${tag}
           done
--- a/.github/workflows/dockerhub-cleanup.yml
+++ b/.github/workflows/dockerhub-cleanup.yml
@@ -21,7 +21,7 @@ jobs:
 
       - name: Setup hub-tool
         run: |
-					LATEST=$(curl --silent "https://api.github.com/repos/docker/hub-tool/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
+          LATEST=$(curl --silent "https://api.github.com/repos/docker/hub-tool/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
           wget https://github.com/docker/hub-tool/releases/download/${LATEST}/hub-tool-linux-amd64.tar.gz -O /tmp/hub-tool-linux-amd64.tar.gz
           tar xzvf /tmp/hub-tool-linux-amd64.tar.gz --strip-components 1 -C /usr/local/bin/ hub-tool/hub-tool
           /usr/bin/expect <(cat << EOF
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-cleanup.yml
```diff
@@ -0,0 +1,42 @@
...
+          mkdir -pv -m 700 ~/.docker
+          chmod -v 600 ~/.docker/config.json
+          echo -ne "ewogICJ1c2VybmFtZSI6ICJESFVTRVIiLAogICJwYXNzd29yZCI6ICJESFRPS0VOIgp9Cg==" | base64 -d > /tmp/auth.json
+          echo -ne "ewogICJhdXRocyI6IHsKICAgICJodWItdG9vbCI6IHsKICAgICAgImF1dGgiOiAiREhVU0VSVE9LRU4iCiAgICB9LAogICAgImh1Yi10b29sLXJlZnJlc2gtdG9rZW4iOiB7CiAgICAgICJhdXRoIjogIkRIVVNFUiIKICAgIH0sCiAgICAiaHViLXRvb2wtdG9rZW4iOiB7CiAgICAgICJhdXRoIjogIkRIVVNFUiIsCiAgICAgICJpZGVudGl0eXRva2VuIjogIkpXVFRPS0VOIgogICAgfQogIH0KfQoK" | base64 -d > ~/.docker/config.json
+          RUSERTOKEN=$(echo -ne "${DHUSER}:${DHTOKEN}" | base64 -w0)
+          RUSER=$(echo -ne "${DHUSER}:" | base64 -w0)
+          RTOKEN=$(echo -ne "${DHTOKEN}" | base64 -w0)
+          sed -i -e "s,DHUSERTOKEN,${RUSERTOKEN},g" -e "s,DHUSER,${RUSER},g" -e "s,DHTOKEN,${RTOKEN},g" /tmp/auth.json ~/.docker/config.json
+          JWT=$(curl -s -XPOST "https://hub.docker.com/v2/users/login" -H "Content-Type:application/json" -d "@/tmp/auth.json" | jq -r .token)
+          sed -i -e "s,JWTTOKEN,${JWT},g" ~/.docker/config.json
```
rsaffi, line 33: Reason for this hacky thing: https://github.com/docker/hub-tool/issues/176#issuecomment-879629162

abargiela (approved):
LGTM 🚀

rsaffi:
Hey @Robert-Steiner, do I have your blessing? 😬

Robert-Steiner:
> Hey @Robert-Steiner, do I have your blessing? 😬

I will take a look 👀

.github/workflows/dockerhub-cleanup.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Delete target tags
+        run: |
+          echo -e "Inactive tags:"
+          /tmp/hub-tool tag ls xaynetwork/xaynet | grep -e STATUS -e inactive
```
Robert-Steiner, line 37: I think it would make sense if we keep the containers tagged with a version even though they are inactive (like `v0.9.0`) wdyt?

Robert-Steiner (approved):
looks good 👍
Added some suggestions

.github/workflows/dockerhub-cleanup.yml
```diff
@@ -0,0 +1,42 @@
...
+          LATEST=$(curl -s "https://api.github.com/repos/docker/hub-tool/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
+          wget https://github.com/docker/hub-tool/releases/download/${LATEST}/hub-tool-linux-amd64.tar.gz -O /tmp/hub-tool-linux-amd64.tar.gz
+          tar xzvf /tmp/hub-tool-linux-amd64.tar.gz --strip-components 1 -C /tmp hub-tool/hub-tool
+          mkdir -pv -m 700 ~/.docker
+          chmod -v 600 ~/.docker/config.json
+          echo -ne "ewogICJ1c2VybmFtZSI6ICJESFVTRVIiLAogICJwYXNzd29yZCI6ICJESFRPS0VOIgp9Cg==" | base64 -d > /tmp/auth.json
+          echo -ne "ewogICJhdXRocyI6IHsKICAgICJodWItdG9vbCI6IHsKICAgICAgImF1dGgiOiAiREhVU0VSVE9LRU4iCiAgICB9LAogICAgImh1Yi10b29sLXJlZnJlc2gtdG9rZW4iOiB7CiAgICAgICJhdXRoIjogIkRIVVNFUiIKICAgIH0sCiAgICAiaHViLXRvb2wtdG9rZW4iOiB7CiAgICAgICJhdXRoIjogIkRIVVNFUiIsCiAgICAgICJpZGVudGl0eXRva2VuIjogIkpXVFRPS0VOIgogICAgfQogIH0KfQoK" | base64 -d > ~/.docker/config.json
+          RUSERTOKEN=$(echo -ne "${DHUSER}:${DHTOKEN}" | base64 -w0)
+          RUSER=$(echo -ne "${DHUSER}:" | base64 -w0)
+          RTOKEN=$(echo -ne "${DHTOKEN}" | base64 -w0)
```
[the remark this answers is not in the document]
rsaffi: I thought of that, we _can_, of course, but what do we gain? I mean, I checked on my private repo where I was running tests and because RTOKEN is defined from DHTOKEN (which itself comes from a secret), nothing is displayed on the logs from the runners. 
So I don't see any gain we'd have by having that one also be a secret. If/when we change the Docker Hub Token, we'd need to update two secrets, the DOCKERHUB_TOKEN and RTOKEN, otherwise the CI would break.

.github/workflows/dockerhub-cleanup.yml
```diff
@@ -0,0 +1,42 @@
...
+          sed -i -e "s,DHUSERTOKEN,${RUSERTOKEN},g" -e "s,DHUSER,${RUSER},g" -e "s,DHTOKEN,${RTOKEN},g" /tmp/auth.json ~/.docker/config.json
+          JWT=$(curl -s -XPOST "https://hub.docker.com/v2/users/login" -H "Content-Type:application/json" -d "@/tmp/auth.json" | jq -r .token)
```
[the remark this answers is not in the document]
rsaffi: That was also a concern, but in the end none of those commands echo anything. Also, why `echo "::add-mask::$JWT"`? Why `echo`? I mean, I think it'd make sense here:

```bash
sed -i -e "s,JWTTOKEN,::add-mask::${JWT},g" ~/.docker/config.json
```

But I'm not sure whether GitHub would be clever to effectively evaluate that and only leave the `${JWT}` to be the replacing part of that `sed` or if it would stupidly replace `JWTTOKEN` for litterally `::add-mask::${JWT}`.

.github/workflows/dockerhub-cleanup.yml
rsaffi: Excelent point. Will definitely do that! :)

rsaffi: @Robert-Steiner ✅ Done!

.github/workflows/dockerhub-cleanup.yml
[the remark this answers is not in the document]
Robert-Steiner: > why echo "::add-mask::$JWT" 
I think that is how the command is sent to their vm but I'm also fine not having it

Robert-Steiner (approved):

PR #928, opened 2021-11-12
Bump rustc and fix compilation

janpetschexain:
**Summary**

- remove `xaynet-analytics` from the workspace for now
  - issues between `isar` and `rustc > 1.52` need to be fixed before this can be enabled again
  - tarpaulin needs to be excluded explicitely otherwise the crate still shows up with zero percent coverage
- update ci to `rustc 1.55.0`
- fix new clippy and doc lints

### DIFF
--- a/rust/xaynet-analytics/src/lib.rs
+++ b/rust/xaynet-analytics/src/lib.rs
@@ -8,7 +8,11 @@
 //! a framework that allows mobile applications to collect and aggregate
 //! analytics data via the _Privacy-Enhancing Technology_ (PET) protocol.
 
+#[cfg(not(tarpaulin))]
 pub mod controller;
+#[cfg(not(tarpaulin))]
 pub mod data_combination;
+#[cfg(not(tarpaulin))]
 pub mod database;
+#[cfg(not(tarpaulin))]
 pub mod sender;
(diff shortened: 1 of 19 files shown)


rust/xaynet-analytics/src/lib.rs
```diff
@@ -8,7 +8,11 @@
 //! a framework that allows mobile applications to collect and aggregate
 //! analytics data via the _Privacy-Enhancing Technology_ (PET) protocol.
 
+#[cfg(not(tarpaulin))]
```
acrrd, line 11: Do we need this since this crate is commented in the manifest?
Same question for the `--exclude ` option in the ci, why it is needed? tarpaulin does not look at the `members` variable but check all directories anyway?

acrrd: ok thanks. Can you maybe open an issue on their repo? I think it is a bug.

acrrd (approved):

rust/xaynet-analytics/src/lib.rs
janpetschexain: sure, we should have a minimum runnable example for them to showcase this, i'll put it on my todo list.

PR #1004, opened 2022-02-21
Remove k8s deployment

Robert-Steiner:


.github/workflows/dockerhub-master.yml
```diff
@@ -31,23 +31,6 @@ jobs:
           push: true
           build-args: COORDINATOR_FEATURES=metrics
 
-      - name: Apply kubeconfig
```
janpetschexain, line 34: the same steps exist in the `dockerhub-pr-with-parameters.yaml`, don't they need to be removed there as well?

Robert-Steiner: good catch! thanks

janpetschexain (approved):
lgtm

